PR #9, opened 2022-02-22
Try using new actions for e2e test

mattmoor:


.github/workflows/mink-e2e.yaml
```diff
@@ -0,0 +1,62 @@
...
+      run: |
+        ko apply -BRf config/
+
+    - name: Build examples
+      run: |
+        for cfg in $(find ./examples/ -name '*.yaml'); do
+          DIGEST=$(mink run task apko -- --path=$cfg)
+
+          echo $cfg produced: $DIGEST
+        done
```
mattmoor, line 59: We don't do anything yet, but this will fail if we break the examples, so it may be worthwhile coverage as-is.

If not, we should consider what interesting inspection we might want to do on the images.

dlorenc (approved):

PR #10, opened 2022-02-23
feature: apko publishing

kaniini:
Introduces `apko` publishing, use `docker login` or equivalent to set up the docker keychain first.

Example published image on Docker Hub (haha I couldn't help myself): https://hub.docker.com/repository/docker/kaniini/apko-example

### DIFF
--- a/cmd/apko/cli/publish.go
+++ b/cmd/apko/cli/publish.go
@@ -1,0 +1,79 @@
+// Copyright 2022 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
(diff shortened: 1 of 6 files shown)


cmd/apko/cli/publish.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func Publish() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "publish",
+		Short: "Build and publish an image",
+		Long: `Publish a built image from a YAML configuration file.
+
+It is assumed that you have used "docker login" to store credentials
+in a keychain.`,
```
mattmoor, line 36: GGCR's `crane` has a `login` cobra command that is set up for downstream tools to link in.  We do this in cosign and ko IIRC.

pkg/build/oci/oci.go
```diff
@@ -97,6 +108,24 @@ func BuildImageRefFromLayer(imageRef string, layerTarGZ string, outputTarGZ stri
...
+	if err != nil {
+		return err
+	}
+
+	imgRef, err := name.ParseReference(imageRef)
+	if err != nil {
+		return errors.Wrap(err, "unable to parse reference")
+	}
+
+	err = remote.Write(imgRef, v1Image, remote.WithAuthFromKeychain(authn.DefaultKeychain))
```
mattmoor, line 125: If we incorporate the credential helper stuff here, then we can use this with workload identities without needing to set up a bunch of extra stuff.  This is how `krane` is different from `crane`.

Speaking of, we should be able to drop the `krane push` from `config/task.yaml` with this in.  I can test and send a PR.

PR #12, opened 2022-02-24
Leverage `apko publish` in Tekton task.

mattmoor:
This change has three main parts:
1. Build in Jason Hall's fancy new keychain logic to support workload identity credentials OOTB,
2. Add an `--image-refs` flag (to match `krane`, and `ko`),
3. Drop the `krane push` step that we no longer need!


cmd/apko/cli/publish.go
```diff
@@ -37,25 +40,39 @@ in a keychain.`,
 		Example: `  apko publish <config.yaml> <tag>`,
 		Args:    cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return PublishCmd(cmd.Context(), args[0], args[1])
+			digest, err := PublishCmd(cmd.Context(), args[0], args[1])
+			if err != nil {
+				return err
+			}
+			if imageRefs != "" {
+				if err := os.WriteFile(imageRefs, []byte(digest.String()), 0666); err != nil {
```
mattmoor, line 48: I originally had 0600 here, but since this runs as `root` and the final build step is `nonroot` I tweaked this.

We can switch the final step in the task to run as root, or `nosec` this.

kaniini: I would suggest `nosec` :)

cmd/apko/cli/publish.go
```diff
@@ -37,25 +40,39 @@ in a keychain.`,
 		Example: `  apko publish <config.yaml> <tag>`,
 		Args:    cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return PublishCmd(cmd.Context(), args[0], args[1])
+			digest, err := PublishCmd(cmd.Context(), args[0], args[1])
+			if err != nil {
+				return err
+			}
+			if imageRefs != "" {
```
kaniini, line 47: Would it be possible to move this into `PublishCmd`?  I'd prefer to keep our indentation levels low.

kaniini (changes requested):
Looks good, just one minor nitpick about the `PublishCmd` part :)

cmd/apko/cli/publish.go
mattmoor: Yeah, I'll just have to plumb the variable thru.

kaniini: We could have a PublishOptions{} struct I guess, seems like the best way to do it.

mattmoor: ok done on both

mattmoor: Github swallowed the second comment until just now.  One trick we use is to make things like `PublishCmd` methods on a struct, and define flags as variables on fields on that struct.

e.g.

```go
type Publish struct {
  imageRefs string
}

func (p *Publish) Execute() ...


...
p := Publish{}
cmd := &cobra.Command{
		Use:   "publish",
		Args:    cobra.ExactArgs(2),
		RunE: p.Execute,
}
cmd.Flags().StringVar(&p.imageRefs, "image-refs", "", "path to file where a list of the published image references will be written")

```

mattmoor: @n3wscott refactored cosign to use this a bunch, and we can define options structs and use interfaces to build up more elaborate flag surfaces.

kaniini: Lets open an issue about it and talk about doing this refactoring in v0.2!

PR #15, opened 2022-02-24
Add goreleaser and ko build

mattmoor:
Here's a sample release on my fork: https://github.com/mattmoor/apko/releases/tag/v0.0.11

I downloaded and verified the signature here (still not sure what the ctlog issue is cc @dlorenc ):
```
$ COSIGN_EXPERIMENTAL=true cosign verify-blob --signature apko_0.0.11_linux_arm64.sig --cert apko_0.0.11_linux_arm64.crt apko_0.0.11_linux_arm64/apko 
Verified OK
Error: verifying blob [apko_0.0.11_linux_arm64/apko]: signature not found in transparency log
main.go:46: error during command execution: verifying blob [apko_0.0.11_linux_arm64/apko]: signature not found in transparency log
```

We can verify the image signature:
```
$ COSIGN_EXPERIMENTAL=true cosign verify ghcr.io/mattmoor/apko:v0.0.11@sha256:7c29ab9bfadeff2367c298d99e8c27479823bdee114265f64beceb7547156753 | jq .

Verification for ghcr.io/mattmoor/apko@sha256:7c29ab9bfadeff2367c298d99e8c27479823bdee114265f64beceb7547156753 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - Any certificates were verified against the Fulcio roots.
[
  {
    "critical": {
      "identity": {
        "docker-reference": "ghcr.io/mattmoor/apko"
      },
      "image": {
        "docker-manifest-digest": "sha256:7c29ab9bfadeff2367c298d99e8c27479823bdee114265f64beceb7547156753"
      },
      "type": "cosign container image signature"
    },
    "optional": {
      "Bundle": {
        "SignedEntryTimestamp": "MEUCIQDTou5va1Gw9AmWJNXDcpB9W+ShLZS57QWaLjWfow5LxAIgYd/VMju4MCPKyHcar6y7eQ2Gi+ubaWOJEi1eZIxXXUM=",
        "Payload": {
[...]

PR #16, opened 2022-02-24
update gorelease to use the sign step to sign the build artifacts

cpanato:
- update gorelease to use the sign step to sign the build artifacts

rehearsal: https://github.com/cpanato/apko/releases/tag/v99.99.99

my question is we will distribute the tar.gz it is not easier just the binary? otherwise, the user needs to download and unpack it just to get one file (besides the license)

follow up of https://github.com/chainguard-dev/apko/pull/15/files

cc @kaniini @mattmoor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.so
 *.dylib
 /apko
+dist/
 
 # Test binary, built with `go test -c`
 *.test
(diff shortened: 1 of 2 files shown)


mattmoor (approved):
re: .tar.gz vs. binary, I'm fine with either, I was just copying what I did in `mattmoor/mink` ages ago (which I think I copied from elsewhere).

cpanato:
any thoughts @kaniini ?

PR #35, opened 2022-03-03
support using proot for rootless image builds

kaniini:
As suggested from an IRC conversation with @boomanaiden154 on `#alpine-devel`.

### DIFF
--- a/pkg/build/executor.go
+++ b/pkg/build/executor.go
@@ -22,6 +22,8 @@ import (
 )
 
 func (bc *Context) Execute(name string, arg ...string) error {
+	logname := name
+
 	if bc.UseProot {
 		arg = append(arg, "")
 		arg = append(arg, "")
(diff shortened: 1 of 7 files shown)


pkg/build/executor.go
```diff
@@ -21,13 +21,24 @@ import (
 	"github.com/pkg/errors"
 )
 
-func Execute(name string, arg ...string) error {
+func (bc *Context) Execute(name string, arg ...string) error {
+	logname := name
+
+	if bc.UseProot {
+		arg = append(arg, "")
+		arg = append(arg, "")
```
dlorenc, line 29: If I'm reading this right, you're shifting the arguments back two so you can prefix them with `-0 name`. It might be a bit simpler to read if it's setup like:

`arg = append([]string{"-0", "name"}, arg...)`
https://go.dev/play/p/eCmQi956x1T

dlorenc (commented):
Nice!

pkg/build/executor.go
kaniini: Thanks!

mattmoor (commented):
Should we be able to test this by having the task `runAsUser: <id of nobody>` and pass this flag here: https://github.com/chainguard-dev/apko/blob/b429e9db6e014aa7e063ffd2a55c5d8eca81f1b9/config/task.yaml#L41-L48

mattmoor:
I am also wondering whether we should default this flag to `true` when we aren't running as `root` 🤔

PR #38, opened 2022-03-03
fetch APK keyring via https

### LINKED ISSUE #27: add support for downloading keyring keys
For distributions other than Alpine v3.14 through edge, due to ongoing evolution of the keyring, the default keyring is inadequate.  Thus, it's important to be able to bootstrap the keyring with keys stored outside the apko Docker image.

Accordingly, it would be nice to be able to download the keys necessary to bootstrap the keyring.  This would make it easier to use apko with distributions such as OpenWrt and Adelie too.

kaniini:
Add support for fetching an APK keyring via https.  Closes #27.

### DIFF
--- a/pkg/build/apk.go
+++ b/pkg/build/apk.go
@@ -15,7 +15,9 @@
 package build
 
 import (
+	"io"
 	"log"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -98,6 +100,16 @@ func (bc *Context) InitApkKeyring() (err error) {
 			data, err = os.ReadFile(element)
 			if err != nil {
 				return errors.Wrap(err, "failed to read apk key")
+			}
+		} else if asUrl.Scheme == "https" {
+			resp, err := http.Get(asUrl.String())
+			if err != nil {
+				return errors.Wrap(err, "failed to fetch apk key")
+			}
+
+			data, err = io.ReadAll(resp.Body)
+			if err != nil {
+				return errors.Wrap(err, "failed to read apk key response")
 			}
 		} else {
 			return errors.Errorf("scheme %s not supported", asUrl.Scheme)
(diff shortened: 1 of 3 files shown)


pkg/build/apk.go
```diff
@@ -85,9 +89,31 @@ func (bc *Context) InitApkKeyring() (err error) {
 	for _, element := range keyFiles {
 		log.Printf("installing key %v", element)
 
-		data, err := os.ReadFile(element)
+		asURI := uri.New(element)
+		asURL, err := url.Parse(string(asURI))
```
mattmoor, line 93: Why are you parsing it twice?

pkg/build/apk.go
```diff
@@ -85,9 +89,31 @@ func (bc *Context) InitApkKeyring() (err error) {
...
+		case "file":
+			data, err = os.ReadFile(element)
+			if err != nil {
+				return errors.Wrap(err, "failed to read apk key")
+			}
+		case "https":
+			resp, err := http.Get(asURL.String())
+			if err != nil {
+				return errors.Wrap(err, "failed to fetch apk key")
+			}
```
mattmoor, line 109: Go doesn't return an `error` when the status code is not `2xx`, so you should also check `resp.StatusCode` and `defer resp.Body.Close()`

pkg/build/apk.go
kaniini: The first time is to normalize the input into a URI, which can then be used with `net/url`.

pkg/build/apk.go
kaniini: Handled

pkg/build/apk.go
kaniini: I added a comment explaining it.

mattmoor (approved):

PR #40, opened 2022-03-03
test: replace mkdir+defer with native t.tmpdir

kruskall:


kaniini:
Nice cleanup but it seems the CI is failing.  I'll merge it once that's resolved!

kruskall:
> Nice cleanup but it seems the CI is failing. I'll merge it once that's resolved!

Should be fixed. 

This also affects the current main but since the error from `os.RemoveAll` (https://github.com/chainguard-dev/apko/blob/c459c8b35f1a0b815e83afb32c4fffb1b3433c3e/pkg/build/apk_test.go#L29) is not checked the test is passing.

kaniini:
Looks great, thanks!

Issue #48, opened 2022-03-03
Zero out timestamps for reproducibility


imjasonh:
For layer metadata:
https://github.com/chainguard-dev/apko/blob/2656385887c06e9f5b48864abbeb32bcbcd19f89/pkg/build/oci/oci.go#L66

And for tar entries:
https://github.com/chainguard-dev/apko/blob/2656385887c06e9f5b48864abbeb32bcbcd19f89/pkg/tarball/tarball.go#L54

This means that running `apko publish` twice a minute apart will produce different image digests, even if their contents would otherwise be the same.

These should default to `time.Time{}` (Unix epoch) unless the `SOURCE_EPOCH_DATE` env var is set, in which case that should be parsed as RFC3339 and used.

PR #51, opened 2022-03-03
feat: make image generation deterministic by default

kruskall:
This PR makes image generation deterministic by default:

- timestamps are zeroed out for reproducibility (see https://github.com/chainguard-dev/apko/issues/48)
- `apk` is writing a tar archive to `lib/apk/db/scripts.tar`, to support reproducible builds the timestamps inside the archive needs to be deterministic/zeroed.

Note: according to the spec of `SOURCE_DATE_EPOCH` (https://reproducible-builds.org/specs/source-date-epoch/):

> Build processes MUST use this variable for embedded timestamps in place of the "current" date and time. 

Since `time.Now` has been replaced with the zero value (January 1, year 1, 00:00:00.000000000 UTC) there's no need for `SOURCE_DATE_EPOCH`.

Note: reproducible image generation has been tested with https://diffoscope.org/

Closes https://github.com/chainguard-dev/apko/issues/48

cc @imjasonh


pkg/build/apk.go
```diff
@@ -174,3 +176,52 @@ func (bc *Context) FixateApkWorld() error {
 
 	return bc.Execute("apk", args...)
 }
+
+func (bc *Context) fixScriptsTar() error {
```
kruskall, line 180: `fix` is a generic/broad term, a better name should be used (suggestions are welcomed :sweat_smile: )

kruskall:
Note: reproducible image generation has been tested with https://diffoscope.org/

kaniini:
> Since time.Now has been replaced with the zero value (January 1, year 1, 00:00:00.000000000 UTC) there's no need for SOURCE_DATE_EPOCH.

Some users do not like seeing timestamps of 1970 in things like `ls` output.  It would be nice to support `SOURCE_DATE_EPOCH`, and also have an option to set `SOURCE_DATE_EPOCH` to time.Now().

pkg/build/apk.go
kaniini: `normalizeApkScriptsTar()` I think.  Once the APK distributions adopt apk3, this file will go away, so we should make sure the file exists before normalizing it too.

kruskall:
> Some users do not like seeing timestamps of 1970 in things like `ls` output. It would be nice to support `SOURCE_DATE_EPOCH`, and also have an option to set `SOURCE_DATE_EPOCH` to time.Now().

I'm ok with supporting `SOURCE_DATE_EPOCH` but an environment variable is a bit opaque in term of UI, I think there should be a easier/simpler way for the user to make the image reproducible. Maybe an additional flag ? WDYT ?

pkg/build/apk.go
kruskall: Oh, good point! Fixed.

kaniini:
> I'm ok with supporting SOURCE_DATE_EPOCH but an environment variable is a bit opaque in term of UI, I think there should be a easier/simpler way for the user to make the image reproducible. Maybe an additional flag ? WDYT ?

Yes, a global option `--build-date` would be great, with `--build-date=now` being equivalent to `time.Now()`.

pkg/build/image_builder.go
```diff
@@ -73,6 +73,10 @@ func (bc *Context) BuildImage() error {
 		return errors.Wrap(err, "failed to fixate apk world")
 	}
 
+	if err := bc.normalizeApkScriptsTar(); err != nil {
+		return err
```
kruskall, line 77: Should this be a warning ?

pkg/build/build.go
```diff
@@ -29,6 +30,7 @@ type Context struct {
 	WorkDir            string
 	TarballPath        string
 	UseProot           bool
+	SourceDataEpoch    time.Time
```
kaniini, line 33: Should this not be `SourceDateEpoch`?

pkg/cli/publish.go
```diff
@@ -76,13 +80,36 @@ func PublishCmd(ctx context.Context, configFile string, outputRefs string, usePr
 		UseProot:           useProot,
 	}
 
+	// SOURCE_DATE_EPOCH takes priority
+	if v, ok := os.LookupEnv("SOURCE_DATE_EPOCH"); ok {
```
kaniini, line 84: Might be better to refactor global options into a global context, kinda like in `cosign`.

pkg/cli/build-minirootfs.go
```diff
@@ -64,6 +68,29 @@ func BuildMinirootFSCmd(ctx context.Context, configFile string, outputTarGZ stri
...
+			// If the value is malformed, the build process
+			// SHOULD exit with a non-zero error code.
+			return err
+		}
+
+		bc.SourceDataEpoch = time.Unix(sec, 0)
+	} else if buildDate == "now" {
+		bc.SourceDataEpoch = time.Now()
+	} else {
+		sec, err := strconv.ParseInt(v, 10, 64)
```
kruskall, line 89: What's the format for the build-date flag ? I'm using the same logic for `SOURCE_DATE_EPOCH` (from the spec: The `value MUST be an ASCII representation of an integer with no fractional component, identical to the output format of date +%s.`) but it's probably better to adopt a more user-friendly format for this flag.

pkg/cli/build-minirootfs.go
```diff
@@ -35,16 +38,17 @@ func BuildMinirootFS() *cobra.Command {
 		Example: `  apko build-minirootfs <config.yaml> <output.tar.gz>`,
 		Args:    cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return BuildMinirootFSCmd(cmd.Context(), args[0], args[1], useProot)
+			return BuildMinirootFSCmd(cmd.Context(), args[0], args[1], useProot, buildDate)
 		},
 	}
 
 	cmd.Flags().BoolVar(&useProot, "use-proot", false, "use proot to simulate privileged operations")
+	cmd.Flags().StringVar(&buildDate, "build-date", "now", "date used for the timestamps of the files inside the image")
```
kruskall, line 46: `build-date` defaults to `now`. This means images are not reproducible by default. Is this ok ?

pkg/build/build.go
kruskall: It should, good catch :+1: 

note to self: I need to sleep :facepalm:

pkg/cli/build-minirootfs.go
kaniini: Yes, that seems like the best option.  If you want a reproducible image, then ask for a reproducible image.

pkg/build/image_builder.go
kaniini: No, if it fails to normalize the `scripts.tar`, we should consider it an error.

pkg/cli/build-minirootfs.go
kaniini: Probably the ISO format outlined in the bug, support both ideally.

pkg/cli/build-minirootfs.go
imjasonh: What's the use case for preferring a non-reproducible image?

I think if we can make it reproducible by default there's a lot of nice wins you can get for free. For one, all your pishes get a lot faster because there's nothing new to upload most of the time.

kaniini: People freak out about the date being 1970.  I don't really *care* per se about the default, as much as I do about having an answer for the people who freak out about the date being 1970.

imjasonh: Yeah, I think that's just a documentation issue. For example: https://github.com/google/ko/#why-are-my-images-all-created-in-1970

I don't want to make the experience subtly worse for the vast majority of users who don't care and could benefit from reproducible images, just to appease the relatively small (but vocal) minority who notice and for some reason care. For them, there's an escape hatch in `SOURCE_DATE_EPOCH` and/or the flag.

For more precedent, buildpacks also defaults to reproducible timestamps (they picked 1980 but whatever), which they [document](https://buildpacks.io/docs/features/reproducibility/#consequences-and-caveats), and they're only just now [working on making it overrideable](https://github.com/buildpacks/spec/pull/292).

kaniini: okay, let's make the default `0` and then document it

kruskall: FWIW if having 1970 (or any specific old date) is an issue, you could pick the date of the first commit to the apko repository (or any specific date that holds value for the project) as a default value I think.

kaniini: Nah, lets focus on `0`, since that's the standard one.  We can address this as an FAQ item.

pkg/cli/build-minirootfs.go
kruskall: Parsing the string as RFC3339

pkg/cli/build-minirootfs.go
kruskall: Should be fixed, thanks everyone!

pkg/cli/publish.go
kruskall: this is cleaner now that it's using functional options, I'll leave other flag-related improvements to another PR.

kruskall:
rebased on the main branch and fixed the merge conflicts :tada:

kaniini (approved):

kaniini:
@imjasonh @puerco what do you think?

imjasonh (approved):
LGTM!

imjasonh:
Ugh, I'm a dummy, it's supposed to be seconds since epoch, not RFC3339: https://github.com/google/ko/blob/3fc720f912ac5192cff518a31fa6621e4a0dba51/pkg/commands/config.go#L129

Sorry for the confusion. I won't be able to send a PR until maybe tonight, but I'll happily review anything.

imjasonh:
Extra dumb! That's exactly what you have. Ignore me, it's the Fridays.

PR #52, opened 2022-03-03
emulate busybox install when running under proot

### LINKED ISSUE #50: proot: emulate busybox --install -s
When running under a `proot` jail, we cannot run scripts, such as the one which installs the busybox install.  One option would be to check for /bin/busybox in the target image and run it ourselves.  This seems to be the only hook we really need to care about.

kaniini:
Closes #50.

### DIFF
--- a/pkg/build/executor.go
+++ b/pkg/build/executor.go
@@ -21,6 +21,27 @@ import (
 	"github.com/pkg/errors"
 )
 
+func runCommand(cmd *exec.Cmd, logname string) error {
+	log.Printf("running: %v", cmd.String())
+
+	output, err := cmd.CombinedOutput()
+	if output != nil {
+		log.Printf("[%s] %s", logname, output)
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to run %s", cmd.String())
+	}
+
+	return nil
+}
+
+// TODO(kaniini): Add support for using qemu-binfmt here for multiarch.
+func (bc *Context) ExecuteChroot(name string, arg ...string) error {
+	arg = append([]string{"-S", bc.WorkDir, name}, arg...)
+	cmd := exec.Command("proot", arg...)
+	return runCommand(cmd, name)
+}
+
 func (bc *Context) Execute(name string, arg ...string) error {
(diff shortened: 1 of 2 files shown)


pkg/build/image_builder.go
```diff
@@ -82,3 +92,25 @@ func (bc *Context) BuildImage() error {
 	log.Printf("finished building filesystem in %s", bc.WorkDir)
 	return nil
 }
+
+// Installs the BusyBox symlinks, if appropriate.
+func (bc *Context) InstallBusyboxSymlinks() error {
+	path := filepath.Join(bc.WorkDir, "/bin/busybox")
```
kruskall, line 98:
```suggestion
	path := filepath.Join(bc.WorkDir, "bin", "busybox")
```

pkg/build/executor.go
```diff
@@ -21,24 +21,35 @@ import (
...
+func runCommand(cmd *exec.Cmd, logname string) error {
 	log.Printf("running: %v", cmd.String())
 
 	output, err := cmd.CombinedOutput()
 	if output != nil {
 		log.Printf("[%s] %s", logname, output)
 	}
 	if err != nil {
-		return errors.Wrapf(err, "failed to run %s", name)
+		return errors.Wrapf(err, "failed to run %s", cmd.String())
```
panekj, line 32:
```suggestion
		return errors.Wrapf(err, "failed to run %s", cmd)
```
If underlying type struct implements `String()` function, you can omit using it when passing variable to fmt when you specify `%s`, but you can ignore this suggestion as it's purely style

pkg/build/executor.go
```diff
@@ -21,24 +21,35 @@ import (
...
-	logname := name
-
-	if bc.UseProot {
-		arg = append([]string{"-0", name}, arg...)
-		name = "proot"
-	}
-
-	cmd := exec.Command(name, arg...)
+func runCommand(cmd *exec.Cmd, logname string) error {
 	log.Printf("running: %v", cmd.String())
```
panekj, line 25: Mixed style `%v` and `%s` as seen below

PR #56, opened 2022-03-04
Multiarch support

imjasonh:
`apko publish` produces a multiarch image index. By default, this includes six common archs, which can be overridden with `archs:` in the image config YAML, or values in the `--arch` flag (in increasing order of preference).

An attempt is made to accept both apk-style arch strings ("x86_86") and OCI-style arch strings ("amd64"), but more could probably be done. There's a TODO to support armv7 and armhf (arm/v7 and arm/v6 respectively, in OCI terms).

apk keys are extracted from arch-specific locations (e.g., `/usr/share/apk/keys/s390x/*`) and placed in the standard location (`/etc/apk/keys/*`).

`apko build` and `apko build-minirootfs` both only produce a single arch's tarball, for the current runtime arch. If the image config specifies archs in this case, a warning is logged to avoid confusion.

### DIFF
--- a/config/task.yaml
+++ b/config/task.yaml
@@ -44,6 +44,7 @@ spec:
       args:
       - publish
       - --image-refs=/tekton/results/IMAGES
+      - --arch=x86_64
       - $(params["path"])
       - $(params["dev.mink.images.target"])
 
--- a/pkg/build/apk.go
+++ b/pkg/build/apk.go
@@ -30,8 +30,6 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-var systemKeyringLocations = []string{"/etc/apk/keys/"}
-
 // Programmatic wrapper around apk-tools.  For now, this is done with os.Exec(),
 // but this has been designed so that we can port it easily to use libapk-go once
 // it is ready.
@@ -42,16 +40,18 @@ var systemKeyringLocations = []string{"/etc/apk/keys/"}
 func (bc *Context) InitApkDB() error {
 	log.Printf("initializing apk database")
 
-	return bc.Execute("apk", "add", "--initdb", "--root", bc.WorkDir)
+	return bc.Execute("apk", "add", "--initdb", "--arch", bc.Arch.ToAPK(), "--root", bc.WorkDir)
 }
 
 // loadSystemKeyring returns the keys found in the system keyring
 // directory by trying some common locations. These can be overridden
 // by passing one or more directories as arguments.
-func (*Context) loadSystemKeyring(locations ...string) ([]string, error) {
+func (bc *Context) loadSystemKeyring(locations ...string) ([]string, error) {
 	var ring []string
 	if len(locations) == 0 {
-		locations = systemKeyringLocations
+		locations = []string{
+			filepath.Join("/usr/share/apk/keys/", bc.Arch.ToAPK()),
+		}
 	}
 	for _, d := range locations {
 		keyFiles, err := os.ReadDir(d)
@@ -75,9 +75,9 @@ func (*Context) loadSystemKeyring(locations ...string) ([]string, error) {
 				log.Printf("%s has invalid extension (%s), skipping...", p, ext)
 			}
 		}
-		if len(ring) > 0 {
-			return ring, nil
-		}
+	}
+	if len(ring) > 0 {
+		return ring, nil
 	}
 	// Return an error since reading the system keyring is the last resort
 	return nil, errors.New("no suitable keyring directory found")
@@ -144,7 +144,7 @@ func (bc *Context) InitApkKeyring() (err error) {
 			}
 
 			// #nosec G306 -- apk keyring must be publicly readable
-			if err := os.WriteFile(filepath.Join(bc.WorkDir, element), data,
+			if err := os.WriteFile(filepath.Join(bc.WorkDir, "etc/apk/keys", filepath.Base(element)), data,
 				0644); err != nil {
 				return fmt.Errorf("failed to write apk key: %w", err)
 			}
@@ -190,7 +190,7 @@ func (bc *Context) InitApkWorld() error {
 func (bc *Context) FixateApkWorld() error {
 	log.Printf("synchronizing with desired apk world")
 
-	args := []string{"fix", "--root", bc.WorkDir, "--no-cache", "--update-cache"}
+	args := []string{"fix", "--root", bc.WorkDir, "--no-cache", "--update-cache", "--arch", bc.Arch.ToAPK()}
 	if bc.UseProot {
 		args = append(args, "--no-scripts")
 	}
--- a/pkg/build/oci/oci.go
+++ b/pkg/build/oci/oci.go
@@ -132,7 +132,7 @@ func publishTagFromImage(image v1.Image, imageRef string, hash v1.Hash) (name.Di
 		return name.Digest{}, fmt.Errorf("unable to parse reference: %w", err)
 	}
 
-	if err := remote.Write(imgRef, image, remote.WithAuthFromKeychain(kc)); err != nil {
+	if err := remote.Write(imgRef, image, remote.WithAuthFromKeychain(keychain)); err != nil {
 		return name.Digest{}, fmt.Errorf("failed to publish: %w", err)
 	}
 	return imgRef.Context().Digest(hash.String()), nil
--- a/pkg/build/oci/oci.go
+++ b/pkg/build/oci/oci.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
-	"runtime"
 	"time"
 
 	"chainguard.dev/apko/pkg/build/types"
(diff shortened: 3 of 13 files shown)


kaniini:
In Alpine parlance, I think all platforms would be:

- x86
- x86_64
- aarch64
- armv7
- ppc64le
- s390x
- riscv64

But other distributions support more or less platforms.  However, I think most will be using this with Alpine-based distributions for now, so defaulting to those 7 seems reasonable.

Maybe @rvs has some input here?  I think EVE is the biggest consumer of multiarch Alpine images.

imjasonh:
Hmm, this seems to reliably fail when building `examples/nginx.yaml` and `nginx-rootless.yaml`:

```
ERROR: unable to select packages:
  /bin/sh (virtual):
    provided by: busybox
    required by: nginx-1.20.2-r0[/bin/sh]
                 s6-2.11.0.1-r0[/bin/sh]
                 alpine-baselayout-3.2.0-r20[/bin/sh]
Error: failed to build layer image: failed to fixate apk world: failed to run /sbin/apk fix --root /tmp/apko-2109078717 --no-cache --update-cache --arch aarch64: exit status 4
```

If I move `amd64` to the end of `AllArchs`, it works for `amd64`/`aarch64`, but then fails for whatever's now second. Is there something else I have to do to sequester these builds from each other?

kaniini:
> Hmm, this seems to reliably fail when building `examples/nginx.yaml` and `nginx-rootless.yaml`:
> 
> ```
> ERROR: unable to select packages:
>   /bin/sh (virtual):
>     provided by: busybox
>     required by: nginx-1.20.2-r0[/bin/sh]
>                  s6-2.11.0.1-r0[/bin/sh]
>                  alpine-baselayout-3.2.0-r20[/bin/sh]
> Error: failed to build layer image: failed to fixate apk world: failed to run /sbin/apk fix --root /tmp/apko-2109078717 --no-cache --update-cache --arch aarch64: exit status 4
> ```
> 
> If I move `amd64` to the end of `AllArchs`, it works for `amd64`/`aarch64`, but then fails for whatever's now second. Is there something else I have to do to sequester these builds from each other?

Each of these need to be built in their own roots, but I'm not seeing any code to loop over the arch list and build images individually.

imjasonh:
> Each of these need to be built in their own roots, but I'm not seeing any code to loop over the arch list and build images individually.

Thanks, that was exactly it!

The code to iterate archs and build arch-specific images is hereish: https://github.com/chainguard-dev/apko/pull/56/files#diff-dd36f582aabe1c6547458d485dd6cc334bb6cdeba1de577a1d0670007efba268R109

The problem was I was passing through the same workDir for all archs 💥

imjasonh:
Hmm, this one I can't seem to reproduce myself. 🤔 

I can run `apko publish examples/nginx.yaml imjasonh/apko-multi` for all the archs just fine: https://gist.github.com/imjasonh/5c74e01862b1ce30a3d3430a8a58beb5

In both cases each `apk fix` happens in a different root, e.g., `working directory: /tmp/apko-2357205812/aarch64`

kaniini:
The tekton task needs to be updated to use proot.  In the meantime, we should restrict it to `--arch=x86_64`.

imjasonh:
> The tekton task needs to be updated to use proot. In the meantime, we should restrict it to `--arch=x86_64`.

This got CI to pass at least 🎉 

> Also, the proot command line needs to add the `-q /usr/bin/qemu-$ARCH` option when emulation is necessary.

I started down this rabbithole but I kinda think I should stop and do it in another change. The default `alpine:latest` base image doesn't seem to include `proot` by default, so there's more to do there, and this change is already getting kinda big.

kaniini:
> I started down this rabbithole but I kinda think I should stop and do it in another change. The default alpine:latest base image doesn't seem to include proot by default, so there's more to do there, and this change is already getting kinda big.

Agreed!  Lets follow it up outside of this MR :)

pkg/cli/build.go
```diff
@@ -66,6 +68,11 @@ func BuildCmd(ctx context.Context, imageRef string, outputTarGZ string, opts ...
 	if err != nil {
 		return err
 	}
+	arch := types.Architecture(runtime.GOARCH)
+	if len(bc.ImageConfiguration.Archs) != 0 {
+		log.Printf("WARNING: ignoring archs in config, only building for current arch (%s)", arch)
+	}
+	bc.Arch = arch
```
mattmoor, line 75: I guess `build` is where we'd use OCI layouts?

pkg/cli/publish.go
```diff
@@ -71,17 +78,53 @@ func PublishCmd(ctx context.Context, outputRefs string, opts ...build.Option) er
...
 
-	layerTarGZ, err := bc.BuildLayer()
-	if err != nil {
-		return fmt.Errorf("failed to build layer image: %w", err)
-	}
-	defer os.Remove(layerTarGZ)
+	var digest name.Digest
+	switch len(bc.ImageConfiguration.Archs) {
+	case 0:
+		return fmt.Errorf("no archs requested: %w", err)
```
mattmoor, line 93: What error is being wrapped here?

pkg/cli/publish.go
```diff
@@ -71,17 +78,53 @@ func PublishCmd(ctx context.Context, outputRefs string, opts ...build.Option) er
...
-	if err != nil {
-		return fmt.Errorf("failed to build OCI image: %w", err)
+		digest, _, err = oci.PublishImageFromLayer(layerTarGZ, bc.ImageConfiguration, bc.SourceDateEpoch, bc.Arch, bc.Tags...)
+		if err != nil {
+			return fmt.Errorf("failed to build OCI image: %w", err)
+		}
+	default:
+		var imgs []v1.Image
+		workDir := bc.WorkDir
+		for _, arch := range archs {
```
mattmoor, line 109: We should open an issue to track doing these in parallel (and track any hurdles that might have to clear)

pkg/build/oci/oci.go
```diff
@@ -119,45 +126,105 @@ func BuildImageTarballFromLayer(imageRef string, layerTarGZ string, outputTarGZ
...
-		authn.DefaultKeychain,
-		google.Keychain,
-		authn.NewKeychainFromHelper(ecr.NewECRHelper(ecr.WithLogOutput(ioutil.Discard))),
-		authn.NewKeychainFromHelper(credhelper.NewACRCredentialsHelper()),
-		github.Keychain,
-	)
+	return digest, v1Image, nil
+}
+
+func PublishIndex(imgs []v1.Image, tags ...string) (name.Digest, error) {
```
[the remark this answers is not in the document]
imjasonh: Good idea: https://github.com/chainguard-dev/apko/issues/74

config/task.yaml
```diff
@@ -44,6 +44,7 @@ spec:
       args:
       - publish
       - --image-refs=/tekton/results/IMAGES
+      - --arch=x86_64
```
mattmoor, line 47: Why not all?

mattmoor: I see: https://github.com/chainguard-dev/apko/pull/56#issuecomment-1061241005

Can we open an issue to tracking changing it to all?

pkg/cli/publish.go
imjasonh: Good catch! Fixed. Technically I think it should never even happen here, but just in case.

config/task.yaml
imjasonh: See above in the thread. Doing this for all archs requires some proot/qemu business that doesn't currently work. It's next.

pkg/cli/publish.go
kaniini:
```
2022/03/08 15:48:08 error during command execution: failed to build layer image for "386": failed to initialize apk database: failed to run /sbin/apk add --initdb --arch aarch64 --root /tmp/apko-4213008895/s390x: exit status 99
```

Why is it using s390x root for aarch64?  Is there an issue with the current code?

config/task.yaml
kaniini: (as an aside, it'd be nice if there was support for binfmt_misc user namespaces in Linux, then we could just use unshare and not bother with proot)

pkg/build/apk.go
```diff
@@ -190,7 +190,7 @@ func (bc *Context) InitApkWorld() error {
 func (bc *Context) FixateApkWorld() error {
 	log.Printf("synchronizing with desired apk world")
 
-	args := []string{"fix", "--root", bc.WorkDir, "--no-cache", "--update-cache"}
+	args := []string{"fix", "--root", bc.WorkDir, "--no-cache", "--update-cache", "--arch", bc.Arch.ToAPK()}
```
kaniini, line 193: `--arch` should not be needed here, as it should learn the arch from the `/etc/apk/arch` data.

pkg/build/apk.go
```diff
@@ -144,7 +144,7 @@ func (bc *Context) InitApkKeyring() (err error) {
 			}
 
 			// #nosec G306 -- apk keyring must be publicly readable
-			if err := os.WriteFile(filepath.Join(bc.WorkDir, element), data,
+			if err := os.WriteFile(filepath.Join(bc.WorkDir, "etc/apk/keys", filepath.Base(element)), data,
```
kaniini, line 147:
```suggestion
			if err := os.WriteFile(filepath.Join(bc.WorkDir, "etc", "apk", "keys", filepath.Base(element)), data,
```

kaniini:
Looks okay except for the two things I pointed out and also the question about s390x root being used for aarch64 arch.

pkg/cli/publish.go
panekj: You could let `apk` wait for lock with `--wait`

imjasonh: > Why is it using s390x root for aarch64? Is there an issue with the current code?

Good catch. I don't believe there's a bug in the current code. I'm pretty sure this was coming from how terribly lazy I was being in adding concurrency -- since each iteration of the loop sets `bc.Arch = arch`, this was shared between concurrent builds (and wouldn't be for sequential builds), which might have led to this bug.

In any case, I'll be sure to do a less lazy thing when we actually make these run concurrently.

kaniini: alright, i'll test it in a bit and merge it :)

PR #57, opened 2022-03-04
fix: replace separator with filepath.Join and fix home in user entries

kruskall:


pkg/build/accounts.go
```diff
@@ -43,7 +43,7 @@ func appendUser(users []passwd.UserEntry, user types.User) []passwd.UserEntry {
 		UserName: user.UserName,
 		UID:      user.UID,
 		GID:      user.GID,
-		HomeDir:  filepath.Join("/home", user.UserName),
+		HomeDir:  "/home/" + user.UserName,
```
kruskall, line 46: this is used inside `/etc/passwd` and it shouldn't use the host separator

PR #58, opened 2022-03-04
refactor: parallelize more tasks in BuildImage

kruskall:
also add missing error message for errors returned by `normalizeApkScriptsTar`

### DIFF
--- a/pkg/build/image_builder.go
+++ b/pkg/build/image_builder.go
@@ -75,15 +75,22 @@ func (bc *Context) BuildImage() error {
 		return errors.Wrap(err, "failed to fixate apk world")
 	}
 
-	if err := bc.normalizeApkScriptsTar(); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/build/image_builder.go
```diff
@@ -75,15 +75,22 @@ func (bc *Context) BuildImage() error {
 		return errors.Wrap(err, "failed to fixate apk world")
 	}
 
-	if err := bc.normalizeApkScriptsTar(); err != nil {
-		return err
-	}
+	eg.Go(func() error {
+		if err := bc.normalizeApkScriptsTar(); err != nil {
+			return errors.Wrapf(err, "failed to normalize scripts tar")
```
kaniini, line 80:
```suggestion
			return errors.Wrapf(err, "failed to normalize scripts.tar")
```

PR #60, opened 2022-03-04
refactor: improve logging when loading keyring

kruskall:


pkg/build/apk.go
```diff
@@ -54,13 +54,24 @@ func (*Context) loadSystemKeyring(locations ...string) ([]string, error) {
 	}
 	for _, d := range locations {
 		keyFiles, err := os.ReadDir(d)
-		if err != nil && !errors.Is(err, os.ErrNotExist) {
+
+		if errors.Is(err, os.ErrNotExist) {
+			log.Printf("%s doesn't exist, skipping...", d)
+			continue
```
kruskall, line 60: this wasn't returning early before but since len(nil) and for/range over nil are valid in Go there wasn't a nil pointer dereference

Issue #14, opened 2022-02-24
SBOM support


kaniini:
While apk-tools 3 will have support for generating SBOMs, we should add some basic SBOM generation by inspecting the APK database.  For this purpose we can probably recycle some code from Trivy.

PR #61, opened 2022-03-04
CycloneDX SBOM support

kaniini:
Add support for CycloneDX SBOM by inspecting the APK database.  APKv2 is only supported for now.

Closes: #14.


pkg/build/sbom.go
```diff
@@ -0,0 +1,180 @@
...
+func bomPurl(ns string, pkg *repository.Package) string {
+	return fmt.Sprintf("pkg:apk/%s/%s@%s", ns, pkg.Name, pkg.Version)
+}
+
+func (bc *Context) GenerateSBOM() error {
+	log.Printf("generating SBOM")
+
+	installedDB, err := os.Open(filepath.Join(bc.WorkDir, "lib", "apk", "db", "installed"))
+	if err != nil {
+		return errors.Wrap(err, "unable to open APK installed db")
```
mattmoor, line 78: nit: I think generally the way to wrap errors now is `fmt.Errorf("unable to open APK installed db: %w")`

pkg/build/sbom.go
```diff
@@ -0,0 +1,180 @@
...
+
+func bomRef(ns string, pkg *repository.Package) string {
+	return fmt.Sprintf("pkg:apk/%s/%s", ns, pkg.Name)
+}
+
+func bomPurl(ns string, pkg *repository.Package) string {
+	return fmt.Sprintf("pkg:apk/%s/%s@%s", ns, pkg.Name, pkg.Version)
+}
+
+func (bc *Context) GenerateSBOM() error {
```
mattmoor, line 73: In case it is useful, @imjasonh rolled the `ko` CycloneDX stuff here if you want to compare or copy useful bits: https://github.com/google/ko/pull/587/files#diff-ad63c642b90be2eed057117c368556d45c8c89a4c7310c948b80166fed73667aR41

pkg/build/sbom.go
kruskall: `Wrap` adds a stack trace at the point it was called, if that's not important then `fmt.Errorf` and go1.13+ errors are a suitable replacement

pkg/build/sbom.go
```diff
@@ -0,0 +1,180 @@
...
+	log.Printf("generating SBOM")
+
+	installedDB, err := os.Open(filepath.Join(bc.WorkDir, "lib", "apk", "db", "installed"))
```
kruskall, line 76: this is leaking, it needs a call to `defer installedDB.Close()` after the err check

pkg/build/sbom.go
imjasonh: Whether or not you roll your own, I think it was a good idea to have e2e validation of the SBOMs in CI: https://github.com/google/ko/blob/main/.github/workflows/sbom.yaml

imjasonh: Also: how do folks feel about generating both/all formats by default? That's something I'd like to do for ko, and apko being onboard would help motivate me. Seems weird to prefer one over the other(s)

pkg/build/sbom.go
mattmoor: FYI the upstream repo is archived in favor of the built-in errors: https://github.com/pkg/errors

Generally when I see folks wrapping errors they do it at ~every level, so the embedded stacktrace becomes somewhat redundant, but I can see the value and have no intent on dying on this hill :)

kruskall: I think it was archived/deprecated because of the go2 proposal (https://github.com/golang/go/issues/29934) :thinking:

That being said, I do prefer the error package in the standard library and I'm always happy to drop a dependency. I'll open a new PR since it seems that's the route forward :+1:

pkg/build/sbom.go
kaniini: I would be alright with generating the JSON-LD version of SPDX, but not the Turtle version presently generated by `ko`.

pkg/build/sbom.go
kaniini: It's not: https://gitlab.alpinelinux.org/alpine/go/-/blob/master/pkg/repository/apkindex.go#L22

pkg/build/sbom.go
kaniini: (I also think that shouldn't block this MR)

pkg/build/sbom.go
kruskall: oh, I missed that! Thanks :bow:

pkg/build/sbom.go
puerco: Absolutely, we should do both. I was holding on commenting on this one to talk on Monday (@kaniini happy to chat on the weekend too if you want). I think we need to align some things to work together, buts lets chat first!

kaniini: It can wait until Monday, I have some reasons for being opinionated here on SBOMs that I can outline outside this MR.

kaniini: After a discussion with @puerco, the plan is for him to take over this branch and add in the SPDX support and so on :)

puerco:
OK, I've reworked the sbom code to its own package. It now supports pluggable sbom generators which can be turned on and off from the options. The cyclonedx code is the first one of these :partying_face: 

I also added tests for most of the functions but I still have to ensure the output looks as expected, also I have to actually test it but the main idea should be ready for review :)

Lets sync tomorrow to talk about some next steps here (like, where should the SBOM go? matching the cyclonedx and spdx structure, etc).

kaniini:
I went ahead and merged this, we can talk more about it later today for the SPDX support :)

PR #70, opened 2022-03-07
style: Prefer if-err style wherever possible

imjasonh:
By keeping errs inside the scope of an if check, fewer errant errs end
up escaping to other calls, which reduces inadvertant shadowing, and
can make it clearer to readers that the err is only considerd during
that if check.

---

If y'all don't prefer this form, that's totally fine. I won't take it personally.


dlorenc (approved):
Nice

panekj (approved):

Issue #71, opened 2022-03-08
failed to mutate accounts: failed to open /tmp/apko-1634229632/etc/passwd


fwaggle:
I upgraded from commit ba2bc74acd8e6cf356080d45707845409c90aded to 7a873c6937b72bf41a43847e97ae82cbccf66ddf and suddenly my container build doesn't work any more, with the above error message. Here's the config file I'm using:

```
contents:
  repositories:
    - https://dl-cdn.alpinelinux.org/alpine/edge/main
    - https://dl-cdn.alpinelinux.org/alpine/edge/community
  keyring:
    - /etc/apk/keys/alpine-devel@lists.alpinelinux.org-4a6a0840.rsa.pub
    - /etc/apk/keys/alpine-devel@lists.alpinelinux.org-5243ef4b.rsa.pub
    - /etc/apk/keys/alpine-devel@lists.alpinelinux.org-5261cecb.rsa.pub
    - /etc/apk/keys/alpine-devel@lists.alpinelinux.org-6165ee59.rsa.pub
    - /etc/apk/keys/alpine-devel@lists.alpinelinux.org-61666e3f.rsa.pub
  packages:
    - collectd
    - collectd-snmp
    - collectd-rrdtool
    - collectd-sensors
    - collectd-network
    - collectd-curl
    - lm_sensors

entrypoint:
  command: /usr/sbin/collectd -f
```

It seems that at some point, I'm guessing the UID stuff, added mutations for /etc/passwd. I'll spare the whole log, but here's the relevant output:

```
Error: failed to build layer image: failed to mutate accounts: failed to open /tmp/apko-1634229632/etc/passwd: open /tmp/apko-1634229632/etc/passwd: no such file or directory
2022/03/07 23:59:38 error during command execution: failed to build layer image: failed to mutate accounts: failed to open /tmp/apko-1634229632/etc/passwd: open /tmp/apko-1634229632/etc/passwd: no such file or directory
```

Looks like it's caused by one of the above depending on `bash` (indeed, a minimal reproduction is a container consisting solely of bash), with one of the post-install scripts trying to alter /etc/passwd without it existing.

It's fixable by adding `alpine-base` to the packages, presumably because it has a minimal /etc/passwd in it, but it worked without that on a version of `apko` from February, and I'm not even sure why it worked before? If it's not intended to work feel free to just close this and I'll add `alpine-base` to my container config.

kaniini:
[...]

PR #72, opened 2022-03-08
refactor: skip user/group mutation if there are no accounts

kruskall:
Related https://github.com/chainguard-dev/apko/issues/71  

This should fix that particular scenario but the error can still be reproduced by adding a UID/GID.

### DIFF
--- a/pkg/build/accounts.go
+++ b/pkg/build/accounts.go
@@ -56,45 +56,49 @@ func (bc *Context) MutateAccounts() error {
 
 	var eg errgroup.Group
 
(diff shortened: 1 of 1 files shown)


kaniini:
We should probably have it print a warning if `/etc/passwd` is not present in the target image though.  Many applications will expect it to be present.

kruskall:
> We should probably have it print a warning if `/etc/passwd` is not present in the target image though. Many applications will expect it to be present.

Good point! I'll open a new PR, I also have a proper fix for the linked issue

PR #75, opened 2022-03-08
feat: introduce build assertions and validate passwd and group files

kruskall:
Add support for assertions to check the result of the build context.  

As an example, two assertions are added to check whether `/etc/passwd` and `/etc/group` are (optionally) present.

### DIFF
--- a/pkg/build/image_builder.go
+++ b/pkg/build/image_builder.go
@@ -20,7 +20,9 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sync"
 
+	"github.com/hashicorp/go-multierror"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/build/image_builder.go
```diff
@@ -88,6 +88,17 @@ func (bc *Context) BuildImage() error {
 		return err
 	}
 
+	for _, a := range bc.Assertions {
+		a := a
+		eg.Go(func() error {
```
imjasonh, line 93: If there are multiple unmet assertions, checking them like this will cause the execution to be random, and the resulting `Wait` will just surface whichever one failed first.

Maybe it's worth doing them in sequence and collecting the failures, so you don't end up in that situation where 20 things are wrong and you have to fix them each one-by-one since you don't know how many others are failing.

(You could still execute them in parallel, but collecting the errors gets uglier since you can't use errgroup)

kruskall: Good point! This looks like a good use case for multierrors (https://github.com/hashicorp/go-multierror) since assertions are usually not related to each other and they don't mutate the build context.

I'm not sure what the policy of this project is on adding a new dependency.

kaniini: As long as it is under an Apache 2.0 compatible license, it is fine to add as a dependency.

kaniini:
This looks really nice!  Once you introduce the multierrors stuff, I think we can merge it.

pkg/build/image_builder.go
kruskall: Added! Thanks everyone!

PR #80, opened 2022-03-09
refactor tarball for Melange

kaniini:
This makes the tarball building package more generic, allowing it to be reused in Melange.

### DIFF
--- a/pkg/tarball/tarball.go
+++ b/pkg/tarball/tarball.go
@@ -25,8 +25,12 @@ import (
 	"time"
 )
 
+type Context struct {
+	SourceDateEpoch time.Time
+}
+
 // Writes a raw TAR archive to out, given an fs.FS.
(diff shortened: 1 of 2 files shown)


pkg/tarball/tarball.go
```diff
@@ -25,8 +25,35 @@ import (
 	"time"
 )
 
+type Context struct {
+	SourceDateEpoch time.Time
+}
+
+type Option func(*Context) error
+
+// Generates a new context from a set of options.
```
jspeed-meyers, line 34: I'm not going to approve because I'm too much of a newb to both Go and this project, but I thought I should point out (as a totally optional point) that the Godoc tool and Go documentation convention for a function is to start sentences with the function's name: `NewContext generates a new context from a set of options.` Thought this might help with documentation later. If this isn't helpful, just ignore. [Relevant official Go blog post on Godoc and documentation](https://go.dev/blog/godoc).

PR #82, opened 2022-03-09
Include arm v6 and v7 in multiarch support

imjasonh:
This turned out to be a little more complicated than I wanted, since the image's config doesn't include the CPU variant. So instead, I just pass through a `map[types.Architecture]v1.Image` to `PublishIndex` (which I also don't super love).

Ideas welcome. I'm also fine just punting on this if we think we'll come up with a better idea later. I don't really think it needs to block anything.

edit: An example image built with this change:
```
index.docker.io/imjasonh/alpine-multi@sha256:735bfedcd7f5e4c7a4655a7d8d23c218e837aa0290c045019270a61fcc0e184a
```

### DIFF
--- a/pkg/build/oci/oci.go
+++ b/pkg/build/oci/oci.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"sort"
 	"time"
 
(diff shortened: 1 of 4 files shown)


kaniini:
It seemed okay for now, so I merged it.  We can revisit it later.

PR #89, opened 2022-03-10
fix handling of https key paths

AmineChikhaoui:
using `uri.New()` on a https url returns paths such as
file:///https:/alpinelinux.org/keys/foo.pub

### DIFF
--- a/pkg/build/apk.go
+++ b/pkg/build/apk.go
@@ -111,7 +111,12 @@ func (bc *Context) InitApkKeyring() (err error) {
 			// Normalize the element as a URI, so that local paths
 			// are translated into file:// URLs, allowing them to be parsed
 			// into a url.URL{}.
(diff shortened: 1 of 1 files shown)


pkg/build/apk.go
```diff
@@ -111,7 +111,12 @@ func (bc *Context) InitApkKeyring() (err error) {
 			// Normalize the element as a URI, so that local paths
 			// are translated into file:// URLs, allowing them to be parsed
 			// into a url.URL{}.
-			asURI := uri.New(element)
+			var asURI uri.URI
+			if ok := strings.HasPrefix(element, "https://"); ok {
```
imjasonh, line 115:
```suggestion
			if strings.HasPrefix(element, "https://") {
```

(the `ok` doesn't really add much)

kaniini (approved):

PR #90, opened 2022-03-10
Build multiarch images concurrently

### LINKED ISSUE #76: Investigate parallel multiarch builds
https://github.com/chainguard-dev/apko/pull/56 added multiarch builds, which runs the necessary apk commands to build a tarball for each arch, then stitches them together into a multiplatform image index.

It currently does so sequentially, and should be noticeably faster if it could do it in parallel.

Simply naively wrapping the sequential bits in an `errgroup.Group` didn't seem to work: https://github.com/chainguard-dev/apko/pull/56#discussion_r821803442 -- there might be something else we can do to bypass this for faster multiarch builds

imjasonh:
This makes a no-op multiarch `apko publish` go from ~17s to ~5s, at the cost of more confusing interleaved logs ([example](https://github.com/chainguard-dev/apko/files/8224089/err.txt))

Other than that, things seem to have worked just fine. 🎉  The bug described in [#56](https://github.com/chainguard-dev/apko/pull/56#discussion_r821803442) seems to have been just improper isolation of build contexts, which is solved here with a `DeepCopy`.

If we want to make the logs better we could either keep them interleaved with some log prefix, or better yet buffer them and make sure they all get logged together at once. I'm happy to go either way here, or punt to later if this seems okay as is.

Fixes #76


panekj:
Why `DeepCopy` is needed? You should be able to make a copy just by assignement

kaniini:
It'd be nice to have a flag to disable concurrency, like if you are trying to debug.

kaniini:
Yes, sounds good.  Maybe we can also prefix the logs with the specific job name, say the architecture being built too.

imjasonh:
Sorry, I haven't had a ton of time to come back and work on this.

I tried to make logging a little better with prefixes, but it ended up being a more expansive change -- having ` log.Logger` in the `build.Context` struct, and replacing `log.Foo` with `bc.logger.Foo` everywhere, so we could set its prefix per-arch. This was... okay, but it'll probably become a rebase-conflict landmine unless I have time to make sure this gets in quickly. Which I don't right now. And it didn't help to isolate the `apk` output, since the prefix was only logged at the first line.

I'll close this for now, since it doesn't sound like it's a high priority for v0.2. If someone else wants to pick this up I won't feel bad. If not, I'll come back when I have some more time to dedicate to it.

PR #93, opened 2022-03-11
feat: add support for chaining tarball context

kruskall:
This should help in melange


pkg/tarball/write.go
```diff
@@ -0,0 +1,106 @@
...
+				return err
+			}
+		}
+
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	tw.Flush()
```
kruskall, line 94: note to self: remove this

kaniini:
Looks cool.  I think I see how we can use this to clean up the tarstreams in Melange!

kruskall:
Ok, this should be ready for review

kaniini (approved):

PR #94, opened 2022-03-11
SBOM Support!

puerco:
This PR finally exposes the SBOM functionality in `apko build` and completes some missing bits in the SBOM code.

It introduces three new flags to control SBOM generation:

```
      --sbom                   generate SBOMs (default true)
      --sbom-formats strings   SBOM formats to output (default [spdx,cyclonedx])
      --sbom-path string       generate SBOMs in dir (defaults to image directory)
```
All of these flags are optional. The default behaviour is to write the pair of sboms in both formats in the same directory of the image tarball. 

There are other changes in the PR geared towards completing the information in the SBOMs:

1. SBOM generation has been moved after the layer is written as the digest is needed to generate purls
2. Bumped the alpine/go module to add the missing package information to the sboms
3.   Added remaining information like architecture, layer digest, and other data

I think we need to do some adjustments to the SBOM structure because we are not describing the image, only the layer. But let's handle this in a follow-up.


kaniini:
Did you also update `apko publish` to support generating SBOMs? :)

pkg/sbom/generator/spdx/spdx.go
```diff
@@ -62,21 +63,52 @@ func (sx *SPDX) Generate(opts *options.Options, path string) error {
 		Relationships: []Relationship{},
 	}
 
+	mainPkgID := "SPDXRef-Package-apko-os-layer-" + uuid.NewString()
```
imjasonh, line 66: Does this mean the SBOM will be different every time it's generated?

pkg/sbom/generator/spdx/spdx.go
```diff
@@ -62,21 +63,52 @@ func (sx *SPDX) Generate(opts *options.Options, path string) error {
...
+	if opts.ImageInfo.Reference != "" {
+		x := ""
+		if !strings.Contains(opts.ImageInfo.Reference, "/") {
+			x = "index.docker.io/library/"
+		}
+		mainPkgID = fmt.Sprintf("SPDXRef-%s%s", x, opts.ImageInfo.Reference)
+	}
+
+	// Main ackage purl
```
imjasonh, line 75: typo: package

Issue #66, opened 2022-03-07
Support for one-layer-per-package?


sipsma:
It seems like right now the image exported by apko is just a single layer. Is it in apko's scope to eventually support multi-layer images? I specifically am imagining that each package that gets installed (including each transitive dependency) would become its own single layer. The exported image would then just consist of the layers for each package, topologically sorted based on the dependency DAG.

This type of approach would be beneficial for caching in many ways, one example being that if you pushed an image built by apko to a registry and the registry already has some or all of the layers, you could skip pushing them.

kaniini:
This is really not so easy to do, but patches would be accepted if somebody wants to try it.

sipsma:
Cool, I will likely not have time in the immediate future, but I may give it a shot sometime if no one else has. Good to know it's in scope either way, thanks.

chainguardian:
Adding issue to global tracking board

PR #99, opened 2022-03-12
WIP support for multi-layered images.

sipsma:
Signed-off-by: Erik Sipsma <erik@sipsma.dev>

WIP fix to #66

The code already technically works when using the `build` command for simple cases, try it out with `apko build --multilayer test.yaml test output.tar` using this test.yaml for instance:
```
contents:
  repositories:
    - https://dl-cdn.alpinelinux.org/alpine/edge/main
  packages:
    - alpine-baselayout
    - curl
    - rsync
    - coreutils
```

On running `docker load <output.tar` you'll see multiple layers loaded, 1 for each package in the DAG.

However, still big caveats:
1. Walking the package DAG has some hacks which probably don't always work, need guidance on how to deal with corner cases there
1. Doesn't support supervision tree, custom keyfiles, sboms, etc. 
1. Just uses downloaded apks directly as layer tar.gzs, which does work but results in `.PKGINFO` and `.SIGN.RSA*` files sticking around in the exported image
1. Doesn't support `publish` yet (this is probably easy though)
1. Various other issues marked with `TODO` comments in the code

---

@kaniini  Opening as a draft so I can get guidance on some questions related to apk before cleaning up the rest. A few big ones:
1. When there are multiple alternatives for a package, how do we disambiguate? i.e. I running `apk info so:libudev.so.1` shows multiple implementations. Right now the code just picks the first package it finds, which feels likely to not always be correct? Obviously if the user specifically requests one version of the package, choose that one, but what about when the user doesn't request any one in particular? Is the default choice in the apk metadata somewhere?
2. If the user provides multiple repositories, is it right to check each one for a requested package in order and use the package from the first repository its found in? Or is there some other logic that should be followed.
3. Is there anything special about the apk keyring that would require special logic to use the keys to download the packages? Or is it just a matter of telling the Go http client to use them. I haven't actually looked into this yet, so apologies if it's a dumb question.

There are some other TODOs in the apk DAG code too and I'm sure there's plenty of "unknown unknowns" I've overlooked too...

Also, let me know if the package DAG parsing code would fit better upsteam in the the `alpine/go` package. Happy to put stuff there if you think it fits better.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -38,6 +38,7 @@ type Context struct {
 	SBOMFormats        []string
 	ExtraKeyFiles      []string
 	Arch               types.Architecture
+	Multilayer         bool
 }
 
 func (bc *Context) Summarize() {
@@ -48,6 +49,7 @@ func (bc *Context) Summarize() {
 	log.Printf("  source date: %s", bc.SourceDateEpoch)
 	log.Printf("  SBOM output path: %s", bc.SBOMPath)
 	log.Printf("  arch: %v", bc.Arch.ToAPK())
+	log.Printf("  multilayer: %t", bc.Multilayer)
 	bc.ImageConfiguration.Summarize()
 }
 
@@ -99,6 +101,21 @@ func (bc *Context) BuildLayer() (string, error) {
 	}
 
 	return layerTarGZ, nil
+}
+
+// TODO: integrate with BuildLayer
+func (bc *Context) BuildMultilayer() ([]string, error) {
+	bc.Summarize()
+
+	// build image filesystem
+	layers, err := bc.BuildMultilayerImage()
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: support SBOM
+
+	return layers, nil
 }
 
 // New creates a build context.
@@ -244,6 +261,13 @@ func WithImageConfiguration(ic types.ImageConfiguration) Option {
 func WithArch(arch types.Architecture) Option {
 	return func(bc *Context) error {
 		bc.Arch = arch
+		return nil
+	}
+}
+
+func WithMultilayer(multilayer bool) Option {
+	return func(bc *Context) error {
+		bc.Multilayer = multilayer
 		return nil
 	}
 }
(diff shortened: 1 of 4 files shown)


sipsma:
I also just found the solver that it seems like apk might use? https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/solver.c

If that is indeed the case, it seems like there's probably a lot of caveats and cases my current code is not handling (more than I even thought originally).

If it's in scope to port that logic in `solver.c` to `alpine/go` then I'd be happy to help out with that. I guess CGo is a possibility too (but not sure if you want apko to go that route). Or maybe there's a different/better way to approach this, like if there was a way to shell out to `apk` to get the information on each versioned package in the dependency DAG rather than trying to do it as a library call. Let me know if any of these approaches sound better to you.

kaniini:
I think it would be better to write a Go client for APKv2 and APKv3 repositories directly instead of using libapk.  Originally I was going to use cgo, but I find the libapk API difficult to reflect in an idiomatic Go way.

First step would be to write a conformant solver, and then parsers for APKv2 and APKv3 packages.  There is already code to generate APKv2 packages in Melange.

kaniini:
(right now my focus is to get some things cooking for demos, but if you want to go further down this rabbit hole, i’m sure that the apk maintainers (of which I am one) will be glad to review apk-specific packages in alpine-go.)

sipsma:
Awesome, thanks for the pointers @kaniini! I started taking a look at porting the solver over to alpine-go; initial impression is that while it's not trivial it's definitely feasible, so I will give it a shot. Sent out a small pre-requisite change here: https://gitlab.alpinelinux.org/alpine/go/-/merge_requests/5

Will probably take a bit to get the solver ported over and used by the code here, but I'll just leave this PR open as a draft in the meantime.

sipsma:
@kaniini Hello, just to update on this, I made some limited initial progress on porting the apk solver but then started a new job and since then this has fallen far back in my TODO list unfortunately. I am still very interested in this feature in general but it's unlikely I'll have time to work on it soon, so I'm just going to close this for now rather than let it languish indefinitely.

If priorities change and I have time I'll check back in, but appreciate your pointers on it either way and glad I got to learn some more about apk if nothing else :-)

PR #100, opened 2022-03-13
fix: readd skipclose option

kruskall:
got lost in the previous PR, my apologies.

On a side note: should this and other APK specific code be moved to melange ? Right now the tarball package serves two different purposes: generating OCI layer tarball and generating APK packages.


kaniini:
I think we keep it where it is for now, unless there's a easy eay to refactor it.

PR #103, opened 2022-03-13
test: use testify consistently

kruskall:
tests are using `go-cmp`, `testify` and `t.Error`.

This standardize the assertions to use testify and forbids the other two

```release-note
-Test assertions are now standardized the to use testify
```

### DIFF
--- a/pkg/sbom/generator/spdx/spdx_test.go
+++ b/pkg/sbom/generator/spdx/spdx_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"chainguard.dev/apko/pkg/sbom/options"
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 	"gitlab.alpinelinux.org/alpine/go/pkg/repository"
 	"sigs.k8s.io/release-utils/command"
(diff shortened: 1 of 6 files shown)


pkg/sbom/generator/spdx/spdx_test.go
```diff
@@ -79,8 +78,7 @@ func TestReproducible(t *testing.T) {
 		require.NoError(t, err)
 		d = append(d, data)
 	}
-	diff := cmp.Diff(d[0], d[1])
-	require.Empty(t, diff)
+	require.Equal(t, d[0], d[1])
```
puerco, line 81:
```suggestion
	diff := cmp.Diff(d[0], d[1])
	require.Empty(t, diff, fmt.Sprintf("difference in expected output %s", diff))
```

puerco (changes requested):
Hey @kruskall thanks for the PR! I also prefer using testify in general but I would not go so far as to ban `google/go-cmp` completely. It has some useful methods which are useful and we can still leverage along with testify (see bellow comment).

Let's not ban outright but ensure in further reviews new tests use testify.

pkg/sbom/generator/spdx/spdx_test.go
kruskall: I'm probably missing something here, testify should already print the diff between "expected" and "got". I tried failing the test on purpose and the only difference I could see was `go-cmp` omitting the identical lines. Did I miss something ?

kruskall:
Hi @puerco :wave:, Thank you for the feedback, I've reverted the last change!

puerco (approved):
Awesome, thank you @kruskall !!

PR #118, opened 2022-03-17
support qemu-binfmt emulation with proot

### LINKED ISSUE #81: Multiarch followup
There's a few remaining tasks to be done for the Multiarch support.

Namely, `proot` needs to be integrated, however, it should not run qemu when building on the native CPU, as that would just slow down native builds for no reason.

There's also no need to use `--arch` with `apk fix`.

kaniini:
Needed for multiarch support, closes #81.

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -33,6 +33,7 @@ import (
 func Build() *cobra.Command {
 	var useProot bool
 	var buildDate string
+	var buildArch string
 	var writeSBOM bool
 	var sbomPath string
 	var sbomFormats []string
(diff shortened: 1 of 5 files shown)


pkg/cli/build.go
```diff
@@ -68,6 +69,7 @@ bill of materials) describing the image contents.
 				build.WithExtraKeys(extraKeys),
 				build.WithTags(args[1]),
 				build.WithExtraRepos(extraRepos),
+				build.WithArch(types.Architecture(buildArch)),
```
imjasonh, line 72: Is there any sense in pulling out and reusing some of [`types.ParseArchitectures`](https://github.com/chainguard-dev/apko/blob/9f2503770798549dcb8f05ca319b3d7e087104ef/pkg/build/types/types.go#L122) so callers can use the apk-style arch strings? e.g., `--build-arch=x86_64`

kaniini: Yeah, go ahead if you want to refactor it :)

PR #122, opened 2022-03-17
refactor multiarch, add parallelization, improve logging

kaniini:
Various multiarch refactorings needed to make it more robust.

Closes: #76

CC @imjasonh, @puerco


puerco (approved):
Awesome! I suggest we yank out the builder function. Let me handle it in a follow-up.

PR #125, opened 2022-03-19
feat: Add apko login

imjasonh:
This uses code exported from `crane auth login`, also used by `ko
login`, to simulate `docker login`, without requiring Docker to be
installed.


dlorenc:
I love that this is two lines of code.

Issue #136, opened 2022-03-21
Set up `PATH`


mattmoor:
While attempting to use `ghcr.io/distroless/busybox` I noticed that `busybox` symlinks in places typically on `PATH` weren't being resolved:
```
/bin/sh: cp: not found
```

The OG distroless images set the following variables, which we seem to not have:
```
    "Env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/busybox",
      "SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt"
    ],
```

PR #137, opened 2022-03-21
Dogfood the apko-built alpine-base image

mattmoor:


mattmoor:
@kaniini I'm guessing that this is the same `PATH` problem here: https://github.com/chainguard-dev/apko/issues/136

kaniini:
Yes, probably.

kaniini:
Works with refreshed alpine-base image, merging.

PR #138, opened 2022-03-21
Set the `mediaType` on the index.

mattmoor:
Unfortunately `empty.Image` defaults to setting the Docker "schema2" media type, but `empty.Index` defaults to no media type.  This is unfortunate because Docker requires media types, so the absence of a media type is interpreted as OCI, and gcr.io doesn't allow mixing OCI index types with Docker image types.

When using `ghcr.io/distroless/static` with `ko` multi-arch I see:
```
MANIFEST_INVALID: Failed to parse manifest for request "/v2/.../manifests/latest": Image index child manifest not supported: application/vnd.docker.distribution.manifest.v2+json.
```


mattmoor:
@kaniini we should pull this into 0.2.2

cpanato (approved):

PR #139, opened 2022-03-21
Have Tekton produce multi-arch by default

mattmoor:


kaniini:
This will fail because the qemu-binfmt needs to be set up.  See [the distroless apko building action](https://github.com/distroless/actions/blob/main/apko-snapshot/action.yaml#L39-L40) for how to do it in Github actions.

mattmoor:
@kaniini works 😁

PR #141, opened 2022-03-26
bugfix: archive hardlinks

kaniini:
When building images with hardlinks, we previously copied the file multiple times.  This deduplicates the hardlinks, bringing [this git-init image](https://github.com/mattmoor/git-init) down to 15MiB instead of hundreds.

### DIFF
--- a/pkg/tarball/write.go
+++ b/pkg/tarball/write.go
@@ -43,15 +43,35 @@ func (ctx *Context) writeArchiveFromFS(dst io.Writer, fsys fs.FS) error {
 }
 
 func hasHardlinks(fi fs.FileInfo) bool {
-	return fi.Sys().(*syscall.Stat_t).Nlink > 1
+	if stat := fi.Sys(); stat != nil {
+		si := stat.(*syscall.Stat_t)
+
+		// if we don't have inodes, we just assume the filesystem
+		// does not support hardlinks
+		if si == nil {
+			return false
+		}
+
+		return si.Nlink > 1
+	}
+
+	return false
 }
 
 func getInodeFromFileInfo(fi fs.FileInfo) (uint64, error) {
-	stat := fi.Sys().(*syscall.Stat_t)
-	if stat == nil {
-		return 0, fmt.Errorf("unable to stat underlying file")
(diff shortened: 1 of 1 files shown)


pkg/tarball/write.go
```diff
@@ -94,6 +129,21 @@ func (ctx *Context) writeTar(tw *tar.Writer, fsys fs.FS) error {
 			header.Gname = ctx.OverrideGname
 		}
 
+		if !info.IsDir() && hasHardlinks(info) {
+			inode, err := getInodeFromFileInfo(info)
+			if err != nil {
+				return err
+			}
```
mattmoor, line 136: So the first occurrence of an inode we see in the `fs.WalkDir` becomes the source of truth and all subsequent visits become hard links to it, and this is made deterministic by `fs.WalkDir` being deterministic, right?

mattmoor (approved):
Thanks for the quick fix 🥳

pkg/tarball/write.go
puerco: Seems good to me, WalkDir [depends on the underlying FS implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/io/fs/walk.go;l=74-91;drc=refs%2Ftags%2Fgo1.18) if we notice inconsistencies we can override ReadDir in apko/fs and make the return order deterministic.

pkg/tarball/write.go
```diff
@@ -41,7 +42,41 @@ func (ctx *Context) writeArchiveFromFS(dst io.Writer, fsys fs.FS) error {
 	return ctx.writeTar(tw, fsys)
 }
 
+func hasHardlinks(fi fs.FileInfo) bool {
```
puerco, line 45: Expanding on my other comment, I'm wondering if we should have these functions live in a FileInfo type in apko/fs and to keep all the file logic constrained to that package.

pkg/tarball/write.go
kaniini: yes

pkg/tarball/write.go
kaniini: yes feel free to refactor this is just a hotfix :)

Issue #145, opened 2022-03-27
Publish SBOMs


mattmoor:
We should publish SBOMs by default for `apko`.

The logic to create a `v1.Image` with an `sbom` attachment is here: https://github.com/google/ko/blob/aafd3e863b40f08e941d991562d74e68e883b1f1/pkg/build/gobuild.go#L810-L821

Here's the logic where we publish optionally attached SBOMs: https://github.com/google/ko/blob/6d0691330a6df9407a1232e8e8835c6860387b22/pkg/publish/default.go#L129-L157

cc @puerco

mattmoor:
Gonna play around with this 😎

mattmoor:
I have something rough working e2e, will try to stage the third piece of this with e2e tests once I'm back from the park 😎

PR #146, opened 2022-03-27
Start to use Cosign's `oci.Signed*` types.

mattmoor:
This is a precursor to starting to attach things to them, such as SBOMs and signatures, which we can access as we publish them.

I have also sprinkled several `TODO(#145)` around where we will need to attach SBOMs and publish SBOMs.

Related: #145


mattmoor:
`oci.SignedImage` and `oci.SignedImageIndex` are extensions of `v1.Image` and `v1.ImageIndex` that include cosign's extensions: https://github.com/sigstore/cosign/blob/34d08380a96506f3353e86a99f3403bba6978c9e/pkg/oci/image.go#L20-L25

puerco (approved):
Nice!

kaniini (approved):

PR #147, opened 2022-03-27
[WIP] Add logic to publish attached SBOMs when writing images and indices.

mattmoor:
This is heavily based on the same logic in `ko`, which publishes SBOMs that are attached to the `oci.SignedEntity` during the build process.

Related: https://github.com/chainguard-dev/apko/issues/145

Based on: https://github.com/chainguard-dev/apko/pull/146


mattmoor:
This should be RFAL.

kaniini (approved):

PR #149, opened 2022-03-27
Attach SBOMs to built images.

mattmoor:
This change attaches SBOMs to the `oci.SignedImage`s we build up, and these are published thanks to the previous commit in this series.

I have added validation that for each of the examples we have that we scan download its SBOM with cosign.

Fixes: https://github.com/chainguard-dev/apko/issues/145

Based on https://github.com/chainguard-dev/apko/pull/147


pkg/build/oci/oci.go
```diff
@@ -119,14 +129,50 @@ func buildImageFromLayer(layerTarGZ string, ic types.ImageConfiguration, created
...
-	return v1Image, nil
+	si := signed.Image(v1Image)
+
+	// Attach the SBOM, e.g.
+	if len(sbomFormats) > 0 {
+		var mt ggcrtypes.MediaType
+		var path string
+		switch sbomFormats[0] {
+		case "spdx":
+			mt = ctypes.SPDXMediaType
```
mattmoor, line 143: I think this media type is intended for non-JSON sboms (like `ko` emits), and `apko` emits JSON sboms, so we should sort out the most appropriate media type to use here.

pkg/build/oci/oci.go
```diff
@@ -119,14 +129,50 @@ func buildImageFromLayer(layerTarGZ string, ic types.ImageConfiguration, created
...
 	}
 
-	return v1Image, nil
+	si := signed.Image(v1Image)
+
+	// Attach the SBOM, e.g.
+	if len(sbomFormats) > 0 {
+		var mt ggcrtypes.MediaType
+		var path string
+		switch sbomFormats[0] {
```
mattmoor, line 141: Currently cosign's method for associating SBOMs supports a single type of SBOM, so I simply pick the first when multiple are emitted.  This can be controlled by restricting or reordering the sbom formats.

pkg/build/oci/oci.go
```diff
@@ -119,14 +129,50 @@ func buildImageFromLayer(layerTarGZ string, ic types.ImageConfiguration, created
...
+		case "spdx":
+			mt = ctypes.SPDXMediaType
+			path = filepath.Join(sbomPath, fmt.Sprintf("sbom-%s.spdx.json", arch.ToAPK()))
```
mattmoor, line 144: I wish the SBOM metadata were surfaced in a way we could plumb through to here, but the file list is effectively discarded today.  We should consider ways to improve the way this is plumbed through.

pkg/build/oci/oci.go
mattmoor: cc @dlorenc

mattmoor: I see cosign now has: `SPDXJSONMediaType`, I'll update to that.

pkg/build/oci/oci.go
```diff
@@ -130,21 +134,45 @@ func buildImageFromLayer(layerTarGZ string, ic types.ImageConfiguration, created
...
+		case "cyclonedx":
+			mt = ctypes.CycloneDXJSONMediaType
+			path = filepath.Join(sbomPath, fmt.Sprintf("sbom-%s.cdx", arch.ToAPK()))
+		default:
+			return nil, fmt.Errorf("unsupported SBOM format: %s", sbomFormats[0])
+		}
+		if len(sbomFormats) > 1 {
+			// When we have multiple formats, warn that we're picking the first.
+			logger.Printf("uploading SBOM with media type: %s", mt)
```
vaikas, line 153: nit: I'd change this to actually be a warning, or rephrase the message if you want to keep it as a Printf.

.github/workflows/mink-e2e.yaml
```diff
@@ -56,9 +58,16 @@ jobs:
     - name: Build examples
       run: |
         for cfg in $(find ./examples/ -name '*.yaml'); do
+          echo "::group:: $cfg"
           DIGEST=$(mink run task apko -- --path=$cfg)
 
           echo $cfg produced: $DIGEST
+
+          # Each of the per-architecture images produces should have an SBOM attached to it.
```
vaikas, line 66: super nit: 'images produced'?

vaikas (approved):

kaniini:
Looks fine to me, given that we can't just fix cosign immediately.

PR #150, opened 2022-03-28
Refactor build into mockable implementation, spin apk package

puerco:
This PR introduces a big refactor that does two main things:

* The first is to refactor the build package into a mockable implementation. The goal is to reduce the API surface and to make each function more unit testable. Code is mostly the same, but has been moved around to expose less functions and get the internal function interfaces somewhat similar to feed into unit tests.

* The second is that it spins the apk code into its own module which is also built around the pattern of a mockable implementation.

`Options` is now out on its own too. The idea is to have a single options struct that we can share among both packages (`apk` and `build`). It also scaffolds all the new mocks and adds new tests to test the logical paths in the modules. I have added the first new unit tests for each of the implementation functions and will work on the week on more.

I left pending the refactor of the image code (functions called by `build.BuildImage()`). We can choose to cut out a new package or keep it inside of `build` and create a new mockable interface for it. Either way I will do those in a follow-up.


puerco:
This is how test coverage improved after the refactor.

**Before:**
```
ok  	chainguard.dev/apko/pkg/build	0.002s	coverage: 4.5% of statements
```
**After:**
```
ok  	chainguard.dev/apko/pkg/apk	(cached)	coverage: 56.1% of statements
ok  	chainguard.dev/apko/pkg/build	(cached)	coverage: 17.4% of statements
```

kaniini:
This looks amazing!  We should probably separate the image FS building into its own package, but we can probably land this.

puerco:
Sounds good, I just wanted to restructure this part before we start adding new features this week :)

PR #153, opened 2022-03-28
tarball: use pargzip to compress images

kaniini:
using pargzip improves compression performance by several orders of magnitude,
for example the git-init image takes less than 1 second to compress now.

Signed-off-by: Ariadne Conill <ariadne@chainguard.dev>


cpanato (approved):
nice

mattmoor (approved):

PR #155, opened 2022-03-29
apko development environment

puerco:
This PR adds a shell script that allows a one-command way to start an apko development environment. 

Running the following script from the top of the apko repo launches a series of steps that use apko itself to build an image and drop the user into a development shell:

```
./hack/make-devenv.sh 
```

The environment is pre loaded with some tools (currently go, git and jq) and the user shell is opened with the apko repository mounted, ready to run any changes you make in your editor. I've also added a first draft of a CONTRIBUTING.md doc with instructions to launch the shell and basic info for contributors.

The image is `apko-inception` because apko is building the apko environment with a script that calls itself multiple times :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ bin/
 .vscode
 .idea
 tmp/
+_output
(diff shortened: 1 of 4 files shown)


puerco:
Let me know of any problems in other (non linux) platforms, etc

PR #157, opened 2022-03-29
Add docker cli to devenv.

amouat:
This PR adds the Docker CLI to the dev environment and  exposes the Docker socket in development container.
 
The outcome of this is that you can now run `docker load` inside the development environment and have it available on the host.

There are drawbacks:
  - it adds to the image size
  - there are security implications (the container effectively has full host access)

I'm assuming those are both ok for the development env, but we might want to add warning.


puerco (approved):
Thanks @amouat this is turning into our swiss army knife to hack on apko :tada:

puerco (approved):
Thanks!!

PR #158, opened 2022-03-30
Script to run apko commands via docker.

amouat:
This isn't quite ready yet, but I wanted to get some feedback.

Basically, if the dev-env has been built, you can run the script like this:

```
hack/run-devenv.sh build examples/nginx.yaml apko/test test.tar
```

This should work on platform with docker. It will also run `make` if the binary isn't there.

I'm not really sure what the best way to do this. We could merge with the other script, but note that it is really for hacking on apko, whilst this is more for being able to run the commands without a VM.

The script needs some improvement e.g. pulling out the image name.

### DIFF
--- a/hack/run-devenv.sh
+++ b/hack/run-devenv.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+# Copyright 2022 Chainguard, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+function run() {
+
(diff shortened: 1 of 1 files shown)


panekj:
Couldn't it be done in Makefile?

amouat:
@panekj you could run it from the makefile, but you'd still need a separate script to run inside the container.

PR #162, opened 2022-04-03
Fix config `archs` support

DamianZaremba:
There appears to be 2 issues with the current `archs` support from the config file.

1.

Currently if you try and specify the `archs` option in the config file,
the publish command fails along the lines of
```
  line 7: cannot unmarshal !!int `386` into types.Architecture
  line 7: cannot unmarshal !!str `amd64` into types.Architecture
```

Add a custom `UnmarshalYAML` function to `Architecture`, supporting
decoding the string values into the correct `Architecture` object.

2.

Currently if we specify 1 arch under `bc.ImageConfiguration.Archs`, the
publish command will handle it, however if we have more than 1 the
default case uses `archs`, which is defaulted to all entries (cli
options).

Swap out the default range case to use `bc.ImageConfiguration.Archs`,
which will be set to `archs` in the default case (no config specified).

### DIFF
--- a/pkg/build/types/types.go
+++ b/pkg/build/types/types.go
@@ -72,7 +72,7 @@ func (a Architecture) String() string { return a.s }
 func (a *Architecture) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	var buf string
 	if err := unmarshal(&buf); err != nil {
-		return nil
+		return err
 	}
 
 	a.s = ParseArchitecture(buf).s
--- a/pkg/build/types/types.go
+++ b/pkg/build/types/types.go
@@ -69,6 +69,16 @@ type Architecture struct{ s string }
 
 func (a Architecture) String() string { return a.s }
 
+func (a *Architecture) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var buf string
+	if err := unmarshal(&buf); err != nil {
+		return nil
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/build/types/types.go
```diff
@@ -69,6 +69,16 @@ type Architecture struct{ s string }
 
 func (a Architecture) String() string { return a.s }
 
+func (a *Architecture) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var buf string
+	if err := unmarshal(&buf); err != nil {
+		return nil
```
dlorenc, line 75: should this return err?

DamianZaremba: 🤦🏼‍♀️

DamianZaremba: Fixed in 59398a, ci failure appears due to an issue in https://github.com/chainguard-dev/actions/pull/35

kaniini:
I think that `alpine-386_amd64.yaml` might be causing problems?

DamianZaremba:
> I think that `alpine-386_amd64.yaml` might be causing problems?

I think everything is good now after https://github.com/chainguard-dev/actions/pull/36, unless I'm missing something.

kaniini:
Yeah seems so.

kaniini (approved):

PR #164, opened 2022-04-06
First pass at apko docs.

amouat:
Still needs work, in particular note things marked TK. Will continue tomorrow, just wanted to put this up in case people wanted to take a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Please note that apko is a work in progress and details are subject to change!
 
 An apko file for building an Alpine base image looks like this:
 
-```
+```yaml
 contents:
   repositories:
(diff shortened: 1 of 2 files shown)


amouat:
Moving some questions from the doc to here:

amouat:
I think this is ready for a look.

This one really is just a starting point, I expect it to evolve heavily. Please check the bit on keyrings - is there an example I could add?

At the moment it's not clear what apko elements are optional and when, but I think we can come back to that when after we get some feedback.

kaniini:
keyrings, idk, the idea there is to allow specification of keyrings with https:// sources, but it's hard to come up with a good example outside of what we're doing with apko and melange

PR #168, opened 2022-04-08
/etc/os-release generator

### LINKED ISSUE #165: implement /etc/os-release mutation
For scenarios where we are not pulling `alpine-base`, we need to generate our own `/etc/os-release` file.  We should probably also generate `/etc/<distro>-release` as well based on the `/etc/os-release` contents.

kaniini:
Generate `/etc/os-release` if it does not already exist.

Closes #165.

### DIFF
--- a/examples/alpine-slim.yaml
+++ b/examples/alpine-slim.yaml
@@ -1,0 +1,23 @@
+contents:
+  repositories:
+    - https://dl-cdn.alpinelinux.org/alpine/edge/main
+  packages:
+    - alpine-baselayout-data
+    - apk-tools
+    - busybox
+
+entrypoint:
+  command: /bin/sh -l
+
+# optional environment configuration
+environment:
+  PATH: /usr/sbin:/sbin:/usr/bin:/bin
+
+# data for /etc/os-release if it does not already exist
+# in the image
+os-release:
+  id: alpine
+  version-id: '3.16.0'
+  name: 'Alpine Slim'
+  pretty-name: 'Alpine Slim (apko)'
+
--- a/pkg/build/os-release.go
+++ b/pkg/build/os-release.go
@@ -1,0 +1,86 @@
+// Copyright 2022 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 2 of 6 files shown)


pkg/build/os-release.go
```diff
@@ -0,0 +1,86 @@
...
+			return err
+		}
+	}
+
+	if ic.OSRelease.BugReportURL != "" {
+		_, err := fmt.Fprintf(w, "BUG_REPORT_URL=\"%s\"\n", ic.OSRelease.BugReportURL)
+		if err != nil {
+			return err
+		}
+	}
```
puerco, line 83: What do you think if we do this in a template later to skip all the ifs and error checks?

puerco (approved):

pkg/build/os-release.go
kaniini: sure

examples/alpine-slim.yaml
```diff
@@ -0,0 +1,23 @@
...
+environment:
+  PATH: /usr/sbin:/sbin:/usr/bin:/bin
+
+# data for /etc/os-release if it does not already exist
+# in the image
+os-release:
+  id: alpine
+  version-id: '3.16.0'
+  name: 'Alpine Slim'
+  pretty-name: 'Alpine Slim (apko)'
```
mattmoor, line 22: Can we default this, or do we need to update all the distroless repos with this block?  I’d love for things to work with scanners by default 🤞

kaniini: it has defaults, but we need to figure out what sensible ones are

pkg/build/types/image_configuration.go
```diff
@@ -64,6 +64,23 @@ func (ic *ImageConfiguration) Validate() error {
...
+		ic.OSRelease.ID = "alpine"
+	}
+
+	if ic.OSRelease.Name == "" {
+		ic.OSRelease.Name = "apko-generated image"
+		ic.OSRelease.PrettyName = "apko-generated image"
+	}
+
+	if ic.OSRelease.VersionID == "" {
+		ic.OSRelease.VersionID = "3.999"
```
mattmoor, line 77: I assume this is what you mean by sensible?

mattmoor: Or things we need to make sensible…? 😂

mattmoor:
I wonder if we should add test files to each of the example directories and run something like:

   go test ./dir —digest=…

We can make them pretty trivial, but then we have a pattern for testing interesting properties of each example image.  In this case, the shape of /etc/os-release

pkg/build/types/image_configuration.go
kaniini: Things we need to make sensible, I guess.  "edge" is probably a good versionid to use here.

kaniini:
> I wonder if we should add test files to each of the example directories and run something like:
> 
> go test ./dir —digest=…
> 
> We can make them pretty trivial, but then we have a pattern for testing interesting properties of each example image. In this case, the shape of /etc/os-release

Lets focus on this for 0.4!

Issue #188, opened 2022-04-22
apko show-config applet


kaniini:
This tracks implementation of the `apko show-config` applet, discussed in #186.

The `show-config` applet will just show the finalized configuration that is generated from loading a YAML config and processing it.  It is intended to be a debug tool.

PR #202, opened 2022-05-17
cli: add show-config command

kaniini:
Closes: #188

### DIFF
--- a/pkg/cli/show-config.go
+++ b/pkg/cli/show-config.go
@@ -1,0 +1,86 @@
+// Copyright 2022 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	"chainguard.dev/apko/pkg/build"
+)
+
+func ShowConfig() *cobra.Command {
+	var extraKeys []string
+	var extraRepos []string
+
+	cmd := &cobra.Command{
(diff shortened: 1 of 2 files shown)


pkg/cli/show-config.go
```diff
@@ -0,0 +1,86 @@
...
+	var extraRepos []string
+
+	cmd := &cobra.Command{
+		Use:   "show-config",
+		Short: "Show the configuration derived from loading a YAML file",
+		Long: `Show the configuration derived from loading a YAML file.
+
+The derived configuration is rendered in YAML.
+`,
+		Example: `  apko show-config <config.yaml>`,
```
imjasonh, line 40:
```suggestion
		Example: "apko show-config <config.yaml>",
```

Unless there's a reason for it to have spaces prefixed?

pkg/cli/show-config.go
```diff
@@ -0,0 +1,86 @@
...
+		},
+	}
+
+	cmd.Flags().StringSliceVarP(&extraKeys, "keyring-append", "k", []string{}, "path to extra keys to include in the keyring")
+	cmd.Flags().StringSliceVarP(&extraRepos, "repository-append", "r", []string{}, "path to extra repositories to include")
+
+	return cmd
+}
+
+func ShowConfigCmd(ctx context.Context, opts ...build.Option) error {
```
imjasonh, line 58: Can this be unexported? Or is this something we want to export so we can use it outside of apko?

pkg/cli/show-config.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	var buf bytes.Buffer
+	enc := yaml.NewEncoder(&buf)
+
+	if err := enc.Encode(bc.ImageConfiguration); err != nil {
+		return fmt.Errorf("failed to encode YAML document: %w", err)
+	}
+
+	if _, err := buf.WriteTo(os.Stdout); err != nil {
```
imjasonh, line 81: Any reason not to write it straight to stdout? `yaml.NewEncoder(os.Stdout).Encode(bc.ImageConfiguration)`

(Obviously not a huge deal to buffer it, just curious)

imjasonh (approved):

pkg/cli/show-config.go
kaniini: Yes, it can be.

pkg/cli/show-config.go
kaniini: I wanted to catch any errors prior to dumping a partial doc, basically.

pkg/cli/show-config.go
kaniini: That's how cosign did it, I copied cosign :P

PR #232, opened 2022-06-15
Fix minor issues with README

jspeed-meyers:
Adding a period makes the bullets under features consistent. I also found, I think, a typo.


amouat (approved):

amouat:
Somehow this broke everything 😂 
Clearly we have flaky tests :(

PR #241, opened 2022-06-22
implement backoff for GGCR remote operations

kaniini:
This PR implements a backoff timer for GGCR remote operations which should hopefully reduce the number of publish failures we have when github container registry flakes out.

We might want the ability to override the default backoff policy, but it seems reasonable enough for now.


imjasonh (approved):
Nice.

PR #248, opened 2022-06-28
use logrus for logging

kaniini:
Use logrus for logging instead of the Go logging module which is not that great to use.


puerco (approved):
LGTM overall, thanks @kaniini . Now that we have the logger migrated to logrus, let's now focus on fine tuning the levels in messages. I think we can optimize them a lot more to be more useful and add context.

PR #249, opened 2022-06-28
build: types: allow "all" to be an alias for all architectures when parsing arch lists

kaniini:


pkg/build/types/types.go
```diff
@@ -212,6 +212,9 @@ func ParseArchitecture(s string) Architecture {
 func ParseArchitectures(in []string) []Architecture {
 	uniq := map[Architecture]struct{}{}
 	for _, s := range in {
+		if s == "all" {
```
imjasonh, line 215: Is the intention to be able to support --arch=x86,all? That seems a little weird.

I think in ko we only allow "all" if it's the only element.

cpanato: Yes, and that emits an warning if im not wrong

Issue #225, opened 2022-06-08
SBOMs point to layer, not image


jonjohnsonjr:
Looking at [this example](https://explore.ggcr.dev/?image=ghcr.io/distroless/static:sha256-f9ad8695390dc6dce29e47a0fd9cb4d5d48a052b45092f9d0e56dabaf0b384f1.sbom), the [blob](https://explore.ggcr.dev/json/ghcr.io/distroless/static@sha256:48d4c4f878554d6ee4da735bf72abd5f426f7210e678dc6275736efeb5a8d010?mt=spdx+json&size=3673) references what appears to be an image:

```
$ crane blob ghcr.io/distroless/static@sha256:48d4c4f878554d6ee4da735bf72abd5f426f7210e678dc6275736efeb5a8d010 | jq -r .packages[0].externalRefs[0].referenceLocator
pkg:oci/ghcr.io%2Fdistroless%2Fstatic:latest-20220608@sha256:eec03b05b2510ba950c79e67ae0723237dd1c37b28d6a12e36dab1e0846a38ba?arch=386&tag=latest-20220608
```

However, `ghcr.io%2Fdistroless%2Fstatic:latest-20220608@sha256:eec03b05b2510ba950c79e67ae0723237dd1c37b28d6a12e36dab1e0846a38ba` does not exist (are those slashes intentionally escaped?).

Based on [this code](https://github.com/chainguard-dev/apko/blob/c8935cf40a1d7d0ef5c931714a53f36f417aef9b/pkg/build/build_implementation.go#L119-L141), it looks like this is a reference to a layer, not the actual image, which we can confirm:

```
$ crane blob ghcr.io/distroless/static:latest-20220608@sha256:eec03b05b2510ba950c79e67ae0723237dd1c37b28d6a12e36dab1e0846a38ba | tar -tvzf - | head
drwx------ bin/bin           0 2022-06-07 18:55 bin
drwxr-xr-x root/root         0 2022-06-07 18:55 dev
crw------- root/root       5,1 2022-06-07 18:55 dev/console
crw-rw-rw- root/root       1,3 2022-06-07 18:55 dev/null
crw-rw-rw- root/root       1,8 2022-06-07 18:55 dev/random
crw-rw-rw- root/root       1,9 2022-06-07 18:55 dev/urandom
crw-rw-rw- root/root       1,5 2022-06-07 18:55 dev/zero
drwxr-xr-x root/root         0 2022-06-07 18:55 etc
[...]

PR #251, opened 2022-06-29
Fix image structure in apko SBOMs

puerco:
This PR implements proper image structure in the individual apko images. Both CycloneDX and SPDX have been updated to capture the structure of the images. The biggest change is when running `apko publish`:

* When running apko publish and generating a Cyclonedx SBOM, the layer is now a subcomponent of a new image component.
* When requesting an SPDX SBOM the image now has its own top level package and is related to the layer package.

It is a big refactor as the SBOM generation is now done after publishing the images and the SBOM is now attached at the end of the publish function.

Easiest to visualize it by looking at the sboms. This is before the change:

image

This is after the change:

image

There are other minor changes, each commit captures the changes as atomic as I could.

Fixes https://github.com/chainguard-dev/apko/issues/225

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -87,7 +87,7 @@ func (o *Options) TempDir() string {
 func (o Options) TarballFileName() string {
 	tarName := "apko.tar.gz"
 	if o.Arch.String() != "" {
-		tarName = fmt.Sprintf("apko-%s.tar.gz", o.Arch.String())
+		tarName = fmt.Sprintf("apko-%s.tar.gz", o.Arch.ToAPK())
 	}
 	return tarName
 }
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -15,6 +15,7 @@
 package options
 
 import (
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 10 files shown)


pkg/options/options.go
```diff
@@ -64,3 +67,27 @@ func (o *Options) Summarize(logger *logrus.Entry) {
...
+	}
+	o.TempDirPath = path
+	return o.TempDirPath
+}
+
+// TarballFileName returns a deterministic filename for the layer taball
+func (o Options) TarballFileName() string {
+	tarName := "apko.tar.gz"
+	if o.Arch.String() != "" {
+		tarName = fmt.Sprintf("apko-%s.tar.gz", o.Arch.String())
```
kaniini, line 90:
```suggestion
		tarName = fmt.Sprintf("apko-%s.tar.gz", o.Arch.ToAPK())
```

Otherwise we get filenames like `apko-arm/v6.tar.gz`.

kaniini (changes requested):

pkg/options/options.go
kaniini: Looks like there are still some fallout here.  Digging into it.

kaniini (dismissed):

kaniini (approved):

puerco:
GH UI seems to be acting up. Here's the mink e2e successful run: https://github.com/chainguard-dev/apko/runs/7120080211?check_suite_focus=true

PR #254, opened 2022-06-29
Add installation instructions before quickstart

jspeed-meyers:
Breaking out installation would, IMO, have accelerated my timeline to running and using `apko`.

I suspect there are additional ways to install `apko` and that might be worth calling out.

Additionally, I suspect there are additional steps here that likely need to be spelled out explicitly. Do mac users need other software to be installed and/or running for this hack script to perform as expected?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,26 @@ apko has the following key features:
 
 Please note that apko is a work in progress and details are subject to change!
 
+## Installation
+
+The quickest way to get an environment for running apko on Mac or Linux is to clone the repo and use the scripts under the hack
+directory:
+
+```
+$ ./hack/make-devenv.sh
+
+Welcome to the apko development environment!
+
+[apko] ❯
+```
+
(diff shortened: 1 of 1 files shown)


amouat (approved):

amouat:
Yeah, I was meaning to do this tbh. It's obvious in retrospect. Thanks for going the extra mile and creating the PR!

erikaheidi:
this is great @jspeed-meyers , apologies for the late reply :) I am consolidating my notes after more tests to figure what is the easiest way to run it on Ubuntu, I think it's gonna be just one additional command to set some git config on the VM before make install. I can make a PR with an update once this is more clear.

erikaheidi:
AFAIK yes, Docker is needed if you are using the scripts at `hack/` to run the dev env, but if you are on Alpine you can probably run it directly. For wider compatibility IMO we should focus on the docker env so yeah this should definitely be added to requirements.

PR #258, opened 2022-06-30
Make apk dependency clear

amouat:


kaniini:
Good for now, but I am about to add code to download and fetch a local copy of apk-tools-static, which will drop this requirement :D

README.md
```diff
@@ -16,8 +16,7 @@ Please note that apko is a work in progress and details are subject to change!
 
 ## Installation
 
-The quickest way to get an environment for running apko on Mac or Linux is to clone the repo and use the scripts under the hack
-directory:
+apko has a dependency on [apk-tools](https://gitlab.alpinelinux.org/alpine/apk-tools). If you're not running on Alpine Linux, the quickest way to get  apko running is to clone the repo and use the scripts under the hack directory:
```
kaniini, line 19:
```suggestion
apko has a dependency on [apk-tools](https://gitlab.alpinelinux.org/alpine/apk-tools). If you're not running on Alpine Linux, the quickest way to get  apko running is to clone the repo and use the scripts under the hack directory:
```
```suggestion
apko has a dependency on [apk-tools](https://gitlab.alpinelinux.org/alpine/apk-tools). If you're not running on Alpine Linux or another apk-based distribution, the quickest way to get apko running is to clone the repo and use the scripts under the hack directory:
```

kaniini (approved):

PR #273, opened 2022-07-07
Detect VCS URLs and set org.opencontainers.image.source if found

### LINKED ISSUE #231: Set org.opencontainers.image.source if we know it
We should set the `org.opencontainers.image.source` annotation if we know it.  Should be possible to do it by deducing a `vcs-url` from the `origin` upstream if `.git` is present.

kaniini:
Probe for Git repositories relative to the apko.yaml file and set a VCS URL annotation if we find one.

Closes #231!


kaniini:
@puerco is there any way we can embed the VCS URL in the SBOMs?

puerco:
We can, in SPDX we could declare apko at buildpoint X as `BUILD_TOOL_OF` the image.

puerco (approved):
Looks good to me
(I assume we're getting a test soon, right? :innocent: )

kaniini:
Yeah, I plan on adding a test for this, but since it is more of an integration test, I want to think about how it works for a second.

PR #275, opened 2022-07-08
add working directory setting

### LINKED ISSUE #271: Add support for setting container workdir
An apko equivalent of the Dockerfile WORKDIR instruction would be useful.

If this already exists, please add it to the docs.

kaniini:
closes #271


amouat (approved):
That was simple!

Nice :)

amouat:
I guess it needs added to the docs, I can do that in separate PR.

PR #277, opened 2022-07-08
Add recursive option to PathMutation to traverse directory and modify perms

priyawadhwa:
For distroless.dev/nginx, the `/var/lib/nginx` and `/run/nginx` directories need to be writable by the nginx process.
Right now, there are at least 6 subdirectories/files that also need updated permissions.

We could write each one out explicitly in apko.yaml, but I think it would be easier to give people the option
to change permissions on a directory and all subdirectories/files.
This way we just have to explicitly call out the two dirs mentions above and set `recursive: true` e.g.

```yaml
  - path: /var/lib/nginx
    uid: 10000
    gid: 10000
    type: directory
    permissions: 0o755
    recursive: true
  - path: /run/nginx
    uid: 10000
    gid: 10000
    type: directory
    permissions: 0o755
    recursive: true
```

Signed-off-by: Priya Wadhwa <priya@chainguard.dev>

### DIFF
--- a/pkg/build/paths.go
+++ b/pkg/build/paths.go
@@ -42,6 +42,10 @@ func mutatePermissions(o *options.Options, mut types.PathMutation) error {
 		return err
 	}
 
+	if err := os.Chown(target, int(mut.UID), int(mut.GID)); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


amouat:
I wonder if "recursive" would be a better name (in keeping with chmod -r)

priyawadhwa:
SGTM!

amouat (approved):
LGTM. I see why you called it walkdir now, but I think recursive is probably better from a user perspective.

PR #278, opened 2022-07-08
Add Docker installation to README installation

### LINKED ISSUE #259: Add docker installed as requirement in README (for MacOS only?)
> AFAIK yes, Docker is needed if you are using the scripts at hack/ to run the dev env, but if you are on Alpine you can probably run it directly. For wider compatibility IMO we should focus on the docker env so yeah this should definitely be added to requirements.

@erikaheidi mentioned this in now closed PR #254. Should docker installation be included in the README in the installation section a prerequisite. Is it only a prerequisite for Mac users?

Once I have an answer, I'm glad to put in a PR.

jspeed-meyers:
Closes #259

Is this helpful? If not, should there be something else? Or should I simply close this issue and PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Please note that apko is a work in progress and details are subject to change!
 
 ## Installation
 
-apko has a dependency on [apk-tools](https://gitlab.alpinelinux.org/alpine/apk-tools). If you're not running on Alpine Linux or another apk-based distribution, the quickest way to get apko running is to clone the repo and use the scripts under the hack directory:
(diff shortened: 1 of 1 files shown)


kaniini:
In general, the hack scripts are replaced by `ghcr.io/distroless/apko:latest`, which does require Docker to run :)

I would recommend mentioning that instead of what is there presently, as its obsolete information now.

jspeed-meyers:
That sounds good. I'm afraid I'll dork up the best way to explain this, so I've unassigned myself and will late someone more familiar with the project or documentation to make this change.

amouat:
I added instructions to use the new Docker image in #311, which I think replaces this.

jdolitsky:
closing as old/inactive. please re-open if necessary

Issue #281, opened 2022-07-10
OCI purl's don't follow the spec


mattmoor:
I see the following in the `apko` `purl`s:
```
pkg:oci/ghcr.io%2Fdistroless%2Fstatic:latest-20220710@sha256:96556b7e982dbca4c582b5bee35ff205d2e1404a616e8f1f822c0def27b5ac3f?arch=amd64&tag=latest-20220710
```

According to [the spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci) this should be:
```
pkg:oci/static@sha256:96556b7e982dbca4c582b5bee35ff205d2e1404a616e8f1f822c0def27b5ac3f?repository_url=ghcr.io%2Fdistroless%2Fstatic&arch=amd64&tag=latest-20220710
```

cc @puerco

kaniini:
Yeah I was wondering about that, if puerco doesn’t get to it tomorrow I can fix it.

kaniini:
Fixed by #283.

PR #283, opened 2022-07-11
Move purl value generation to SBOM options

puerco:
This PR modifies the way we generate the purl data for the apko images, shifting the logic to the options object.

This has the benefit of ensuring we get the same purls in both standards by reusing the same logic.

It also fixes a bug where the purl names were formed incorrectly in the image SBOMs, fixing https://github.com/chainguard-dev/apko/issues/281

New sample image purl:

```json
"purl": "pkg:oci/apko-test@sha256:ea959168730bccc00500a22b58a1891c1b07dc9039273b07d5a833f7fa7d7b54?arch=amd64\u0026os=linux",

```

/cc @mattmoor @kaniini 


kaniini:
what's with the `\u0026` there?  shouldn't it be an `&`?

puerco:
I thnk its because the json marshaller is encoding the string.

kaniini:
ugh, is there a way to convince it to not do that :D

puerco:
Nice and clean now:

```json
      "components": [
        {
          "bom-ref": "pkg:oci/apko-test@sha256:ea959168730bccc00500a22b58a1891c1b07dc9039273b07d5a833f7fa7d7b54?arch=amd64&os=linux",
          "type": "container",
```

PR #286, opened 2022-07-12
Simplify purl index/image names, abstract to function

puerco:
This commit adds a parser to the name returned by `RepositoryStr()` to return as the purl image name only the last part when several slashes are in the name. We now have a specific function in options to craft the purl name of the index.

The same commit also reuses the index function in the SPDX and CycloneDX implementations.

The previous purl names will now change to be as expected:

From: `pkg:oci/distroless%2Fstatic@sha256:ed7854ed`

To: `pkg:oci/static@sha256:ed7854ed`

### DIFF
--- a/pkg/sbom/options/options.go
+++ b/pkg/sbom/options/options.go
@@ -15,6 +15,7 @@
 package options
 
 import (
+	"strings"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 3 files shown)


pkg/sbom/options/options.go
```diff
@@ -85,7 +86,17 @@ func (o *Options) ImagePurlName() string {
 		if err != nil {
 			return repoName
 		}
-		repoName = ref.Context().RepositoryStr()
+		parts := strings.Split(ref.Context().RepositoryStr(), "/")
+		repoName = parts[len(parts)-1]
```
mattmoor, line 90: You may be able to use `filepath.Base`

mattmoor (approved):

pkg/sbom/options/options.go
puerco: Done!

mattmoor (approved):

PR #289, opened 2022-07-12
accounts: resolve non-UID run-as users to their UIDs

### LINKED ISSUE #129: Favor `uid` over username in User directive.
I notice that I can use either of these syntaxes:
```yaml
  run-as: nonroot
  run-as: 65532
```

... and we in fact favor the former in our examples.

I believe that certain runtime security features in Kubernetes don't work properly with the string variant, but do work with the uid variant.  I believe the reason is that if user is a uid, all you need is the config to evaluate policy, where if it is username then you have to fetch the image contents to access `/etc/passwd` to determine the uid, which in general means downloading arbitrary amounts of data.

If we can, then I'd suggest we bias towards doing the "right" thing regardless of input, and using the uid variant.  Regardless, we should probably switch the images where we use this to get the uid result.

kaniini:
```
accounts:
  run-as: nonroot
```

now automatically becomes

```
accounts:
  run-as: 65532
```

closes #129

### DIFF
--- a/pkg/build/accounts.go
+++ b/pkg/build/accounts.go
@@ -143,6 +143,16 @@ func (di *defaultBuildImplementation) MutateAccounts(
 			return err
 		}
 
+		// Resolve run-as user if requested.
+		if ic.Accounts.RunAs != "" {
+			for _, ue := range uf.Entries {
+				if ue.UserName == ic.Accounts.RunAs {
+					ic.Accounts.RunAs = fmt.Sprintf("%d", ue.UID)
+					break
+				}
+			}
+		}
+
 		return nil
 	})
 


pkg/build/accounts.go
```diff
@@ -143,6 +143,15 @@ func (di *defaultBuildImplementation) MutateAccounts(
 			return err
 		}
 
+		// Resolve run-as user if requested.
+		if ic.Accounts.RunAs != "" {
+			for _, ue := range uf.Entries {
+				if ue.UserName == ic.Accounts.RunAs {
+					ic.Accounts.RunAs = fmt.Sprintf("%d", ue.UID)
```
jdolitsky, line 150: `break` ?

Issue #288, opened 2022-07-12
List all images in multiarch images when using --image-refs


puerco:
In order to simplify batch operations (like signing, for example) on images built with `apko publish` we should list all arch images built when specifying `--image-refs`. 

Currently, when specifying the flag, the file which gets written only has the reference to the multi-arch index.

This behavior change would make `apko` follow `ko` as proposed in this change by @mattmoor  https://github.com/google/ko/pull/761

PR #295, opened 2022-07-16
Tag and omit references to the refs file

puerco:
We now tag and emit references into the refs file for all arch images to help external
tools know what we've built (for example to sign  them).

Closes: https://github.com/chainguard-dev/apko/issues/288

/cc @mattmoor 

### DIFF
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	coci "github.com/sigstore/cosign/pkg/oci"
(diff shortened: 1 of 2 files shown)


internal/cli/publish.go
```diff
@@ -243,23 +249,18 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
 
 // publishImage publishes a specific architecture image
 func publishImage(bc *build.Context, layerTarGZ string, arch types.Architecture) (imgDigest name.Digest, img coci.SignedImage, err error) {
-	// We only tag the images if we're building a single one
-	imageTags := []string{}
-	if len(bc.ImageConfiguration.Archs) == 1 {
```
mattmoor, line 248: I'm curious why we need this part of the change, or whether it regresses something 🤔

mattmoor (approved):

mattmoor:
Once this lands, this downstream change will get pretty interesting 😉 

https://github.com/distroless/static/pull/11

internal/cli/publish.go
puerco: Do you think it's better if I dont pass it to the tagging function and build the reference by hand? I did passed it because it was returning an empty string .

mattmoor: I don't have enough context here, so I'd defer to @kaniini

puerco: OK, I thought you had a specific problem in mind. I think we're fine, tests are ok and I've been using this branch today ok.

PR #305, opened 2022-07-26
Refactor use of options

priyawadhwa:
This refactor is in preparation for adding a new flag, which will allow us to tag the final image with the version of a package, so we can build `distroless.dev/go` and automatically tag with `1.18` or whatever

Signed-off-by: Priya Wadhwa <priya@chainguard.dev>


jdolitsky (approved):
good stuff

PR #306, opened 2022-07-26
Add --package-version-tag flag to `apko publish`

priyawadhwa:
When this flag is set to a particular package, e.g. `--package-version-tag=go`, apko will immediately take the version of that package (e.g. `1.18`) and tag all images it was going to publish with that version tag.

It does this by grabbing the version for the package from the `/lib/apk/db/installed` file. If there's a better way to do this I'm happy to update, this was just the only way I could find. 

This should make it easy for us to start tagging all images with version numbers. 

This will be smaller once https://github.com/chainguard-dev/apko/pull/305 merges first

### DIFF
--- a/pkg/apk/apk.go
+++ b/pkg/apk/apk.go
@@ -17,9 +17,15 @@ package apk
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
 	"runtime"
+	"strings"
 
+	"github.com/google/go-containerregistry/pkg/name"
 	"golang.org/x/sync/errgroup"
 
 	"chainguard.dev/apko/pkg/build/types"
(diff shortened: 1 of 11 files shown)


kaniini:
Nice!

pkg/apk/apk.go
```diff
@@ -126,6 +132,64 @@ func (a *APK) Initialize(ic *types.ImageConfiguration) error {
...
 }
 
+// AdditionalTags is a helper function used in conjunction with the --package-version-tag flag
+// If --package-version-tag is set to a package name (e.g. go), then this function
+// returns a list of all images that should be published with the associated version of that package tagged (e.g. 1.18)
+func AdditionalTags(opts options.Options) ([]string, error) {
+	if opts.PackageVersionTag == "" {
+		return nil, nil
+	}
+	dbPath := filepath.Join(opts.WorkDir, "lib/apk/db/installed")
```
kaniini, line 142: There is already code to parse the IDB into package objects.  See https://github.com/chainguard-dev/apko/blob/main/pkg/sbom/sbom.go#L88-L97.

This is probably more robust than parsing the `V:` field out by hand, since the order of fields is not formally defined.

kaniini (changes requested):

pkg/apk/apk.go
priyawadhwa: Thanks, this is much better!

kaniini (approved):

Issue #308, opened 2022-07-28
Surface SBOMs from /var/lib/db/sbom


jdolitsky:
After building the image filesystem, scan the `/var/lib/db/sbom` directory looking for files in the form:

- `$package.cdx` (CycloneDX)
- `$package.spdx.json` (SPDX)

These files should be included during a melange build of APK with name `$package`.

The package data from these should be included in the final SBOM(s), noting the APK which it derives from.

amouat:
Upped this to P1 as I think it would give us a strong sell in demos if our SBOMs are bomb.

luhring:
Closed by https://github.com/chainguard-dev/apko/pull/309 (thanks @puerco!)

PR #309, opened 2022-07-28
WIP: Merge packages from SBOMs found in apks 

puerco:
When assembling the layer SBOM, apko will now check for SBOMs included in the APKs it is adding to the layer. If they are found, it will parse them, yank out the packages in the document and merge them into the proper place in the layer SBOM.

Still to do: Our SPDX document type does not have `File` support, therefore any files found in the apk SBOMs will not be transferred.


kaniini:
We should add `File` support, it will unlock a lot of possibilities.

jdolitsky:
tied to #308

jdolitsky (approved):

PR #330, opened 2022-08-10
VFS core (for rootless apko)

kaniini:
The VFS package provides an overlay filesystem, where files are physically backed by an underlay filesystem, but permission and ownership are tracked by the VFS itself.  This allows for `Chown` and `Chmod` operations to work without requiring root.

### DIFF
--- a/pkg/vfs/dirfs.go
+++ b/pkg/vfs/dirfs.go
@@ -1,0 +1,53 @@
+// Copyright 2022 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vfs
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+type dirFS string
+
+// DirFS is a DirFS implementation that is suitable for use as a
+// BaseFS.
+func DirFS(path string) (BaseFS, error) {
+	return dirFS(path), nil
+}
+
+func (dir dirFS) finalPath(path string) string {
+	return filepath.Join(string(dir), path)
+}
+
+func (dir dirFS) Create(path string) (fs.File, error) {
+	return os.Create(dir.finalPath(path))
+}
+
+func (dir dirFS) Open(path string) (fs.File, error) {
+	return os.Open(dir.finalPath(path))
+}
+
+func (dir dirFS) ReadDir(path string) ([]fs.DirEntry, error) {
+	return os.ReadDir(dir.finalPath(path))
+}
+
+func (dir dirFS) ReadFile(path string) ([]byte, error) {
+	return os.ReadFile(dir.finalPath(path))
+}
+
+func (dir dirFS) Stat(path string) (fs.FileInfo, error) {
+	return os.Stat(dir.finalPath(path))
+}
--- a/pkg/vfs/dirfs.go
+++ b/pkg/vfs/dirfs.go
@@ -52,3 +52,11 @@ func (dir dirFS) ReadFile(path string) ([]byte, error) {
 func (dir dirFS) Stat(path string) (fs.FileInfo, error) {
 	return os.Stat(dir.finalPath(path))
 }
+
+func (dir dirFS) Remove(path string) error {
+	return os.Remove(dir.finalPath(path))
+}
+
+func (dir dirFS) RemoveAll(path string) error {
+	return os.RemoveAll(dir.finalPath(path))
+}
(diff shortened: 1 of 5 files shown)


pkg/vfs/vfs.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	out := []fs.DirEntry{}
+	for _, dentry := range de {
+		if patchedINode, ok := baseINode.Children[dentry.Name()]; ok {
+			out = append(out, patchedINode)
+		} else {
+			out = append(out, dentry)
+		}
+	}
```
puerco, line 263: So this is the core of the VFS, right? I'll try to regurgitate this to see if I'm getting it:

The way I understand the code is that the VFS will return the calls chrooted to the underlying directory. But I can create "phantom" paths (handled with `iNode`) which don't really exist in the underlying directory. Those phantom paths take precedence over the real files in the source directory. 

Is that a fair interpretation?

puerco (commented):
Thanks @kaniini ! I left a couple of suggestions and a question just to confirm I'm getting all of this right.

Since we are introducing a new package would you mind adding a description of its purpose to the PR body? 

Great work! :)

pkg/vfs/vfs.go
```diff
@@ -0,0 +1,283 @@
...
+// An INode contains overlay metadata for a filesystem entry that would
+// otherwise be represented by a traditional inode on disk.
+type INode struct {
+	Filename      string
+	Children      map[string]*INode
+	UnderlayINode syscall.Stat_t
+}
+
+func parseElements(path string) (string, string, bool) {
+	pathElements := strings.Split(path, "/")
```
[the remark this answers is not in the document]
kaniini: filepath.SplitList splits on `:`, not on `/`.

pkg/vfs/vfs.go
kaniini: well, the file content is backed by the underlying directory, but the permissions / ownership are phantom :)

pkg/vfs/vfs_unit_test.go
```diff
@@ -0,0 +1,89 @@
...
+	err = vfs.Chmod("./etc/motd", 0755)
+	require.NoError(t, err)
+
+	fi, err = vfs.Stat("./etc/motd")
+	require.NoError(t, err)
+
+	st, ok = fi.Sys().(*syscall.Stat_t)
+	require.True(t, ok, "must present a Stat_t")
+
+	require.Equal(t, st.Mode&0755, uint32(0755), "must have mode 0755")
```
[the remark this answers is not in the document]
kaniini: These are the underlying Stat_t, which requires `uint32`.

pkg/vfs/dirfs_unit_test.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2022 Chainguard, Inc.
```
imjasonh, line 1: Idiomatically this should probably be called `dirfs_test.go`

pkg/vfs/vfs.go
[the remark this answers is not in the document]
imjasonh: Can we at least `strings.Split(path, filepath.Separator)` ?

pkg/vfs/dirfs.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2022 Chainguard, Inc.
```
imjasonh, line 1: Is DirFS something we expect to use in melange or elsewhere? In general I'd bias toward keeping it `internal` until the API settles and we decide we want external folks to use it.

pkg/vfs/dirfs_unit_test.go
```diff
@@ -0,0 +1,75 @@
...
+	assert.Equal(t, st.IsDir(), false, "etc/motd is a normal file")
+
+	inF, err := dir.Open("./etc/motd")
+	require.NoError(t, err)
+	defer inF.Close()
+
+	data, err := io.ReadAll(inF)
+	require.NoError(t, err)
+
+	assert.Equal(t, data, []byte{'H', 'e', 'l', 'l', 'o', ' ', 'w', 'o', 'r', 'l', 'd', '\n'}, "motd should return Hello world")
```
imjasonh, line 58:
```suggestion
	assert.Equal(t, data, []byte("Hello world"), "motd should return Hello world")
```

pkg/vfs/dirfs.go
kaniini: Yes, I expect to use it in Melange.

pkg/vfs/vfs.go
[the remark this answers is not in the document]
kaniini: it would need to be `strings.Split(path, string(filepath.Separator))`, since filepath.Separator is a byte :(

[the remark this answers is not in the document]
puerco: Yeah I was trying to be clever and assumed SplitList would split on the `filepath.Separator` char

pkg/vfs/dirfs_unit_test.go
puerco: Blame me for that pattern, I tend to put the unit tests in a `_unit_test.go` file like that because I usually use `package_test.go` for integration tests (running in a different package, outside of the test subject). Like this:

https://github.com/chainguard-dev/apko/tree/main/pkg/sbom

Sorry (and also happy to do it in another way) :grimacing:

imjasonh: Oh okay. I don't feel that strongly, especially if there's precedent elsewhere in this code.

imjasonh (approved):

Issue #352, opened 2022-09-16
Allow configuring OCI annotations


imjasonh:
https://github.com/opencontainers/image-spec/blob/main/annotations.md

Let users configured annotations in apko.yaml

We could also do something smart with `image.source`, `image.revision` and `image.version` if we can detect we're running from a clean Git commit, but those smarts should probably be opt-in.

imjasonh:
Oh most of this is already done in https://github.com/chainguard-dev/apko/pull/296, it's just not documented.

Should we do the smart thing and try to detect source URL/revision/version?

imjasonh:
There's already `--vcs` (default true) which sets the source URL: https://github.com/chainguard-dev/apko/blob/868f0dc23e721039f9669b56d01ea4b897f2fb24/pkg/build/oci/oci.go#L112

JTarasovic:
It also seems like `publish` doesn't currently actually add the annotations from the config file.

I set up a repro with a fork of chainguard-images/static at JTarasovic/static:
https://github.com/chainguard-images/static/compare/89993cb...JTarasovic:static:main

And published using the action to ghcr: https://github.com/JTarasovic/static/actions/runs/3147616207/jobs/5117290180

Without luck
```
regctl image manifest ghcr.io/jtarasovic/static:latest
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:59c6fed2d149b743da3c62915307ada35752233b545f902260dea45cf46c56f2",
    "size": 509
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:daad3cbec1232c756efc4df8dfb471fab62a7f9f38b7fb495ca403b2f163b43e",
      "size": 489220
    }
  ],
  "annotations": {
    "org.opencontainers.image.source": "gh:chainguard-images/static@065bb0cba7b326a3258eeab84585195cb46cd674"
  }
}
```

If there's anything I can do to help, don't hesitate to let me know.

imjasonh:
https://github.com/chainguard-dev/apko/pull/353 should fix the issue. There's a question in the PR about how to handle the vcsurl, if there's any pushback there we can at least merge the annotations support.

JTarasovic:
Nice! I thought it might but couldn't tell for sure with a brief review.

[...]

PR #353, opened 2022-09-16
add annotations to indexes, separate repo and revision annotations, document

imjasonh:
Fixes https://github.com/chainguard-dev/apko/issues/352

### DIFF
--- a/docs/apko_file.md
+++ b/docs/apko_file.md
@@ -85,6 +85,9 @@ archs:
  - amd64
  - 386
 
+annotations:
+  foo: bar
+  bar: baz
 ```
 
 Details of each field can be found below.
@@ -166,7 +169,7 @@ There are several child elements:
 
 ### Environment
 
-`environment` defines a list of environment variables to set within the image e.g: 
+`environment` defines a list of environment variables to set within the image e.g:
 
 ```yaml
 environment:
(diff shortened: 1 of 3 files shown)


kaniini:
It looks fine to me, but the splitting the VCS URL seems wrong.  Going to wait to see what @puerco says.

imjasonh:
ping @puerco

puerco:
Sorry, I missed this while I was out.

I'm a bit unsure here as we're in OCI territory but I too would prefer to have the VCS URLs be full [SPDX download locator-compatible URLs](https://spdx.github.io/spdx-spec/package-information/#771-description) as it is one piece of data with everything you need in there:

`<vcs_tool>+<transport>://<host_name>[/<path_to_repository>][@<revision_tag_or_branch>][#<sub_path>]`

But perhaps it is not a valid entry in the OCI annotation? I guess it depends on the intended use. Without knowing more context, it seems to me it could work as the spec lists the field as:

> URL to get source code for building the image (string)

(I would add the hash to `.revision` regardless of what we put in `.source`)

imjasonh:
The OCI annotations are specified as:

> - **org.opencontainers.image.source** URL to get source code for building the image (string)
> - **org.opencontainers.image.revision** Source control revision identifier for the packaged software.

This says to me it should be a separate URL and revision.

The vcsurl passed through to the SBOM is still the single SPDX locator, it's only when setting the annotations that they get `strings.Cut` apart.

kaniini (approved):

PR #355, opened 2022-09-20
Add --tag-suffix to apko publish command.

kaniini:
This is necessary to add a suffix to the generated version tags, e.g. `foo:1.2.3-glibc` instead of just `foo:1.2.3`.

### DIFF
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -42,6 +42,7 @@ func publish() *cobra.Command {
 	var buildDate string
 	var sbomPath string
 	var packageVersionTag string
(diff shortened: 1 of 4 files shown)


cpanato (approved):

amouat (approved):
LGTM. I wonder if we should do some verification on the input though e.g. disallow certain characters (spaces?) or starting with a number?

amouat:
Oh, I was a bit late :)

kaniini:
Improvements welcome for the next release cycle, of course ❤️

PR #367, opened 2022-09-29
update references to distroless

imjasonh:
- distroless.dev -> cgr.dev/chainguard
- github.com/distroless -> github.com/chainguard-images
- remove an unused reference to gcr.io/distroless

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ endif
 
 GOFILES ?= $(shell find . -type f -name '*.go' -not -path "./vendor/*")
 
-RUNTIME_IMAGE ?= gcr.io/distroless/static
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -11,7 +11,6 @@ endif
 
 GOFILES ?= $(shell find . -type f -name '*.go' -not -path "./vendor/*")
 
-RUNTIME_IMAGE ?= gcr.io/distroless/static
```
amouat, line 14: This was unreferenced?

amouat (approved):

Makefile
kaniini: Yeah we don't use it

PR #369, opened 2022-09-30
upgrade to go1.19

cpanato:
- upgrade to go1.19

### DIFF
--- a/pkg/sbom/generator/spdx/spdx.go
+++ b/pkg/sbom/generator/spdx/spdx.go
@@ -197,7 +197,7 @@ func (sx *SPDX) imagePackage(opts *options.Options) (p *Package) {
 }
 
 // apkPackage returns a SPDX package describing an apk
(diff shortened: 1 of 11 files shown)


pkg/sbom/generator/spdx/spdx.go
```diff
@@ -197,7 +197,7 @@ func (sx *SPDX) imagePackage(opts *options.Options) (p *Package) {
 }
 
 // apkPackage returns a SPDX package describing an apk
-func (sx *SPDX) apkPackage(opts *options.Options, pkg *repository.Package) (p Package, err error) {
```
imjasonh, line 200: Nope, this isn't to satisfy an interface. We can just remove the err return. I'll send a PR if you don't.

cpanato: https://github.com/chainguard-dev/apko/pull/371

PR #388, opened 2022-10-19
don't double-escape purl qualifiers

imjasonh:
We had been manually escaping qualifiers until purl fixed this upstream. It looks like the upstream fix landed, so now we're double-escaping, which breaks some parsing.

cc @eddiezane @mattmoor


mattmoor (approved):

mattmoor:
🙏

PR #390, opened 2022-10-19
Fix unecoded characters in purls

puerco:
This PR fixes a problem where apko would not encode properly some values in JSON values.


imjasonh (approved):

puerco:
/hold 

No, I just noticed an error

imjasonh (approved):

PR #391, opened 2022-10-20
FIx purl qualifier encoder bug, add unit test

puerco:
This PR fixes a bug in the format of purl qualifiers and introduces a unit test to check the key value string and the encoding


cpanato (approved):
looks fancy
/lgtm

PR #406, opened 2022-11-10
Document build process

deitch:
As described. Ready for some feedback.

### DIFF
--- a/docs/build-process.md
+++ b/docs/build-process.md
@@ -1,0 +1,75 @@
+# apko Build Process
+
+apko builds an OCI-compliant image, stored in a tar file, that can then be loaded into a container runtime.
+The entire build is configured from a declarative file `apko.yaml`, which does not allow the execution of arbitrary
+commands. Instead, all content is generated from installing `apk` packages.
+
+The build process is driven by the implementation of the `apko build` command, specifically
+[`BuildCmd()`](../internal/cli/build.go#L104).
+
+The entire build process involves laying out the desired files in a temporary working directory, and then
+packaging that working directory as an OCI filesystem layer `.tar.gz`. With that layer file in hand,
+it can be packaged into an OCI image, and an SBoM can be created.
+
+The process is as follows:
+
+1. Create a temporary working directory.
+1. Create a [`build.Context`](../pkg/build/build.go#L37-45). This `Context` contains:
+   * The path to the config file, default to `apko.yaml`
+   * The parsed config file into an internal structure [`ImageConfiguration`](../pkg/build/types/types.go#L55-83)
+   * The [`buildImplementation`](../pkg/build/build_implementation.go#L43-59), which is the engine responsible for executing the actual build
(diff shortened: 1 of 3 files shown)


docs/build-process.md
```diff
@@ -0,0 +1,75 @@
...
+
+1. Create a temporary working directory.
+1. Create a [`build.Context`](../pkg/build/build.go#L37-45). This `Context` contains:
+   * The path to the config file, default to `apko.yaml`
+   * The parsed config file into an internal structure [`ImageConfiguration`](../pkg/build/types/types.go#L55-83)
+   * The [`buildImplementation`](../pkg/build/build_implementation.go#L43-59), which is the engine responsible for executing the actual build
+   * The [`Executor`](../pkg/exec/exec.go#L26-31), which handles external command execution by the `buildImplementation`
+   * The [`s6.Context`](../pkg/s6/s6.go#L23-26), which contains configuration for optionally installing the s6 supervisor to manage the process in the container
+   * Build-time options
+1. Refresh the `build.Context`, which sets initialization and runtime parameters, such as isolation, working directory, the executor and the s3 context.
```
luhring, line 24:
```suggestion
1. Refresh the `build.Context`, which sets initialization and runtime parameters, such as isolation, working directory, the executor and the s6 context.
```

luhring, line 24: (non-blocking) This is probably a concern about the code flow more than the docs, but I'm having a hard time wrapping my head around this "refresh" concept. For later, it'd be nice to have doc comments for important methods in the Go code.

luhring (commented):
This is amazinggg 😍  I learned a ton just reading this, honestly. I left just a couple of comments, but it looks good to me. It might be good to have someone more experienced review this for accuracy, too.

deitch:
> This is amazinggg 😍 I learned a ton just reading this, honestly

Probably nowhere near as much as I learned writing this! I am sure I made mistakes, so it would be great to have someone who knows this cold review it. Hoping @kaniini will be able to review (who knows far more than I do).

docs/build-process.md
deitch: It is a really small supervisor, only _half_ the size of s6. 😆 

Fixed.

deitch: The idea of `Refresh()` threw me for a bit of a loop the first time I read it too. As far as I can tell from skimming the code, it pretty much does what it says: Checks exec parameters (like, did we request qemu but cannot use it with proof in this circumstance, etc.), create a new executor and a new s6 setup, and then send them back.

I am not sure why this is named `Refresh()`, it seems more of an initializer. I think @kaniini would know the history a bit better.

deitch:
I added some of the inline docs.

kaniini (approved):

kaniini:
I looked through it and it all looked right to me!

docs/build-process.md
deitch: So if it were named precisely what it is, it would be called `InitOrIfAlreadyInitializedThenRefresh()`, but that is a lot more convoluted than `Refresh()`. :-)

PR #412, opened 2022-11-20
Pure go apk.InitDB()

puerco:
This PR implements the apk.InitDB() method in pure go to remove another call to 
the apk binary. We still need to run as root as this function needs to create device
files in the apk filesystem inside the working directory.

The functionality is implemented in three utility functions:

 `apk.initCreateFileSystem()`
 `apk.initCreateDeviceFiles()`
 `apk.initWriteScriptsTarball()`

Unit tests are included for these. 

@kaniini I think I've captured everything happening during the initialization, can you confirm?

PR #426, opened 2022-12-04
execute-free apk installation

deitch:
Ability to run apk on non-Linux platforms, without chroot/proot, without qemu, without root, without installed utilities.

This is pretty hefty - it might be better broken down into multiple PRs (or at least commits), I am open to suggestions - so I will try and break it down:

* apk implementation: all in `pkg/apk/impl/`. This intentionally is built as something that we could remove, either upstream to alpine-go or to a separate library of our own. It is consumed by `pkg/apk/`. This replaces all `apk` exec calls.
* rwfs: an interface defined an extension of `fs.FS` that supports reading, writing, mknod, chmod, chown
* rwosfs: an implementation of `rwfs` that uses an underlying filesystem dir for implementation; similar to `os.DirFS` for `fs.FS`.
* an update to the goheader `header.tmpl` to support any whitespace before the apache URL; some editors change the format and it fails on tab vs space vs 2 space vs space+tab vs I have no idea what. This makes it much more tolerant.
* updates to `pkg/vfs/` so it will work on non-Linux platforms, or at least compile.

There is some extensive documentation in `docs/apk.md`.

Let the comments begin. 😄 

Something important: I modified the original version of the initdb part based on the work in #412. Thanks to @puerco's PR for some good ideas there.


deitch:
I already pulled the header template changes into their own PR #427

deitch:
There are some arbitrary or hard-coded things that could use some flexibility - like the base OS version, or alpine-keys version. I would be happy to get some good ideas on the best approach for those.

deitch:
Ah, nice. go tests and lint are clean.

pkg/apk/impl/install.go
```diff
@@ -0,0 +1,99 @@
...
+				return nil, fmt.Errorf("error creating file %s: %w", header.Name, err)
+			}
+			defer f.Close()
+			if _, err := io.CopyN(f, tr, header.Size); err != nil {
+				return nil, fmt.Errorf("unable to write content for %s: %w", header.Name, err)
+			}
+		case tar.TypeSymlink:
+			// some underlying filesystems and some memfs that we use in tests do not support symlinks.
+			// attempt it, and if it fails, just copy it.
+			if err := a.fs.Symlink(header.Linkname, header.Name); err != nil {
```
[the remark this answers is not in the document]
kaniini: I think we should definitely validate the linknames before doing the symlink operation, in this warning and the one below.

pkg/apk/impl/sign.go
```diff
@@ -0,0 +1,150 @@
+// Copyright 2022 Chainguard, Inc.
```
kaniini, line 1: we should upstream this to alpine-go

kaniini:
This looks good as a first cut, but one thing I noticed is missing is an implementation of the solver algorithm.  Am I just missing it, or is this implementation unable to solve the full DAG given an `/etc/apk/world` file still?

pkg/apk/impl/rwfs/rwfs.go
```diff
@@ -0,0 +1,44 @@
...
+	fs.FS
+	fs.ReadDirFS
+	fs.StatFS
+	fs.ReadFileFS
+	OpenFile(name string, flag int, perm fs.FileMode) (File, error)
+	Mkdir(string, fs.FileMode) error
+	MkdirAll(string, fs.FileMode) error
+	WriteFile(string, []byte, fs.FileMode) error
+	Mknod(path string, mode uint32, dev int) error
+	Symlink(oldname, newname string) error
```
kaniini, line 35: We definitely need the ability to track permissions and ownership somehow, to ensure apko does not need to run as root.  This was the main reason for `pkg/vfs`.

pkg/apk/impl/repo.go
```diff
@@ -0,0 +1,443 @@
...
+	// each dependency has several possibilities:
+	// - /bin/sh   - this means, "I need package busybox"
+	// - name      - this means, "I need package 'name'"
+	// - so:<lib>  - this means, "find the package that provides so:<lib>"
+	// - cmd:<cmd> - this means, "find the package that provides cmd:<cmd>"
+	for _, dep := range ref.Dependencies {
+		var newDep string
+		switch {
+		case dep == "/bin/sh":
+			newDep = "busybox"
```
kaniini, line 243: `/bin/sh` is a virtual, which is presently provided by `busybox`, but might be provided by something else in the future.  I think we should not have special handling for it.

The upstream `apk` solver algorithm looks for a direct package which matches `apk_name`, and then falls back to looking for packages which *provide* the `apk_name`.  This would remove the need for special casing `so:` and `cmd:` in your algorithm as well (and there other namespaces like `py:`, `ruby:`, `pc:` in some apk distributions).

pkg/apk/impl/repo.go
```diff
@@ -0,0 +1,443 @@
...
+	parts = strings.SplitN(name, "<", 2)
+	if len(parts) == 2 {
+		// we have a less than but not equal
+		return strings.TrimSuffix(parts[0], "<"), parts[1], versionLess
+	}
+	// we have no =, < or >, so we just return the name
+	return name, "", versionNone
+}
+
+// compare semver versions
```
kaniini, line 322: apk versions aren't semver, but rather derived from Gentoo versions.  See https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-250003.2 for details about the version grammar and comparison rules.

deitch: Ah you know, I was wondering about that. I didn't bother digging too deeply, figuring that since it is all in a single function, I can fix it to whatever is "correct" (after I find the definition of "correct") later. This helps greatly, thank you. I will update it.

pkg/apk/impl/repo.go
deitch: It is clear you have dug more deeply into how it does it than I have (but the gap has narrowed, if only a bit).

My understanding was that the algorithm is:

1. dep is `!<something>` (special case)
1. dep is package name
1. dep is a special case (i.e. `/bin/sh`)
1. dep is something provided by a package, when it has the `so:` or `cmd:` cases

You are saying (I think):

1. dep is `!<something>` (special case)
1. dep is package name
1. dep is any package that _provides_ that dependency

The last case covers package alternate names, but also `/bin/sh` (since it is included in the `p:` of busybox), but also the `so:` and `cmd:` etc.

Is that correct?

pkg/apk/impl/sign.go
deitch: Just the sign changes? Or the whole `pkg/apk/impl`?

pkg/apk/impl/rwfs/rwfs.go
deitch: The approach I took was to track all target permissions and ownership at each stage, then add it to the tar file later as overrides. That definitely works. But, at some point I did think about doing away with that and replacing it with the vfs/rwfs layer and have that be responsible for tracking. Even if it is just wrapping memfs or os.DirFS or anything else, let it handle the above.

I almost did it, but had some concerns. First, I wanted everything else to be in the right place and looking good before revamping that part. Second, I was concerned about resource usage. The FS likely would need to track it all in memory, which can get big. It shouldn't get too big, but it can get big. Or it can add a "magic file" somewhere in its filesystem, but I don't like that much. Third, I was worried that if it stopped in the middle, we wouldn't be able to pick it up again.

But then I realized that I already am putting some things in memory, and premature optimization is not something to worry about - when it gets really resource hungry, we can look at alternatives, not before - and we pretty much clean it all up if we stop in the middle anyways. So why not?

To take an approach: are you suggesting we switch from my current implemented approach of, "reconstruct what we need" (based on permissions in `installed`, home dir overrides, etc.) to "have the FS track it all"?

pkg/apk/impl/const.go
```diff
@@ -0,0 +1,35 @@
...
+	reposFilePath     = "etc/apk/repositories"
+	archFilePath      = "etc/apk/arch"
+	keysDirPath       = "etc/apk/keys"
+	worldFilePath     = "etc/apk/world"
+	installedFilePath = "lib/apk/db/installed"
+	scriptsFilePath   = "lib/apk/db/scripts.tar"
+	scriptsTarPerms   = 0o644
+	triggersFilePath  = "lib/apk/db/triggers"
+
+	// for fetching the alpine keys
```
kaniini, line 31: I don't think we should depend on a random package for a set of keys, since they can rotate frequently.  Instead, we should talk with the Alpine project about implementing key bundles that can be fetched from a well-known location.

cc @ikke for ideas here.  Perhaps alpine-go can have a package for fetching keys?

deitch: Yeah I wasn't enamoured of it either, but it got it to work. I would be very happy with a better alternative.

Ikke: Would that do more then just doing an HTTP GET from example here? https://alpinelinux.org/keys/

deitch:
@Ikke great to meet you. If you have a moment, can you look at what is in here, specifically under `pkg/apk/impl`? All of that (incomplete but mostly functional) implementation of local `apk` commands. It is built to be extracted from this project into its own, whether `alpine-go` or a separate library? See what is appropriate for being part of official alpine go libs?

pkg/apk/impl/const.go
deitch: @Ikke yeah I think that would work. 

The behaviour, though, is slightly different, in that `alpine-keys` package installs them all into `/usr/share/apk/keys/`, with symlinks for the relevant ones for each platform in `<arch>/`. Then the install places the ones just for the given platform in `/etc/apk/keys/`.

If I get them directly from https://alpinelinux.org/keys/ , I just have the keys, no platform-specific ones. Would I install them all in `/etc/apk/keys`?

pkg/apk/impl/repo.go
deitch: Completely revamped to follow the logic on that link (and includes the link), and lots of test cases for it in a table.

jdolitsky:
I think a lot of this can make its way into alpine-go (if desired), but I think it can treated as a separate task after this is merged.

pkg/apk/impl/const.go
Ikke: I'm thinking about proposing adding the keys that are used per arch to https://alpinelinux.org/releases.json, would that work?

deitch: It would, I believe. Do you mean the actual keys, or just the names of the keys as references?

Ikke: names and urls where to find them.

deitch: Sounds pretty good.

Should I wait for it? What's your estimate on when it will go in? And any way we can help?

Ikke: https://gitlab.alpinelinux.org/alpine/infra/alpine-mksite/-/merge_requests/58

deitch: Cool, thanks. I signed up for notifications on that MR. Once that is in, we will consume it.

Ikke: I will also update gitlab.alpinelinux.org/alpine/go/releases to include this data

deitch: @Ikke thanks for merging that in. I am not quite sure what the structure of `releases.json` is. I _think_ that what I need to do is, "look for value of `".keys"` in the `release_branches` array where `".rel_branch" == desired`, essentially, if I am looking for the aarch64 keys for 3.16:

```
curl https://alpinelinux.org/releases.json | jq -r '.release_branches[] | select(.rel_branch == "v3.16") | .keys["aarch64"][].url'
```

Is that correct?

deitch: OK, got it to work, all looks good. No longer trying to install arbitrary `alpine-keys` package.

I will mark this as "Resolved", but @Ikke you might want to look at the implementation in `pkg/apk/impl/`; we can upstream some stuff from there.

pkg/apk/impl/rwfs/rwfs.go
kaniini: Either approach works.  If you have full support for the apkdb, that seems fine to me!

pkg/apk/impl/repo.go
kaniini: The `!` just means "conflicts with <thing>".  Otherwise your analysis is correct.

deitch: @kaniini I cleaned it up. The logic now is essentially what is above.

With one variant.

I had to add explicit support for the virtual `/bin/sh`. Without it, anything that provides `/bin/sh` could satisfy it, which leads to strange loops. For example:

1. I already have `busybox`, which provides `/bin/sh`
1. I want to install `dash-binsh`, which also provides `/bin/sh`
2. That depends on `dash`
3. That depends on `/bin/sh`

So do I satisfy it with `busybox` or `dash-binsh`. We know that it means the baseline `/bin/sh`, or perhaps anything that provides it. The dependency is created because `dash` has a pre-install script that requires `/bin/sh`, but how would this resolve it?

I guess it is possible to say, "as long as something has provided it, we are good"?

An alternative scenario is where I install something that depends on `/bin/sh`, e.g. `dhcp`. When it asks for `/bin/sh`, if there is none installed yet, do I install `busybox`? `dash`? `yash`?

pkg/apk/impl/rwfs/rwfs.go
deitch: If you are ok with it, I am going to punt on that change for a bit. I do like the elegance of doing it all in FS, but what we have now works. If everything else is ok, when we are done - either as part of this PR or as a follow-on - I am happy to switch it.

pkg/apk/impl/sign.go
kaniini: I would say at least the signing, but we should do it incrementally.

pkg/apk/impl/install.go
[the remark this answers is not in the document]
deitch: Done. Hopefully this will pass the test now.

pkg/build/ldconfig.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	apkofs "chainguard.dev/apko/pkg/fs"
+)
+
+// original alpine ldconfig available at https://git.alpinelinux.org/aports/tree/main/musl/ldconfig
```
kaniini, line 27: For Wolfi this needs to run proper `/sbin/ldconfig` because of glibc `/etc/ld.so.cache`.

pkg/apk/impl/sign.go
deitch: We had a bit of a discussion on slack on this. I think the consensus was not to hold out for upstreaming, but rather to do it in chainguard-dev org, and then eventually upstream it.

deitch:
I see one lint and one lint/codeql errors, not sure what to do with them.

* lint is complaining about the copyrights; I fixed them on every file we changed, and running `golangci-lint run` locally shows no copyright errors
* both lint and codeql are complaining about the symlink arbitrary file write, but it already was evaluated

deitch:
Actually, it is odder than that. `EvalSymlinks` will not help, as it validates them against the _local_ filesystem, but we are using them internally in an `fs.FS`. So it doesn't do any good, and actually fails.

I think we may need to bypass it, unless someone has a better idea?

pkg/apk/impl/const.go
deitch: @Ikke is there someone I can find a reference on signature validation? It looks like the algorithm changed somewhere down the line. I see the larger vs smaller keys (451 bytes in the PEM file vs 800 bytes, which I assume are the 4096 RSA keys), and larger vs smaller signatures (512 vs 256). Yet was the algorithm different prior to the longer keys? We are using a fairly straightforward logic that works for the 4096, but not for the older keys. Simplified below:

```go
func RSAVerifySHA1Digest(sha1Digest, signature []byte, publicKey []byte) error {
	block, _ := pem.Decode(publicKey)
	pub, err := x509.ParsePKIXPublicKey(block.Bytes)
	rsaPub, ok := pub.(*rsa.PublicKey)
	return rsa.VerifyPKCS1v15(rsaPub, crypto.SHA1, sha1Digest, signature)
}
```

deitch: FWIW file sizes. I can verify the `aarch64.pub` and `aarch64.sig` but not the `riscv64` counterparts.

```
-rw-r--r--   1 user  wheel     800 Jan  4 12:47 aarch64.pub
-rw-r--r--   1 user  wheel     512 Jan  4 13:01 aarch64.sig
-rw-r--r--   1 user  wheel     451 Jan  4 12:47 riscv.pub
-rw-r--r--   1 user  wheel     256 Jan  4 13:01 riscv64.sig
```

kaniini:
Bypassing it is probably fine for now.

jdolitsky:
On an M1 mac, and ran the following:
```
mkdir tmp
cd tmp
git clone https://github.com/deitch/apko.git -b apk-cli-free .
make apko
curl -sLO https://raw.githubusercontent.com/chainguard-images/images/main/images/wolfi-base/configs/latest.apko.yaml
./apko build latest.apko.yaml wolfi-base output.tar
```

and get an error related to symlinks:
```
Jan  3 17:17:49.041 [INFO] [arch:aarch64] synchronizing with desired apk world
Error: failed to build layer image: initializing apk: failed to fixate apk world: unable to install files for pkg wolfi-baselayout: invalid symlink /proc/mounts: lstat /proc: no such file or directory
2023/01/03 17:17:49 error during command execution: failed to build layer image: initializing apk: failed to fixate apk world: unable to install files for pkg wolfi-baselayout: invalid symlink /proc/mounts: lstat /proc: no such file or directory
```

deitch:
@jdolitsky yeah, it was that `EvalSymlinks()` that caused the issue. It evaluates in the current filesystem, which is not what we want. I removed it and pushed out, so pull the branch and try again please?

deitch:
@kaniini do you recall the "ignore this syntax" offhand?

pkg/apk/impl/const.go
Ikke: I'm not aware of any algorithm change, only that we recently created new 4096-bits keys. As riscv64 was generated on a different moment. `abuild-keygen` uses `openssl genrsa`, so we'd have to look at openssl to see if anything changed.

deitch: OK. If there are pointers you can share, it would be appreciated.

Ikke: Is this what you mean?

riscv64:
```
$ openssl rsa -pubin -in alpine-devel@lists.alpinelinux.org-60ac2099.rsa.pub -noout -text | head -n1
Public-Key: (2048 bit)
```

aarch64:
```
$ openssl rsa -pubin -in alpine-devel@lists.alpinelinux.org-616ae350.rsa.pub -noout -text | head -n1                                                                     
Public-Key: (4096 bit)
```

riscv64 is a 2048-bits key, while aarch64 is a 4096-bits key.

jdolitsky:
@deitch thanks. after re-doing above commands I get passed the symlink error but now have:
```
Jan  4 10:13:38.279 [INFO] [arch:aarch64] synchronizing with desired apk world
Error: failed to build layer image: initializing apk: failed to fixate apk world: unable to update scripts.tar for pkg glibc: unable to write scripts header for glibc-2.36-r3.Q1/ido5cSLfh1lALe6GCoMfKXSf6Q=.trigger: archive/tar: cannot encode header: Format specifies USTAR; and USTAR cannot encode ChangeTime=1969-12-31 18:00:00 -0600 CST
2023/01/04 10:13:39 error during command execution: failed to build layer image: initializing apk: failed to fixate apk world: unable to update scripts.tar for pkg glibc: unable to write scripts header for glibc-2.36-r3.Q1/ido5cSLfh1lALe6GCoMfKXSf6Q=.trigger: archive/tar: cannot encode header: Format specifies USTAR; and USTAR cannot encode ChangeTime=1969-12-31 18:00:00 -0600 CST
```

jdolitsky:
however works fine with alpine-base !!

jdolitsky:
using https://raw.githubusercontent.com/chainguard-images/images/main/images/alpine-base/configs/latest.apko.yaml

and doing docker load we have a different issue:
```
$ docker -D load < output.tar 
492d1793c47d: Loading layer [==================================================>]  3.249MB/3.249MB
Error processing tar file(duplicates of file paths not supported):
```

deitch:
@jdolitsky take another look w/r/t the USTar issue. I will look at the docker load next. Docker really should be happy to accept sequential tar files that overwrite, but we can fix that if it does not.

deitch:
The mink failure is because of riscv64, which has to do with the different key verification algorithm. Hopefully we will know what that is soon enough.

Ikke:
> The mink failure is because of riscv64, which has to do with the different key verification algorithm. Hopefully we will know what that is soon enough.

Is it the size that is the issue (2048-bits instead of 4096)? Because if so, there are still packages on edge that are signed with the 2048-bits keys.

deitch:
@Ikke here you go, [this gist](https://gist.github.com/deitch/f14dce581c9d779bf3b0ca02dfb77594). I must be missing _something_

jdolitsky:
update - wolfi-base appears to build, but the docker load still failing with the same

deitch:
> update - wolfi-base appears to build, but the docker load still failing with the same

I had no problem with the load, I 

```sh
$ curl -sL https://raw.githubusercontent.com/chainguard-images/images/main/images/alpine-base/configs/latest.apko.yaml -o /tmp/alpine-latest.apko.yaml
$ go run . build /tmp/alpine-latest.apko.yaml alpine-base /tmp/alpine-base.tar
$ cat /tmp/alpine-base.tar | docker -D load
$
```

Same if I build with `https://raw.githubusercontent.com/chainguard-images/images/main/images/wolfi-base/configs/latest.apko.yaml`

deitch:
Had to rebase with a lot of things, but seems to be back in the right place.

As far as I can tell, we have two outstanding issues:

1. The signature verification issue with riscv64; hopefully with the gist, @Ikke will be able to shed some light on it.
2. The docker load issue; hopefully @jdolitsky will be able to replicate it.

Ikke:
> @Ikke here you go, [this gist](https://gist.github.com/deitch/f14dce581c9d779bf3b0ca02dfb77594). I must be missing _something_

Not anything concrete yet, but if I add ppc64le (which happens to use 2048-bits keys atm), it fails as well, so it's something to do with 2048-bits keys, but not sure why that would be failing yet.

deitch:
Yeah, strange, is it not? Probably something tiny that we are just missing.

deitch:
OK, I think I have something.

I changed the calculation of what to use for the hash (and then compared to the signature) to this:

```go
        allBytes := len(b)
        unreadBytes := buf.Len()
        readBytes := allBytes - unreadBytes
        indexData := b[readBytes:]
```

Same functionality, but it gives me the ability to inspect the byte count. I got the following:

* aarch64 (works): readBytes = 666
* riscv64 (fails): readBytes = 391

I think tried chopping off the first 666 (aarch64) or 391 (riscv64) bytes from each file and seeing if it still is valid tgz file, which it should be. First aarch64, which works:

```sh
$ dd if=APKINDEX-aarch64.tar.gz of=APKINDEX-aarch64-part.tar.gz ibs=1 skip=666
656533+0 records in
1282+1 records out
656533 bytes transferred in 0.358321 secs (1832248 bytes/sec)
$ ls -la *aarch*
-rw-r--r--  1 adeitcher  wheel  656533 Jan  8 17:54 APKINDEX-aarch64-part.tar.gz
-rw-r--r--  1 adeitcher  wheel  657199 Jan  8 17:26 APKINDEX-aarch64.tar.gz
$ tar tvf APKINDEX-aarch64-part.tar.gz
-rw-r--r--  0 root   root       24 Jan  8 14:28 DESCRIPTION
-rw-r--r--  0 root   root  1951912 Jan  8 14:28 APKINDEX
```

That looks good. 

Next riscv64 (which fails):

```sh
$ dd if=APKINDEX-riscv64.tar.gz of=APKINDEX-riscv64-part.tar.gz ibs=1 skip=391
643376+0 records in
1256+1 records out
643376 bytes transferred in 0.359835 secs (1787975 bytes/sec)
$ ls -l *riscv*
-rw-r--r--  1 adeitcher  wheel  643376 Jan  8 17:53 APKINDEX-riscv64-part.tar.gz
-rw-r--r--  1 adeitcher  wheel  643767 Jan  8 17:26 APKINDEX-riscv64.tar.gz
$ tar tvf APKINDEX-riscv64-part.tar.gz
tar: Error opening archive: Unrecognized archive format
```

Oops. Something about either the calculation, or how we are using go to read the unread bytes, is failing.

I am pretty sure the logic is correct. Treat it as two concatenated gzip streams, read the first as the signature, lop it off, hash the rest, verify the signature. It is the "calculation of lop it off" that is wrong.

I then used xxd to look at the file in that area for the gzip stream typical header magic number `0x1f8b`. I found it at 399 instead of 391. So I tried again:

```sh
$ dd if=APKINDEX-riscv64.tar.gz of=APKINDEX-riscv64-part.tar.gz ibs=1 skip=399
643368+0 records in
1256+1 records out
643368 bytes transferred in 0.364523 secs (1764959 bytes/sec)
$ tar tvf APKINDEX-riscv64-part.tar.gz
-rw-r--r--  0 root   root       24 Jan  8 14:28 DESCRIPTION
-rw-r--r--  0 root   root  1920522 Jan  8 14:28 APKINDEX
```

That worked! OK, so the calculation is off by 8 bytes. But why and how?

deitch:
Yup. And if I force it with:

```go
        if readBytes == 391 {
                readBytes = 399
        }
```

then it all works. So this is a gzip stream reading error. Figure that out, and it works. Still no idea why, though.

deitch:
More. I stepped through the code, checking the `buf.off` which is the offset, essentially the amount read. I get the following.

aarch64:
* after `signatureFile, err := tarReader.Next()`: 658
* after `signature, err := io.ReadAll(tarReader)`: 666

riscv64:
* after `signatureFile, err := tarReader.Next()`: 391
* after `signature, err := io.ReadAll(tarReader)`: 391

Some sort of caching it is doing? 🤷‍♂️

deitch:
Aha! Got it! Successful day! Unless you call `Next()` and check for an `EOF` (with multistream), it will not complete the reading. We simply got lucky with the longer ones. Try the gist now.

deitch:
Phew. riscv64 works now.

@jdolitsky , tell me how to replicate the docker load issue - exactly what build it was - and I will figure that out.

Ikke:
> Aha! Got it! Successful day! Unless you call `Next()` and check for an `EOF` (with multistream), it will not complete the reading. We simply got lucky with the longer ones. Try the gist now.

Ah yes, nice find!

deitch:
mink test is failing. I cannot quite figure out what build command it is executing, as it does not seem to list anywhere, "this is the command I am going to execute".

I think it is `apko publish` something, but no really sure. Can someone describe what it is?

jdolitsky:
regarding the `docker load` issue, here's how I reproduce (similar to comment [here](https://github.com/chainguard-dev/apko/pull/426#issuecomment-1370894881))
```
mkdir tmp
cd tmp
git clone https://github.com/deitch/apko.git -b apk-cli-free .
make apko
curl -sLO https://raw.githubusercontent.com/chainguard-images/images/main/images/wolfi-base/configs/latest.apko.yaml
./apko build latest.apko.yaml wolfi-base output.tar
docker -D load < output.tar
```

Error:
```
a53226dbe638: Loading layer [==================================================>]  4.677MB/4.677MB
Error processing tar file(duplicates of file paths not supported): 
```

 In fact, even if I follow your instructions [here](https://github.com/chainguard-dev/apko/pull/426#issuecomment-1374563269) I get the same error.. maybe we can sit down this week to look at it if needed. I'm using Docker Desktop on M1 mac

jdolitsky:
mink error seems to be part of the apko build:
```
  Error: failed to build layer image for "arm/v7": failed to generate tarball for image: writing TAR archive failed: mknod not supported by this fs: path (dev/console)
  2023/01/08 17:56:43 error during command execution: failed to build layer image for "arm/v7": failed to generate tarball for image: writing TAR archive failed: mknod not supported by this fs: path (dev/console)
  ```
  
unclear to me right now what that means.. but perhaps indicates a difference in how tar is built

**Comment by deitch** [comment]
> In fact, even if I follow your instructions https://github.com/chainguard-dev/apko/pull/426#issuecomment-1374563269 I get the same error.. maybe we can sit down this week to look at it if needed. I'm using Docker Desktop on M1 Mac

My Mac has docker with experimental containerd backend support, so it is possible mine doesn't have an issue with duplicate paths. But I managed to reproduce it on a different installation.

**Comment by deitch** [comment]
> mink error seems to be part of the apko build
> unclear to me right now what that means.. but perhaps indicates a difference in how tar is built

I think it calls `apko publish`, and thus does a slightly different code path. I am unsure, maybe @kaniini knows? If I could get the proper invocation, I can track it down.

**Comment by jdolitsky** [comment]
My guess is the mink task definition is here: https://github.com/chainguard-dev/apko/blob/main/config/task.yaml#L45

**Comment by kaniini** [comment]
`mknod` needs to be supported (via faking) for all FSes.  we have to support anything that can be in a tar stream.

**Comment by deitch** [comment]
@jdolitsky can you try the `docker load` again with the latest commits here? I reworked it.

**Comment by jdolitsky** [comment]
@deitch the load works fine with wolfi-base!

**Comment by jdolitsky** [comment]
Was trying to build other images to see if the image can run, and ran into another issue at least for the following 2 configs:
- https://raw.githubusercontent.com/chainguard-images/images/main/images/python/configs/latest-dev.apko.yaml
- https://raw.githubusercontent.com/chainguard-images/images/main/images/ruby/configs/latest-3.2.apko.yaml

```
Error: failed to build layer image: initializing apk: failed to fixate apk world: unable to install files for pkg ncurses-terminfo-base: unsupported file type 49
2023/01/09 17:06:54 error during command execution: failed to build layer image: initializing apk: failed to fixate apk world: unable to install files for pkg ncurses-terminfo-base: unsupported file type 49
```

The melange source for that wolfi package is here: https://github.com/wolfi-dev/os/blob/main/ncurses.yaml

and .apk can be found here: https://packages.wolfi.dev/os/aarch64/ncurses-terminfo-base-6.3-r2.apk

**Comment by deitch** [comment]
With #458 now in, I rebased it so it can use those checks, but removed the tests that build the examples in docker; after all, the whole purpose of this PR is to run it all natively without calling out.

**Comment by deitch** [comment]
This almost is good. One build image is failing, `./examples/wolfi-base.yaml`. The reason is as follows. The order of packages to be installed is:

1. wolfi-keys
2. glibc
3. busybox
4. lib
5. libcrypto3
6. libssl3
7. wolfi-baselayout

The problem is that the second one, glibc, installs `/lib64` as a directory and puts some things in there, while `wolfi-baselayout` tries to create a symlink `/lib64 -> /lib`. That fails because `/lib64` already exists as a directory.

I am not sure how this worked in the past, should it not have failed? Does it ignore the link? Or did the ordering here get messed up?

FWIW, I ran cgr.dev/chainguard/wolfi-base and it has `/lib64 -> /lib`.

**Comment by deitch** [comment]
While looking at it, in the actual image now:

```sh
# cat /etc/apk/world
ca-certificates-bundle
wolfi-base
wolfi-baselayout
```

while the example has:

```yaml
contents:
  keyring:
  repositories:
  packages:
    - ca-certificates-bundle
    - wolfi-base
```

The dependency tree looks as follows (I walked it through manually based on https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz)

- ca-certificates-bundle :
- wolfi-base
   - apk-tools
      - ca-certificates-bundle
      - glibc
         - wolfi-baselayout
      - libcrypto3
      - libssl3
      - zlib
   - busybox
   - wolfi-keys

So wolfi-base -> glibc -> wolfi-baselayout.

That means that it should have executed wolfi-baselayout first, which would make the lib64 symlink exist. Ordering clearly is a bit off, but we will get it.

deitch:
Only complaints here are false positives from CodeQL, which, apparently, has no inline code suppression, only "dismiss from the UI". codeql misses that the calls are on sanitized paths.

jdolitsky (approved):

jdolitsky:
I've gone ahead and marked the CodeQL alerts as "false positive"

deitch:
> I've gone ahead and marked the CodeQL alerts as "false positive"

Thanks Josh. I would mark it inline in the code, but apparently codeql does not support it 🤷‍♂️

kaniini:
Going ahead and merging this, and lets see what breaks.

PR #463, opened 2023-01-19
fix some build-breaking bugs with the native apk implementation

kaniini:
With these fixes, apko should be able to build all images again without using apk-tools.


imjasonh (approved):

deitch:
Thank you for fixing these @kaniini

PR #465, opened 2023-01-19
read keyfiles from local; read local repositories

### LINKED ISSUE #464: failed to read apk key: open /tmp/apko-3984880358/work/local-melange.rsa.pub: no such file or directory
When building locally apko has started to fail this morning.

```
failed to read apk key: open /tmp/apko-3984880358/work/local-melange.rsa.pub: no such file or directory
```

```
--keyring-append /work/local-melange.rsa.pub --repository-append /work/packages
```

full command...
```
docker run --rm -v ${PWD}:/work -w /work cgr.dev/chainguard/apko build latest.apko.yaml foo foo.tar --debug --build-arch amd64 --keyring-append /work/local-melange.rsa.pub --repository-append /work/packages
```

I was able to go back to a previous apko version and things worked again.
```
ghcr.io/chainguard-dev/apko:cf9e6f264bcf38fd8ebbd95fb6a954cce5d16b57
```

deitch:
Fixes #464 

Issue with two problems was a regression caused by #426 

It was treating keyfiles to read when passed via `--keyring-append` as relative to workdir, which made no sense; when someone types `build --keyring-append /path/to/file`, they mean, "relative to where I am running apko".

This fixes it.

It also was doing the same for `--repository-append`; fixes that too.

### DIFF
--- a/pkg/apk/impl/implementation.go
+++ b/pkg/apk/impl/implementation.go
@@ -338,7 +338,7 @@ func (a *APKImplementation) InitKeyring(keyFiles, extraKeyFiles []string) (err e
 			var data []byte
 			switch asURL.Scheme {
 			case "file":
-				data, err = a.fs.ReadFile(element)
+				data, err = os.ReadFile(element)
 				if err != nil {
 					return fmt.Errorf("failed to read apk key: %w", err)
 				}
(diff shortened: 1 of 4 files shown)


deitch:
I do have an issue with the way this is done, would be good to get some feedback @kaniini 

There are two kinds of repos we add to `/etc/apk/repos`:

* URLs, e.g. `https://dl-cdn.alpinelinux.org/alpine/v3.16/main`
* Paths added via `--repository-append /some/path/to/repo`

For the latter, we add it to `/etc/apk/repos` as is, i.e. `/some/path/to/repo`. This PR adds the logic to find those and read from the local filesystem where it is running, not inside the workdir, which works.

But it does leave an odd and non-usable path in `/etc/apk/repos`. I think this existed before, so we aren't adding any new regressions, but this is not exactly very clean. My vote is to leave it as is for this and get this fix in, but we should fix it at some point.

jdolitsky (approved):

PR #471, opened 2023-01-22
replace filesystem access with writable fs.FS

deitch:
This replaces lots of different fs.FS and implementations at various levels with a single "Full FS", i.e. an `fs.FS` implementation for full read-write capability, plus handling things that some filesystems might not, like chmod, chown, mknod, etc.

There are 2 implementations of it. One is memfs, the other is based on a disk directory.

The disk one keeps a copy of all of the files and directories - but **not** the contents - in memory, which means that once you exit the program, the contents of the FS cannot be recreated fully. However, we pretty much delete all of these things right away anyways, so that should be fine. An alternative implementation could be plugged in, one that would use "special files" on disk to store that information. 

The important parts here are:

* we no longer put everything in memory when merging filesystems
* we no longer create different filesystems and then merge them; instead, it is all done in a single FS
* `BuildImage` returns that filesystem, which, in turn, can be used at any point. We no longer have to make a tarball layer in order to get at all of the additional information, which should help solve [this PR](https://github.com/chainguard-dev/melange/pull/263)
* This all is a lot cleaner; we just work with an FS like it is a normal fully-powered Linux filesystem as root, even if we are not


deitch:
The build is failing when trying to build wolfi-base. Here is why it is failing, but I am not sure how to fix it. Two packages are trying to modify `/lib64`: one is trying to make it a symlink to `/lib`, the other is trying to make it a directory.

The build file is (simplified):

```yaml
contents:
  keyring:
  repositories:
  packages:
    - ca-certificates-bundle
    - wolfi-base
```

I went through those packages as follows:

1. download the index https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz
2. Look at the dependencies of each package, eventually building a dependency tree
3. Look at the contents of each 

`ca-certificates-bundle` does not affect that dir or have any dependencies.

The relevant dependency treeis:

```
wolfi-base
    apk-tools
        so:libc.so.6 (provided by glibc)
            wolfi-baselayout
```

wolfi-base -> apk-tools -> glibc -> wolfi-baselayout

glibc includes the following in its apk file (simplified):

```
drwxr-xr-x  0 root   root        0 Jan  1  1970 lib64
```

wolfi-baselayout includes the following:

```
drwxr-xr-x  0 root   root        0 Jan  1  1970 lib
lrwxrwxrwx  0 root   root        0 Jan  1  1970 lib64 -> lib
```

This causes the failure. By the time glibc installs, it tries to `mkdir /lib64`, but it fails because a symlink already exists. It wouldn't help if it ran in the other direction either.

In theory, we could have it do, "if I try to mkdir, and it already exists as symlink to a directory, then treat it like it already succeeded," but that is messy. I wonder if we got something wrong with the conflict between glibc and wolfi-base/wolfi-baselayout, or conversely if I just misunderstood how to resolve these.

@kaniini or @jdolitsky any ideas?

kaniini:
this is intentional: LSB requires the stupid lib64 directory, but we don't want it.  but every build system on the planet wants to use it.

wolfi-baselayout installs first, and creates the symlink, and apk-tools just follows the symlink.  it is an intentional feature of apk that we can do this :)

deitch:
Hopefully this will catch it. If it encounters a need to install a directory, and that target name already is a symlink pointing to a directory, it will accept it.

I had to implement all sorts of symlink resolution stuff, because the underlying memfs does not, but it looks good.

kaniini (approved):

PR #473, opened 2023-01-22
add versions with constraints

### LINKED ISSUE #470: native apk implementation: does not solve for highest constrained version
Previously, the apk solver would choose the first package it encountered which met the constraints.  @jdolitsky changed this in #468 to use the highest version encountered as an interim improvement, but ultimately the solver should pick the highest version which meets the constraint.

Consider the following repository:

- `Package{Name: 'foo', Version: '1.2.3-r0'}`
- `Package{Name: 'foo', Version: '1.3.6-r0'}`
- `Package{Name: 'foo', Version: '1.2.8-r0'}`
- `Package{Name: 'foo', Version: '1.7.1-r0'}`
- `Package{Name: 'foo', Version: '1.7.1-r1'}`
- `Package{Name: 'foo', Version: '2.0.6-r0'}`

If we specify in our config `foo>2.0`, then `foo-2.0.6-r0` should be selected.
If we specify in our config `foo~1.7`, then `foo-1.7.1-r1` should be selected.
If we specify in our config `foo<1.7`, then `foo-1.3.6-r0` should be selected.
If we specify in our config `foo<1.3`, then `foo-1.2.8-r0` should be selected.
If we specify in our config `foo<1.2.8`, then `foo-1.2.3-r0` should be selected.

deitch:
Fixes #470 

@kaniini can you validate that this actually resolves what you were looking for? Specifically, look at the test case for version resolution [here](https://github.com/chainguard-dev/apko/pull/473/files#diff-0fe948ffc19f6666deee62dfea127804fc3033d1c7ad2971b1178715a50735bdR94).

### DIFF
--- a/pkg/apk/impl/repo.go
+++ b/pkg/apk/impl/repo.go
@@ -204,46 +204,42 @@ func (a *APKImplementation) getRepositoryIndexes(ignoreSignatures bool) ([]*repo
 // If the indexes change, you should generate a new pkgResolver.
 type PkgResolver struct {
 	indexes     []*repository.RepositoryWithIndex
-	nameMap     map[string]*repository.RepositoryPackage
+	nameMap     map[string]map[string]*repository.RepositoryPackage
 	providesMap map[string]string
 }
 
 // NewPkgResolver creates a new pkgResolver from a list of indexes.
 func NewPkgResolver(indexes []*repository.RepositoryWithIndex) *PkgResolver {
 	var (
-		pkgNameMap     = map[string]*repository.RepositoryPackage{}
+		pkgNameMap     = map[string]map[string]*repository.RepositoryPackage{}
 		pkgProvidesMap = map[string]string{}
 	)
 	p := &PkgResolver{
 		indexes: indexes,
 	}
-	// create a map of every package to its RepositoryPackage
+	// create a map of every package by name and version to its RepositoryPackage
 	for _, index := range indexes {
 		for _, pkg := range index.Packages() {
 			existingPkg, ok := pkgNameMap[pkg.Name]
 			if !ok {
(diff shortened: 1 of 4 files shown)


deitch:
Got it rebased, it messed up the order of package installation. I will figure it out, resolve, and push it out.

kaniini:
Sounds good! :+1:

deitch:
OK, fixed that. Letting CI have a run and tell us, and good for your review @kaniini

deitch:
It's a really good thing I added that `examples/alias-only.yaml`, and added testing all examples in CI. It found a regression in this change (which I have fixed, of course).

deitch:
I always like "clean and green" CI 😁

pkg/apk/impl/repo.go
```diff
@@ -310,10 +278,37 @@ func (p *PkgResolver) GetPackagesWithDependencies(packages []string) (toInstall
...
+		}
+		pkgsWithVersions, ok = p.nameMap[trueName]
+		if !ok {
+			return nil, nil, nil, fmt.Errorf("looking for %s, found provided by %s, but could not find %s in indexes", pkgName, trueName, trueName)
+		}
+	}
+
+	// pkgsWithVersions contains a map of all versions of the package
+	// get the one that most matches what was requested
+	versions := make([]string, 0, 10)
```
jdolitsky, line 302: significance of size 10? imagine this could grow bigger?

deitch: Nothing really, just a starting size from which it could grow. lint insists you give it something, 10 seemed as reasonable as any other number.

jdolitsky (approved):

kaniini (approved):

PR #484, opened 2023-02-02
deprecate build-arch and use-arch

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

PTAL @kaniini @strongjz


kaniini:
#499 did the same thing, as part of a larger refactor, so I merged it instead.

PR #487, opened 2023-02-07
handle when file already exists

deitch:
The case-sensitive extension for filesystems in rwosfs has a func `createOnDisk()` that determines (only when running on an underlying filesystem that is case-insensitive) whether a particular filename should be created on disk (first one of this name in any case) or in the memory overrides.

`createOnDisk()` misses the case where that exact file _already_ exists on disk. It handles it just fine when the file does not exist, or when it is a conflict, e.g. a new file named `FOO` when one named `foo` already exists. But it fails when it tries to open `foo` and `foo` already exists. Whether the open behaviour will cause a problem is an issue for the `OpenFile()` that eventually gets called, and handles it correctly.

This PR fixes that issue.

### DIFF
--- a/pkg/apk/impl/fs/rwosfs.go
+++ b/pkg/apk/impl/fs/rwosfs.go
@@ -499,18 +499,22 @@ func (f *dirFS) caseSensitiveOnDisk(p string) bool {
 }
 
 // createOnDisk given a path p, determine if it should be created on disk, and, if relevant,
-// add it to the caseMap.
-func (f *dirFS) createOnDisk(p string) (createOnDisk bool) {
+// add it to the caseMap. If the file already exists on disk, also returns true.
+// This func is responsible solely for determining if you _should_ created it on disk.
+// If that would cause a conflict, that is up to the calling routing to figure out.
+func (f *dirFS) createOnDisk(p string) bool {
 	f.caseMapMutex.Lock()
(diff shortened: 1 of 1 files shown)


pkg/apk/impl/fs/rwosfs.go
```diff
@@ -499,18 +499,22 @@ func (f *dirFS) caseSensitiveOnDisk(p string) bool {
 }
 
 // createOnDisk given a path p, determine if it should be created on disk, and, if relevant,
-// add it to the caseMap.
+// add it to the caseMap. If the file already exists on disk, also returns true.
+// This func is responsible solely for determining if you _should_ created it on disk.
+// If that would cause a conflict, that is up to the calling routing to figure out.
 func (f *dirFS) createOnDisk(p string) (createOnDisk bool) {
```
jdolitsky, line 505: can we mod the signature to `func (f *dirFS) createOnDisk(p string) (bool)` if `createOnDisk`  var is never used?

jdolitsky (changes requested):

pkg/apk/impl/fs/rwosfs.go
deitch: Yeah, good point, that was an artifact. Updated.

jdolitsky (approved):

PR #495, opened 2023-02-09
use dedicated sub directory for building each arch

### LINKED ISSUE #492: Race condition with multi-arch builds
Note: this issue is on main, not in latest release.

Given the following apko yaml in  (wolfi-base):
https://raw.githubusercontent.com/chainguard-images/images/main/images/wolfi-base/configs/latest.apko.yaml
```yaml
contents:
  keyring:
  repositories:
  packages:
    - ca-certificates-bundle
    - wolfi-base
    - wolfi-baselayout

cmd: /bin/sh -l

archs:
- x86_64
- aarch64
```

Building apko:
```
make apko
```

Try publish
```
curl -sLO https://raw.githubusercontent.com/chainguard-images/images/main/images/wolfi-base/configs/latest.apko.yaml

./apko publish latest.apko.yaml ghcr.io/some-user/some-repo
```

Produces an error such as:
```
Feb  7 18:07:58.996 [INFO] [arch:aarch64] publishing image tag ghcr.io/some-user/some-repo
Error: failed to build layer image for "amd64": initializing apk: failed to initialize apk database: failed to stat directory /etc/apk: stat /etc/apk: file does not exist
2023/02/07 18:08:54 error during command execution: failed to build layer image for "amd64": initializing apk: failed to initialize apk database: failed to stat directory /etc/apk: stat /etc/apk: file does not exist
```

If we modify the apko yaml to be just one arch with
```yaml
archs:
- x86_64
```

or

```yaml
archs:
- aarch64
```

Then things appear to work fine (reaching github DENIED):
```
Feb  7 18:10:46.632 [INFO] [arch:x86_64] publishing image tag ghcr.io/some-user/some-repo
Error: publishing amd64 image: failed to build OCI image for "amd64": failed to publish: All attempts fail:
#1: GET https://ghcr.io/token?scope=repository%3Asome-user%2Fsome-repo%3Apush%2Cpull&service=ghcr.io: DENIED: denied
```

deitch:
Fixes #492 

`build` command only builds one architecture, while `publish` builds multiple. But the workdir provided for `builder.New(workdir)` only included the root of the tempdir. That meant that when building multiple archs, they would trounce each other.

This explicitly builds in an arch-specific subdirectory, which keeps everything clean.

### DIFF
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -134,18 +134,14 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
 		return err
 	}
 
-	if len(bc.Options.SBOMFormats) > 0 {
-		bc.Options.WantSBOM = true
-	} else {
-		bc.Options.WantSBOM = false
-	}
-
 	if len(archs) == 0 {
 		archs = types.AllArchs
 	}
 	if len(bc.ImageConfiguration.Archs) == 0 {
 		bc.ImageConfiguration.Archs = archs
 	}
+	// save the final set we will build
+	archs = bc.ImageConfiguration.Archs
 	bc.Logger().Infof(
 		"Publishing images for %d architectures: %+v",
 		len(bc.ImageConfiguration.Archs),
@@ -179,13 +175,23 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
 
 	mtx := sync.Mutex{}
 
-	for _, arch := range bc.ImageConfiguration.Archs {
+	for _, arch := range archs {
 		arch := arch
-		bc := *bc
+		// working directory for this architecture
+		wd := filepath.Join(workDir, arch.ToAPK())
+		bc, err := build.New(wd, opts...)
+		if err != nil {
+			return err
+		}
+
+		// we do not generate SBOMs for each arch, only possibly for final image
+		bc.Options.SBOMFormats = []string{}
+		bc.Options.WantSBOM = false
+		bc.ImageConfiguration.Archs = archs
 
 		errg.Go(func() error {
 			bc.Options.Arch = arch
-			bc.Options.WorkDir = filepath.Join(workDir, arch.ToAPK())
+			bc.Options.WorkDir = wd
 
 			if err := bc.Refresh(); err != nil {
 				return fmt.Errorf("failed to update build context for %q: %w", arch, err)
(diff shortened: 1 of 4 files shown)


jdolitsky (approved):

internal/cli/publish.go
```diff
@@ -179,13 +175,23 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
...
+		// working directory for this architecture
+		wd := filepath.Join(workDir, arch.ToAPK())
+		bc, err := build.New(wd, opts...)
+		if err != nil {
+			return err
+		}
+
+		// we do not generate SBOMs for each arch, only possibly for final image
+		bc.Options.SBOMFormats = []string{}
+		bc.Options.WantSBOM = false
```
luhring, line 189: @deitch @jdolitsky Help me understand the thinking here 😃 why wouldn't we want an SBOM for each arch? cc: @kaniini

deitch: OK, now I feel silly. I thought this comment was on #499 , not on this PR #495 . Oops. I only realized when I went to show the output with sbom-per-arch-and-index.

Thinking back, I cannot recall now why this was added. Which sounds even stranger. I cannot recall if this was something we specifically needed to do, or if it was covered elsewhere, or if this was a copy-paste logic error.

deitch: Oh, OK, I do remember. See [this line](https://github.com/chainguard-dev/apko/pull/495/files#diff-cdd538ee9a139b124886b018f25d1dbd7f4731f27c6305847563a62a37bff775L166-L171) in the original:

```go
	// This is a hack to skip the SBOM generation during
	// image build. Will be removed when global options are a thing.
	formats := bc.Options.SBOMFormats
	wantSBOM := bc.Options.WantSBOM
	bc.Options.SBOMFormats = []string{}
	bc.Options.WantSBOM = false
```

It removes the SBoM generation for each image arch, and only restores it at the end [here](https://github.com/chainguard-dev/apko/pull/495/files#diff-cdd538ee9a139b124886b018f25d1dbd7f4731f27c6305847563a62a37bff775L265-L280) (both versions of the code, old and new):

```go
	// This is a hack to skip the SBOM generation during
	// image build. Will be removed when global options are a thing.
	formats := bc.Options.SBOMFormats
	wantSBOM := bc.Options.WantSBOM
	bc.Options.SBOMFormats = []string{}
	bc.Options.WantSBOM = false
```

deitch: Basically, it defers building SBoM until after all images are done. I am not sure why, but it worked, and I saw no reason to change it.

PR #496, opened 2023-02-09
Fix for SBOMs not being generated

jdolitsky:
Regression possibly introduced by #495 cc @deitch 

```
time="2023-02-09T21:00:42Z" level=info msg="Generating arch image SBOMs"
Error: generating sbom for amd64: reading layer tar: failed to create OCI layer from tar.gz: open /tmp/apko-temp-2864749245/apko-x86_64.tar.gz: no such file or directory
2023/02/09 21:00:43 error during command execution: generating sbom for amd64: reading layer tar: failed to create OCI layer from tar.gz: open /tmp/apko-temp-2864749245/apko-x86_64.tar.gz: no such file or directory
```

### DIFF
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -158,6 +158,8 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
 	var errg errgroup.Group
 	workDir := bc.Options.WorkDir
 	imgs := map[types.Architecture]coci.SignedImage{}
+	archWorkdirs := map[types.Architecture]string{}
+	archTarballs := map[types.Architecture]string{}
 
 	// This is a hack to skip the SBOM generation during
 	// image build. Will be removed when global options are a thing.
@@ -190,6 +192,8 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
 		bc.ImageConfiguration.Archs = archs
 
 		errg.Go(func() error {
+			archWorkdirs[arch] = wd
+
 			bc.Options.Arch = arch
 			bc.Options.WorkDir = wd
 
@@ -201,6 +205,7 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
 			if err != nil {
 				return fmt.Errorf("failed to build layer image for %q: %w", arch, err)
 			}
+			archTarballs[arch] = layerTarGZ
 			// TODO(kaniini): clean up everything correctly for multitag scenario
 			// defer os.Remove(layerTarGZ)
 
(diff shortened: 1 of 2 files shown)


internal/cli/publish.go
```diff
@@ -279,8 +284,8 @@ func PublishCmd(ctx context.Context, outputRefs string, archs []types.Architectu
 		for arch, img := range imgs {
 			bc.Options.WantSBOM = true
 			bc.Options.Arch = arch
-			bc.Options.TarballPath = filepath.Join(bc.Options.TempDir(), bc.Options.TarballFileName())
-			bc.Options.WorkDir = filepath.Join(workDir, arch.ToAPK())
+			bc.Options.TarballPath = archTarballs[arch]
+			bc.Options.WorkDir = archWorkdirs[arch]
```
deitch, line 288: I don't get why these changes are necessary. The arch-specific workdir is provided by the `filepath.Join()`, so shouldn't he prior and future be identical?

pkg/build/build_implementation.go
```diff
@@ -138,7 +138,8 @@ func (di *defaultBuildImplementation) GenerateImageSBOM(o *options.Options, ic *
 		return nil
 	}
 
-	s := newSBOM(di.workdirFS, o, ic)
+	archWorkdirFS := apkfs.DirFS(o.WorkDir, apkfs.WithCreateDir(true))
+	s := newSBOM(archWorkdirFS, o, ic)
```
deitch, line 142: This part confuses me. Is it trying to read the workdir? Or the tarball? Or is it making a new one?

The build process builds the entire arch-specific image layout, then wraps it up in a tarGZ layer:

```go
	for _, arch := range archs {
		arch := arch
		// working directory for this architecture
		wd := filepath.Join(workDir, arch.ToAPK())
		bc, err := build.New(wd, opts...)
		if err != nil {
			return err
		}

		// we do not generate SBOMs for each arch, only possibly for final image
		bc.Options.SBOMFormats = []string{}
		bc.Options.WantSBOM = false
		bc.ImageConfiguration.Archs = archs

		errg.Go(func() error {
			bc.Options.Arch = arch
			bc.Options.WorkDir = wd

			if err := bc.Refresh(); err != nil {
				return fmt.Errorf("failed to update build context for %q: %w", arch, err)
			}

			layerTarGZ, err := bc.BuildLayer()
```

Trying to rebuild an `apko.FullFS` based on that working directory might, or might not, work, depending on the semantics of the underlying filesystem. On the other hand, the layerTGZ definitely has everything correct.

Can we switch this? Can we either:

* generate the SBoM from the arch-specific layerTGZ? (might need to expand into a tempdir first); OR
* generate the SBoM after each arch-specific image is built, while we still have the original `FullFS` in hand?

I think the latter would work better.

I am happy to fix this, if you can give me exactly how to reproduce it.

deitch (commented):
There are some parts I do not understand why they are necessary, but I have concerns that we may be abusing the filesystem and will not get a valid consistent SBoM.

But also equally likely that I misunderstood.

pkg/build/build_implementation.go
kaniini: Yes, lets refactor it to use the original FullFS.  I think we should do some refactoring of the SBOM analysis engine anyway.  cc @luhring who probably has opinions.

Happy to hop on a gmeet in a few hours when people are awake to flesh this out.

luhring: Yeah happy to jump on a call. I think I'm missing some context but would love to help out here

PR #498, opened 2023-02-11
support chmod/chown

### LINKED ISSUE #497: Error building image with s6 (nginx)
Some permissions related issue for an nginx build:

```
curl -LO https://raw.githubusercontent.com/chainguard-images/images/main/images/nginx/configs/latest.apko.yaml

./apko build latest.apko.yaml nginx output.tar

docker load < output.tar 

docker run --rm nginx:latest
```

Error:
```
s6-svscan: fatal: unable to mkdir .s6-svscan: Permission denied
```

And running as root, different type of error:
```
docker run --rm -u root nginx:latest
```

```
s6-supervise nginx: warning: unable to spawn ./run - waiting 10 seconds
s6-supervise nginx (child): fatal: unable to exec run: Permission denied
```

deitch:
Fixes #497

There were several issues:

* chown and chmod were non-functional (it was one of those thing set for "later" that never got addressed)
* Uname and Gname were not being provided for the tar headers
* Uname/Gname/UID/GID override were set, which caused _everything_ to be owned by root, and thus nonrunnable

### DIFF
--- a/pkg/apk/impl/fs/memfs.go
+++ b/pkg/apk/impl/fs/memfs.go
@@ -15,6 +15,7 @@
 package fs
 
 import (
+	"archive/tar"
 	"bytes"
 	"encoding/binary"
 	"errors"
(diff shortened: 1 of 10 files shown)


deitch:
As an aside, this does show up some of the weaknesses in the memfs we are using. I more urgently want to fix the chown/chmod issues, but at a certain point, we should bite the bullet and just replace it.

kaniini (approved):

Issue #450, opened 2022-12-22
UNSUPPORTED: Invalid parameter at 'ImageManifest' failed to satisfy constraint: 'Invalid JSON syntax' when push to ECR


tuananh:
I got this invalid parameter at 'ImageManifest' when pushing to ECR

```
Error: attaching sboms to amd64 image: writing sbom: All attempts fail:
#1: PUT https://xxx.dkr.ecr.ap-southeast-1.amazonaws.com/v2/platform-images/demo-image/manifests/sha256-186a7f24e5a81f8b56c0c94cc8b343683a93780ca852ad318cf9d2c0385702c1.sbom: UNSUPPORTED: Invalid parameter at 'ImageManifest' failed to satisfy constraint: 'Invalid JSON syntax'
```

eddiezane:
Bumping to cosign v2 should resolve this per @imjasonh 

https://github.com/chainguard-dev/apko/blob/dfc8b9a0732595aebb95e79b95e305932eb9da91/go.mod#L21

PR #502, opened 2023-02-13
use cosign v2 and correct SPDX mediaType

imjasonh:
Fixes https://github.com/chainguard-dev/apko/issues/450

The main change is that `ctypes.SPDXJSONMediaType` used [here](https://github.com/chainguard-dev/apko/blob/v0.6.0/pkg/build/oci/oci.go#L251) is `text/spdx+json` in v2, and just `spdx+json` in v1, and some registries barf when encountered with `spdx+json`.

Similer `ko` bug fixed the same way: https://github.com/ko-build/ko/issues/878

Signed-off-by: Jason Hall <jason@chainguard.dev>


eddiezane (approved):
Ty!

PR #508, opened 2023-02-16
handle multiple packages providing same files sanely

deitch:
This is an issue for dependencies that are not names but in the "provides" list. Previously, it just took the last one, which could be different each time.

Here, it improves, if not completes it:

1. If the same package we are using provides the dependency, then nothing to add
2. If another package that provides the dependency has the same origin as our package, use that
3. Otherwise, use the first one on the list

The last step still isn't great, but it is an improvement as we add the first 2 steps. 

@kaniini pointed out that the apk-tools logic is:

> if a package is already enqueued to be installed, and it provides something, then that package will be considered first for the provided dependency

We should do that, but in a later PR.

### DIFF
--- a/pkg/apk/impl/repo.go
+++ b/pkg/apk/impl/repo.go
@@ -205,14 +205,14 @@ func (a *APKImplementation) getRepositoryIndexes(ignoreSignatures bool) ([]*repo
 type PkgResolver struct {
 	indexes     []*repository.RepositoryWithIndex
 	nameMap     map[string]map[string]*repository.RepositoryPackage
-	providesMap map[string]string
+	providesMap map[string][]*repository.RepositoryPackage
 }
 
 // NewPkgResolver creates a new pkgResolver from a list of indexes.
 func NewPkgResolver(indexes []*repository.RepositoryWithIndex) *PkgResolver {
 	var (
 		pkgNameMap     = map[string]map[string]*repository.RepositoryPackage{}
-		pkgProvidesMap = map[string]string{}
+		pkgProvidesMap = map[string][]*repository.RepositoryPackage{}
 	)
 	p := &PkgResolver{
 		indexes: indexes,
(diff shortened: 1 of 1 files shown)


jdolitsky (approved):

jdolitsky (changes requested):
waiting on ci

pkg/apk/impl/repo.go
```diff
@@ -285,30 +287,34 @@ func (p *PkgResolver) GetPackageWithDependencies(pkgName string) (pkg *repositor
...
+		// get the one that most matches what was requested
+		versions := make([]string, 0, 10)
+		for version := range pkgsWithVersions {
+			versions = append(versions, version)
+		}
+		targetVersion := getBestVersion(versions, version, compare)
+		if targetVersion == "" {
+			return nil, nil, nil, fmt.Errorf("could not find package %s in indexes: %w", pkgName, err)
+		}
+		pkg = pkgsWithVersions[targetVersion]
```
jdolitsky, line 301: nil deref here?

pkg/apk/impl/repo.go
```diff
@@ -378,33 +384,45 @@ func (p *PkgResolver) getPackageDependencies(pkg *repository.RepositoryPackage,
...
+			// get the one that most matches what was requested
+			var versions []string
+			for version := range depPkgWithVersions {
+				versions = append(versions, version)
+			}
+			targetVersion := getBestVersion(versions, version, compare)
+			if targetVersion == "" {
+				return nil, nil, fmt.Errorf("could not find package %s in indexes: %w", dep, err)
+			}
+			depPkg = depPkgWithVersions[targetVersion]
```
jdolitsky, line 398: nil dereference here?

deitch:
Ah, now it is much happier

jdolitsky (approved):

PR #509, opened 2023-02-16
use highest version of package when found by provided files

deitch:
Before, when a dependency was found not by package name, but by a provided file, we just took whichever version was first. That was inconsistent, and not the right answer anyways.

This consistently does the following:

1. If the package declaring the dependency also provides the dependency, do nothing, as we satisfy it ourselves.
2. If any of the packages providing the dependency are of the same origin `o:` as the package requesting the dependency, restrict our selection set to just those packages with a matching origin; else look in all packages providing the dependency
3. Of the packages remaining that provide the dependency (matching origin or, if none, then all), take the highest version (highest as defined by the package logic)

### DIFF
--- a/pkg/apk/impl/repo.go
+++ b/pkg/apk/impl/repo.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 
 	"gitlab.alpinelinux.org/alpine/go/repository"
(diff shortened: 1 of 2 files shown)


deitch:
So far, I have tested the results of provided versions solely against apko-amd64. Am running checks against many others and will post here when done.

deitch:
Ready to run.

kaniini (approved):

PR #537, opened 2023-02-24
Windows

imjasonh:
Following up from https://github.com/chainguard-dev/apko/pull/536, this is a brief attempt to hide unix-isms behind Go build flags so that we can build apko for Windows.

There's one more, here, that I can't easily snip out:
https://github.com/chainguard-dev/apko/blob/89912f3e264800286f641cf86eeee367db1723ea/pkg/vfs/vfs.go#L34

If we think we care about Windows I think we're within throwing distance of it. If it's not worth it we can close this PR and carry this unspeakable horror to our graves.


jdolitsky (commented):
🤯

xnox:
i think WSL is a better option. Also needs a rebase. Closing for now.

PR #551, opened 2023-03-07
apk: version: filterPackages: fall back to comparing provided versions when the package version itself is mismatched

kaniini:
cc @vaikas


vaikas (approved):
Woot Woot!

PR #586, opened 2023-03-21
When build and publish, carry buildcontext to sbom generation

### LINKED ISSUE #584: SBOMs are missing data about files
This appears to be a regression. Before:

```console
$ cosign download sbom --platform linux/arm64 cgr.dev/chainguard/maven:latest-20230131 | jq 'keys | contains(["files"])'
...
true
```

Now:

```console
$ cosign download sbom --platform linux/arm64 cgr.dev/chainguard/maven | jq 'keys | contains(["files"])'
...
false
```

The suspected cause of this behavior is that `apko` is no longer incorporating SBOM data that Melange puts into individual apks at `/var/lib/db/sbom/...`. This would mean that additional data is missing from SBOMs now, beyond just file data.

cc: @rawlingsj

deitch:
Fixes #584 

We were creating a new `BuildContext` for the SBoM generation, rather than reusing the one from the build stage. That meant a new `fsys`, which lost all of the information about the existing one.

This carries it through, enabling us to get all of the information.

This could use a cleaner review, as I am _fairly_ confident I understood the issue in #584 and that this fixes it, but not 100% sure.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -31,3 +31,11 @@ jobs:
         run: |
           make snapshot
           ./dist/apko-build_linux_amd64_v1/apko version
+
+      # Required in SBOM quality test below
+      - name: Setup cosign
+        uses: sigstore/cosign-installer@v3
+
+      - name: Check SBOM quality
+        run: |
+          ./hack/sbom-quality.sh
--- a/hack/sbom-quality.sh
+++ b/hack/sbom-quality.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -ex
+
+if [[ ! -f apko ]]; then
+    echo "Please first run \"make apko\". Exiting."
+    exit 1
+fi
+
+SBOM_FILENAME="sbom.json"
+APKO_CONFIG="sbom-quality-test.yaml"
+REGISTRY_BASE_IMAGE="index.docker.io/library/registry:2.8.1"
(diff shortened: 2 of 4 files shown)


rawlingsj:
I know we're fixing a bug here and am asking the wider group, not putting this on you @deitch :) ... 

I wonder how we can add an automated test for this to help protect the feature in the future?

Is that unit tests?  Maybe more of an integration test that verifies using the steps described in the [issue](https://github.com/chainguard-dev/apko/issues/584#issue-1632940340)?  Something that can run in CI and help us in the future.

rawlingsj:
Yeah agreed - sounds like it's worth it though

jdolitsky:
This change appears to resolve the issue

luhring:
> Yeah agreed - sounds like it's worth it though

I agree, @rawlingsj. I talked with @deitch and @kaniini, and we're hoping that a solid integration test can get prioritized in the very near future. 🤞  This will let us begin to assert aspects of SBOM generation, and can be extended to assert more things about the build process over time.

Ariadne will talk with @patflynn about exact prioritization.

jdolitsky:
> I wonder how we can add an automated test for this to help protect the feature in the future?

I've just added an "SBOM quality" CI test to this PR to ensure that there will not be regressions. We can also add more SBOM assertions to that script later

jdolitsky (approved):

PR #587, opened 2023-03-21
Add SBOM quality CI test

jdolitsky:
Add a test to prevent regression related to #584

Note: this should fail if #586 is not merged, but pass after


jdolitsky:
looks like test fails as expected:
```
SBOM does not have files. Exiting.
```

adding this in #586 and closing

PR #590, opened 2023-03-21
Add CI test harness

jdolitsky:


hack/ci/00-build.sh
```diff
@@ -0,0 +1,51 @@
...
+
+set -ex
+
+# Go to repo root
+cd "$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )/../../"
+
+if [[ ! -f apko ]]; then
+    echo "Please first run \"make apko\". Exiting."
+    exit 1
+fi
```
deitch, line 14: Can we remove lines 9-14, as it already exists in the wrapper, and move 20 into `ci-tests.sh` as a wrapper? And then make `apko` available that way? So it looks something like:

```sh
set -ex

 # Go to repo root
 cd "$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )/../../"

 if [[ ! -f apko ]]; then
     echo "Please first run \"make apko\". Exiting."
     exit 1
 fi

export APKO=$(pwd)/apko
for exe in `find hack/ci -name '*.sh' | sort`; do
    echo "Running test: ${exe}"
    tmpdir="tmp/${exe}"
    mkdir -p "${tmpdir}"
    (cd $tmpdir && sh -ex ${exe})
    rm -rf "${tmpdir}"
 done
```

And then each file need not check just call:

```sh
$APKO build ...
```

deitch (dismissed):
Other than minor structure changes to make it DRYer, let's get this in. We always can iterate.

hack/ci/00-build.sh
jdolitsky: 👍

luhring (approved):

PR #601, opened 2023-03-30
Add StopSignal support.

amouat:
Adds stop-signal to YAML which sets the OCI Config paramater StopSignal. This parameter controls the signal that is sent to the container to shutdown.

This is motivated by the nginx image, which should be sent the SIGQUIT signal, not SIGTERM (the default).


amouat:
Reviewer, please check:
 - is there somewhere I can easily insert a test?
 - have I missed anywhere I should have plumbed this change in? (I nearly missed the debug output for example)

imjasonh (approved):

amouat:
Ok, I'm going to merge and add indirect testing via the nginx image.

PR #606, opened 2023-04-04
ensure truncate when creating new file in case one already was there

deitch:
Fixes the issue discovered by @amouat (and memorialized by @kaniini with the great term "acropalypse bug").


amouat (approved):

amouat:
Could we add a test to prevent regressions?

deitch:
Test added.

amouat (approved):

amouat:
Nice, glad you managed to do it in the same test!

PR #616, opened 2023-04-07
native in-memory filesystem

deitch:
The memfs we were using has a lot of limitations that we had to work around: no support for symlinks or hardlinks, no support for mknod (devices), etc. We put many layers on top to get it to work, and finally hit the limit with https://github.com/wolfi-dev/os/actions/runs/4618853760/jobs/8171294258?pr=1074

This eliminates the external memfs and replaces it with a simple memfs that has full support.


deitch:
Lots of CI to fix. We will get to it.

deitch:
And green CI!

jdolitsky:
Lets get this in prior to merging: https://github.com/chainguard-dev/apko/pull/590

deitch:
Just need someone to review and approve.

pkg/apk/impl/fs/memfs.go
```diff
@@ -17,77 +17,156 @@ package fs
...
 }
 
-type ownership struct {
-	uid, gid int
-	perms    fs.FileMode
-}
-type memFS struct {
-	*origMemfs.MemFS
-	perms map[string]*ownership // used only for overrides
+func NewMemFS() *memFS {
```
luhring, line 36: nit: It might be nice to return the interface type here, if you don't want to export the `memFS` type. It would make it more clear in this context what interface `memFS` is intending to implement, and it also makes it easier to traverse expected behaviors in docs/IDEs for folks calling `NewMemFS()`.

pkg/apk/impl/fs/memfs_test.go
luhring, line 1: Just a general note in case it's helpful (no change needed necessarily) — the `require.<assertion>` functions stop test execution, whereas the corresponding `assert.<assertion>` variations don't. `require...` is good for when following statements in the test depend on an outcome from the thing being required. But I'm a huge fan of using the `assert` variants in all other cases, because they let someone running tests see all of the problems at once.

pkg/apk/impl/fs/memfs.go
deitch: Yeah, that is a good idea. I went back and forth on it and just defaulted, but I think you are correct.

pkg/apk/impl/fs/memfs_test.go
deitch: Fair enough. I think it is easier to default to `assert`, because then I don't need to start figuring out which depend and which do not. Since you said, "no changed needed", I will leave it for now.

pkg/apk/impl/fs/memfs.go
deitch: Done

deitch:
I enabled auto-merge; as soon as someone approves, it will merge in.

jdolitsky:
relies on #619

deitch:
#619 is in

deitch:
Aha! Now we got it!

Issue #627, opened 2023-04-19
Problems with python finding itself


luhring:
Opening for @deitch  (feel free to retitle and/or correct the below!)

When debugging failing E2E tests in Melange (e.g. https://github.com/chainguard-dev/melange/actions/runs/4743961439/jobs/8424331419?pr=395) we noticed two issues:

1. the apk implementation doesn't have support for using tilde (`~`) to specify a version for a dependency
2. the apk implementation doesn't have support a package depending directly on itself (i.e. not on a provides, but the package itself) — e.g. `python3` depending on `python3~3.11`

luhring:
@kaniini Any ideas about why Melange's E2E tests weren't failing before now? Could it be that the outside world change (e.g. Alpine's config for `python3`?)

jdolitsky:
FWIW I also have problems finding myself

kaniini:
I suspect that python 3.11.3 having that dependency is just a side effect of https://github.com/alpinelinux/abuild/commit/6806a0d39b25c30ca7368b29a9ff78c57c0deec9

deitch:
The first part (tilde support) is in #630, but I am a little concerned about the second part.

apko already has support for "I depend on `foo`, my package provides `foo`, I satisfy my own requirement." This would expand it to include, "I depend on `foo=3.11`, I _am_ `foo=3.11`, I satisfy my own requirement.

We can do that, and add tests, without too much pain.

But then what happens with the wolfi build dependencies? The issue discussed [here](https://github.com/wolfi-dev/os/pull/1480) means that it expects, "if I depend on my own name, do _not_ satisfy it myself, rather look for a different (possibly lower) version to find that." In the wolfi case, that means potentially stage3.

We do not (yet) use the same code for resolving the build graph in wolfictl and apko, but we are working to get there, and the logic should be the same.

Can a package A itself resolve a dependency on package A (what this suggests), or can it not (what the wolfi-dev/os issue suggest)?

kaniini:
> Can a package A itself resolve a dependency on package A (what this suggests)

Yes.  It has to be this way.

The `-stage0` stuff in Wolfi is a policy decision that does not imply any behavioral requirement of the solver.

deitch:
So the normal apk resolution graph says, "A-2.0 depends on A can be satisfied by itself." We took a decision that says, "no, it cannot".

That can be satisfied with options. Cool, thanks.

deitch:
@luhring see #631

PR #630, opened 2023-04-20
add support for tilde matcher

deitch:
If a package requirement is `foo~1.2`, we do not handle it correctly. This now adds support for it. Also adds a few unit tests.

Stage 1 of fixing #627 ; follow-on coming later.

### DIFF
--- a/pkg/apk/impl/version.go
+++ b/pkg/apk/impl/version.go
@@ -305,7 +305,7 @@ func includesVersion(actual, required packageVersion) bool {
 		return false
 	}
 
-	// was there post-suffi
+	// was there post-suffix
 	if required.postSuffix != packageVersionPostModifierNone && actual.postSuffix != required.postSuffix {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


pkg/apk/impl/version.go
```diff
@@ -274,6 +274,52 @@ func compareVersions(actual, required packageVersion) versionCompare {
...
+	if required.preSuffix != packageVersionPreModifierNone && actual.preSuffix != required.preSuffix {
+		return false
+	}
+
+	// was there pre-suffix number
+	if required.preSuffixNumber != 0 && actual.preSuffixNumber != required.preSuffixNumber {
+		return false
+	}
+
+	// was there post-suffi
```
imjasonh, line 308:
```suggestion
	// was there post-suffix
```

imjasonh (approved):

deitch:
Thanks for the fix @imjasonh ; did you not sign the DCO?

imjasonh:
Ugh. I thought I could get away with commiting a suggestion to save you time, but uhhh I guess not. Thanks DCO!

kaniini (approved):

PR #631, opened 2023-04-21
add optional support for self-package resolution

### LINKED ISSUE #627: Problems with python finding itself

deitch:
Fixes #627 

Since apk normally does allow self-package resolution (if I fulfill one of my dependencies, then look no further), but other places sometimes do not (e.g. wolfictl, as a policy choice, see [this comment](https://github.com/chainguard-dev/apko/issues/627#issuecomment-1517042175) by @kaniini), we make it optional in the resolver.

We always make it true here, but given the overlap between apko logic and wolfictl graph, and the plans for merging them going forward, making this optional only can help.

### DIFF
--- a/examples/alpine-python3.yaml
+++ b/examples/alpine-python3.yaml
@@ -1,0 +1,12 @@
+contents:
+  repositories:
+    - https://dl-cdn.alpinelinux.org/alpine/edge/main
+  packages:
+    - alpine-base
+    - python3
+
+cmd: /bin/sh -l
+
+# optional environment configuration
+environment:
+  PATH: /usr/sbin:/sbin:/usr/bin:/bin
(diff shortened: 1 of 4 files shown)


luhring (changes requested):
Holding this one while we walk through (and maybe debunk 😄 ) testing failures I'm seeing

deitch:
Debunk? Sounds entertaining!

The CI is clean; what is failing?

kaniini (approved):

PR #633, opened 2023-04-23
handle versioning in provides

### LINKED ISSUE #632: Solver does not consider provided package version in provides entries
Consider the following repository:

- `a-1-r0`
- `b-2-r0`, provides `a=0.9`

If a user requests package `a`, `b` will be selected instead of `a` as it provides `a` (at version `0.9`) but has a higher package version (`2`).  Instead, `a` should be selected as it provides `a` at version `1`.

deitch:
Fixes #632 

As described in #632, the solver found all packages that either were named with the target (e.g. `a`) and then sorted them. The sorting algorithm, when it came to version order, used the _package_ version. But if the package is there just because it provides something, it should use the _provides_ version.

This fixes it.

### DIFF
--- a/pkg/apk/impl/repo.go
+++ b/pkg/apk/impl/repo.go
@@ -345,14 +345,14 @@ func (p *PkgResolver) ResolvePackage(pkgName string) (pkgs []*repository.Reposit
 		if len(packages) == 0 {
 			return nil, fmt.Errorf("could not find package %s in indexes: %w", pkgName, err)
 		}
-		sortPackages(packages, nil, nil, pin)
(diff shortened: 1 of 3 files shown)


deitch:
Found the CI build error, working it.

kaniini:
I'm confused.  Why is `wolfi-baselayout` queued for installation twice?

deitch:
It doesn't. That's just the entire list. It dedups. 

It still could be clearer though. We can get a fix in for that after this one is in.

imjasonh (approved):

PR #635, opened 2023-04-23
Support onbuild instruction

mesaglio:
Hi!

With this [onbuild instruction](https://docs.docker.com/engine/reference/builder/#onbuild) support we can add behavior to the apko images when we use them as base images.

### DIFF
--- a/pkg/build/types/build_option.go
+++ b/pkg/build/types/build_option.go
@@ -41,6 +41,7 @@ type BuildOption struct {
 	Environment map[string]string `yaml:"environment,omitempty"`
 
 	Entrypoint ImageEntrypoint `yaml:"entrypoint,omitempty"`
+	OnBuild    []string        `yaml:"on_build,omitempty"`
 }
 
 // Apply applies a patch described by a BuildOption to an apko environment.
@@ -68,6 +69,8 @@ func (bo BuildOption) Apply(ic *ImageConfiguration) error {
 	for k, v := range bo.Environment {
 		ic.Environment[k] = v
 	}
+
+	ic.OnBuild = append(ic.OnBuild, bo.OnBuild...)
 
 	if bo.Entrypoint.Type != "" {
 		ic.Entrypoint = bo.Entrypoint
--- a/pkg/build/types/build_option.go
+++ b/pkg/build/types/build_option.go
@@ -41,7 +41,7 @@ type BuildOption struct {
 	Environment map[string]string `yaml:"environment,omitempty"`
 
 	Entrypoint ImageEntrypoint `yaml:"entrypoint,omitempty"`
-	OnBuild    []string        `yaml:"on_build,omitempty"`
+	OnBuild    []string        `yaml:"on-build,omitempty"`
 }
 
 // Apply applies a patch described by a BuildOption to an apko environment.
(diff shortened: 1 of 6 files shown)


mesaglio:
Hi! @imjasonh @kaniini, can you help me with this pr?

pkg/build/types/build_option.go
```diff
@@ -41,6 +41,7 @@ type BuildOption struct {
 	Environment map[string]string `yaml:"environment,omitempty"`
 
 	Entrypoint ImageEntrypoint `yaml:"entrypoint,omitempty"`
+	OnBuild    []string        `yaml:"on_build,omitempty"`
```
kaniini, line 44:
```suggestion
	OnBuild    []string        `yaml:"on-build,omitempty"`
```

pkg/build/types/image_configuration.go
```diff
@@ -205,4 +205,11 @@ func (ic *ImageConfiguration) Summarize(logger log.Logger) {
 			logger.Printf("      %s: %s", k, v)
 		}
 	}
+
+	if len(ic.OnBuild) > 0 {
+		logger.Printf("  on_build:")
```
kaniini, line 210:
```suggestion
		logger.Printf("  on-build:")
```

pkg/build/types/image_configuration.go
```diff
@@ -205,4 +205,11 @@ func (ic *ImageConfiguration) Summarize(logger log.Logger) {
...
+		for _, onBuildInstruccion := range ic.OnBuild {
```
kaniini, line 211:
```suggestion
		for _, onBuildInstruction := range ic.OnBuild {
```

pkg/build/types/image_configuration.go
```diff
@@ -205,4 +205,11 @@ func (ic *ImageConfiguration) Summarize(logger log.Logger) {
...
+			logger.Printf("    - %s", onBuildInstruccion)
```
kaniini, line 212:
```suggestion
			logger.Printf("    - %s", onBuildInstruction)
```

kaniini (changes requested):
In general, I don't think we should encourage doing this, but it seems fine.

imjasonh:
I'm a little curious to learn more about your use case. Do you have an example scenario in mind? I thought `ONBUILD` instructions in Dockerfiles were deprecated and recommended against for quite some time, but maybe I'm mistaken.

mesaglio:
Hello @imjasonh, I did not find in the official docker documentation that the onbuild instructions had been deprecated but I will look for it.

One use case we are looking at is having a distroless image for production and another in the development stage. Seeking that the environment is not so different, we think that the development environment is the production environment + busybox + compilation.
I thought some dockerfile with this style
```dockerfile
FROM prod_img:latest
# this can be added with melango + apko
RUN apk add busybox
ONBUILD COPY . /app
WORKDIR /app
ONBUILD CMD ["python3", "-m", "app"]
```

Final Dockerfile that devs can use
```dockerfile
FROM dev_img:latest
```

mesaglio:
This have no sense 😢

mfortin:
@mesaglio did you find an alternative ?
I was looking for this feature too

mesaglio:
No I dont. But I understand that `ONBUILD` instruction is in [Docker image spec](https://github.com/moby/docker-image-spec/blob/v1.2.0/v1.2.md) no in [OCI image spec](https://specs.opencontainers.org/image-spec/config/?v=v1.0.1)

PR #642, opened 2023-04-25
Optimize ggcr interactions

jonjohnsonjr:
This is a first pass at speeding up the registry portion of apko publish.

```
hyperfine --warmup 1 --runs 3 "apko publish examples/nginx.yaml $APKO_REPO" "./apko publish examples/nginx.yaml $APKO_REPO"
Benchmark 1: apko publish examples/nginx.yaml $APKO_REPO
  Time (mean ± σ):     22.535 s ±  0.478 s    [User: 5.212 s, System: 3.160 s]
  Range (min … max):   22.097 s … 23.046 s    3 runs

Benchmark 2: ./apko publish examples/nginx.yaml $APKO_REPO
  Time (mean ± σ):      8.323 s ±  0.200 s    [User: 4.510 s, System: 2.379 s]
  Range (min … max):    8.193 s …  8.554 s    3 runs

Summary
  './apko publish examples/nginx.yaml $APKO_REPO' ran
    2.71 ± 0.09 times faster than 'apko publish examples/nginx.yaml $APKO_REPO'
```


pkg/build/oci/oci.go
```diff
@@ -213,9 +233,26 @@ func buildImageFromLayerWithMediaType(mediaType ggcrtypes.MediaType, layerTarGZ
...
+	}
+	desc, err := remote.Get(srcRef, remoteOpts...)
+	if err != nil {
+		return fmt.Errorf("fetching %s: %w", src, err)
+	}
+	pusher, err := remote.NewPusher(remoteOpts...)
+	if err != nil {
+		return err
+	}
+	if err := pusher.Push(context.TODO(), dstRef, desc); err != nil {
```
imjasonh, line 252: I think it's fine to change the signature of `Copy`, if that's why you're going with `context.TODO()` here. It's pre-1.0, and I don't expect anybody should be using this in apko anyway, vs using something like `crane.Copy` directly.

pkg/build/oci/oci.go
```diff
@@ -68,6 +68,26 @@ var keychain = authn.NewMultiKeychain(
...
+	remoteOpts = []remote.Option{remote.WithAuthFromKeychain(keychain)}
+
+	pusher, err := remote.NewPusher(remoteOpts...)
+	if err == nil {
+		remoteOpts = append(remoteOpts, remote.Reuse(pusher))
+	} else {
+		log.DefaultLogger().Infof("NewPusher(): %v", err)
+	}
+
+	puller, err := remote.NewPuller(remoteOpts...)
```
imjasonh, line 83: I'm probably dumb, but why isn't a `pusher` also a `puller`, so we can create just one and `Reuse` it for both pushes and pulls?

imjasonh (approved):
Nice!

Do you think that most of the speed-up came from errgroups, or reusing pushers?

pkg/build/oci/oci.go
jonjohnsonjr: Because of how many (including ours) registries implemented auth incorrectly, but also because they're structs.

pkg/build/oci/oci.go
jonjohnsonjr: Plumbed it around.

jonjohnsonjr:
> Do you think that most of the speed-up came from errgroups, or reusing pushers?

Mostly groups:

```
hyperfine --warmup 1 --runs 3 "apko publish examples/nginx.yaml $APKO_REPO" "./apko publish examples/nginx.yaml $APKO_REPO"
Benchmark 1: apko publish examples/nginx.yaml $APKO_REPO
  Time (mean ± σ):     22.448 s ±  1.025 s    [User: 5.148 s, System: 3.399 s]
  Range (min … max):   21.673 s … 23.611 s    3 runs

Benchmark 2: ./apko publish examples/nginx.yaml $APKO_REPO
  Time (mean ± σ):     16.222 s ±  0.627 s    [User: 4.459 s, System: 2.679 s]
  Range (min … max):   15.724 s … 16.926 s    3 runs

Summary
  './apko publish examples/nginx.yaml $APKO_REPO ran
    1.38 ± 0.08 times faster than 'apko publish examples/nginx.yaml $APKO_REPO
```

https://github.com/jonjohnsonjr/apko/commit/79d5a211ba1c473482996d57fbee7ebb9015e81e

PR #643, opened 2023-04-27
allow overwrite of existing file if origin matches

deitch:
Ensures that file overwrites happen only when the overwriting package has the same origin as the one that originally installed the file. 

Partially addresses https://github.com/chainguard-dev/melange/issues/401


jdolitsky (commented):
looks like real failures?
```
❕ x86_64    | installing alpine-keys (2.4-r1)
❕ x86_64    | buildImage failed: installing apk packages: unable to install files for pkg alpine-keys: unable to install file over existing one: etc/apk/keys/alpine-devel@lists.alpinelinux.org-6165ee59.rsa.pub
❕ x86_64    | image configuration:
```

deitch:
They are. This opened a whole can of worms. They needed to be dealt with, just not today. But I guess yes today 😁 . Almost got them, been a long haul.

PR #648, opened 2023-04-28
change type structs encoding tag to json

joshrwolf:
changes the struct types encoding to `json`. this does _not_ change apko's config to be json based, but uses json as the common denominator of the two to appease external parsers that live in `json` land (like the json schema format used by [schemastore](https://www.schemastore.org/json/)).

This works assuming apko doesn't use any `yaml` specific syntax, which I'm guessing it doesn't, but please correct me if I'm wrong!

The encoding is fairly simple, but I did my best to add/remove `omitempty` where necessary.

This PR is related to https://github.com/wolfi-dev/wolfictl/pull/190

### DIFF
--- a/pkg/build/types/build_option.go
+++ b/pkg/build/types/build_option.go
@@ -17,30 +17,30 @@ package types
 // ListOption describes an optional deviation to a list, for example, a
 // list of packages.
 type ListOption struct {
-	Add    []string `yaml:"add,omitempty"`
-	Remove []string `yaml:"remove,omitempty"`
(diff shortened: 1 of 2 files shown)


mepi262:
@joshrwolf 
It seems that this pull request has conflicts.
Would you resolve it?

luhring:
Can this be closed now, since https://github.com/chainguard-dev/apko/pull/933 merged?

joshrwolf:
close it!

PR #655, opened 2023-05-02
document the fields in types/

joshrwolf:
first stab at documenting the fields for apko's public config surface. I best efforted the required/optional fields, knowing that these aren't codegen'd to actually mean anything, but I still think they are useful for an end user.

this will be used by external parsers (like [this](https://github.com/wolfi-dev/wolfictl/pull/193) one) to generate various external documentation that should hopefully improve the devx onboarding/overall experience.

### DIFF
--- a/pkg/build/types/types.go
+++ b/pkg/build/types/types.go
@@ -23,45 +23,72 @@ import (
 )
 
 type User struct {
+	// Required: The name of the user
 	UserName string
-	UID      uint32
-	GID      uint32
+	// Required: The user ID
+	UID uint32
+	// Required: The user's group ID
+	GID uint32
 }
 
 type Group struct {
+	// Required: The name of the group
 	GroupName string
-	GID       uint32
-	Members   []string
+	// Required: The group ID
+	GID uint32
+	// Required: The list of members of the group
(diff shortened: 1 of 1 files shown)


pkg/build/types/types.go
```diff
@@ -23,72 +23,125 @@ import (
...
-	WorkDir     string            `yaml:"work-dir,omitempty"`
-	Accounts    ImageAccounts     `yaml:"accounts,omitempty"`
-	Archs       []Architecture    `yaml:"archs,omitempty"`
+	// Required: The apk packages in the container image
+	Contents ImageContents `yaml:"contents,omitempty"`
+	// Required: The entrypoint of the container image
+	//
+	// This typically is the path to the executable to run. Since many of
+	// Chainguard Images do not include a shell, this should be the full path
+	// to the executable.
```
jdolitsky, line 114: I'd omit the chainguard piece to separate the OSS from product
```suggestion
	// This typically is the path to the executable to run. Since many of
	// images do not include a shell, this should be the full path
	// to the executable.
```

jdolitsky (approved):

PR #668, opened 2023-05-11
Add test of determinism

mattmoor:


deitch (changes requested):
Once #670 is in, this one should be rebased and should pass.

Once requested change. There is no need to publish; we can use `apko build config.yaml imagename output.tar` and compare the contents of `output.tar`

mattmoor:
To add to my comment above (since I forgot), we also had zero coverage of multi-arch builds because even our "all architectures" build fans out to do per-arch builds.

imjasonh (approved):

mattmoor:
I'm going to bias towards getting this protection in, so we don't regress.  @deitch happy to discuss "to registry or not to registry" more, and we can always change things, but I'm going to optimistically bias towards the additional coverage for now.

Thanks so much for fixing this 🙏

PR #670, opened 2023-05-11
latest go-apk with consistent file ordering

deitch:
@mattmoor had found that the file ordering from `go-apk` was not consistent, which led to differing sha256 sums of the layer.tgz. This fixes that.

In my tests, every time I built any image, I got the exact same contents in the tar, right down to sha256sum of each.

Note that the output tar itself is _not_ consistent, despite the consistent contents. This could be for various reasons, and should be resolved separately.


deitch:
Once this is in, we should rebase https://github.com/chainguard-dev/apko/pull/668 to it and see that it passes, then merge that in.

kaniini (approved):

deitch:
This has a race condition. Fix is in https://github.com/chainguard-dev/go-apk/pull/30 . Wait for that to be merged in, I will update this, then this should be clear.

deitch:
Updated now that https://github.com/chainguard-dev/go-apk/pull/30 is in with that commit.

Also made sure that the test that was getting stuck, sails through. There might be more, but that is why we have CI!

mattmoor:
awesome, I'll rebase my other pr 🙏

PR #675, opened 2023-05-12
Feature: Compute the default timestamp from installed APKs

mattmoor:
:gift: When `SOURCE_DATE_EPOCH` is not specified, this change infers the timestamp to assign the image from the build date of the oldest installed APK.

/kind feature

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -236,7 +236,7 @@ func BuildCmd(ctx context.Context, imageRef, outputTarGZ string, archs []types.A
 			// This computation will only affect the timestamp of the image
 			// itself and its SBOMs, since the timestamps on files come from the
 			// APKs.
-			if err := bc.SetBuildDateEpoch(); err != nil {
+			if bc.Options.SourceDateEpoch, err = bc.GetBuildDateEpoch(); err != nil {
 				return fmt.Errorf("failed to determine build date epoch: %w", err)
 			}
 			if bc.Options.SourceDateEpoch.After(multiArchBDE) {
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -19,9 +19,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"time"
 
-	apkimpl "github.com/chainguard-dev/go-apk/pkg/apk"
 	"github.com/google/go-containerregistry/pkg/name"
 	coci "github.com/sigstore/cosign/v2/pkg/oci"
 	"github.com/sirupsen/logrus"
@@ -195,7 +193,11 @@ func BuildCmd(ctx context.Context, imageRef, outputTarGZ string, archs []types.A
 		}
 	}()
 
-	ade := bc.Options.SourceDateEpoch
+	// We compute the "build date epoch" of the multi-arch image to be the
+	// maximum "build date epoch" of the per-arch images.  If the user has
+	// explicitly set SOURCE_DATE_EPOCH, that will always trump this
+	// computation.
+	multiArchBDE := bc.Options.SourceDateEpoch
 
 	for _, arch := range archs {
 		arch := arch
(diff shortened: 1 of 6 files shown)


mattmoor:
I think I like the term "build date epoch" for this.

internal/cli/build.go
```diff
@@ -223,6 +227,17 @@ func BuildCmd(ctx context.Context, imageRef, outputTarGZ string, archs []types.A
 				return fmt.Errorf("failed to build layer image: %w", err)
 			}
 
+			if _, ok := os.LookupEnv("SOURCE_DATE_EPOCH"); !ok {
```
deitch, line 230: I think that this chunk of code should be in its own reusable func, as you reproduce it below in `cli/publish.go` as well.

Although I am wondering if this shouldn't just be an option we pass (likely in `pkg/build/options.go`):

```go
func WithInstallFilesTimestamp(ts string) ...
```

Where that says, "calculate the latest date and use it for anything that you add in". 

What we have here is duplicated in `cli/publish.go`, and then goes for the same variable in `pkg/build/build.go` [here](https://github.com/chainguard-dev/apko/blob/a89818861f525e38332c6caac33a79003598236d/pkg/build/build.go#L187) (which really is not good; we should be looking up vars here in `internal/cli` and not in `pkg/build`).

deitch (commented):
I think we should have the following, based on your logic:

* for files from a `.apk`: _always_ use the original timestamp from the file
* for files inserted by apko itself (just ldconfig, busybox links and char devices for now, I think), two options:
   * fixed override when `SOURCE_DATE_EPOCH` is set
   * calculated override when `SOURCE_DATE_EPOCH` is blank

So the logic in this PR is solid. Just needs to be in the right location. My suggestion (can be convinced otherwise):

1. Move all logic that checks `os.LookupEnv()` into the `RunE` functions of cobra
2. Have that call `WithInstalledFilesTimestamp()` (or similar) that you create
3. Calculate that inside [buildImage](https://github.com/chainguard-dev/apko/blob/a89818861f525e38332c6caac33a79003598236d/pkg/build/build_implementation.go#L256) rather than here in `cli/`, where it gets messy

The only issue I see with that flow, is that we don't actually use the timestamp overrides when laying down the filesystem, where we cannot really control it. We use it when generating the tarball, which happens [here](https://github.com/chainguard-dev/apko/blob/a89818861f525e38332c6caac33a79003598236d/pkg/build/build_implementation.go#L121-L123).

So a little bit of plumbing to do, but not too much. Maybe the right thing to do is have `InstallBusyboxLinks()` and `InstallLdConfig()` etc. return their list of files, and then change 

[tarball.WithSourceDateEpoch()](https://github.com/chainguard-dev/apko/blob/a89818861f525e38332c6caac33a79003598236d/pkg/tarball/tarball.go#L50-L55)

take a list of `tar.Header`, like the other options, so it can override just for this.

internal/cli/build.go
mattmoor: I think we cleared up a bunch of confusion in slack, but this code could definitely be clearer and cleaner still.

I'll refactor it a bit, and add a bunch of comments because my [ab]use of `SourceDateEpoch` to tunnel the configuration timestamp through is subtle.

mattmoor:
So I shuffled the logic around a bit, the `apkDateEpoch` I had is now a method on `build.Context` called `SetBuildDateEpoch` (which will make it easier to reuse in `tf-apko`!).  I think we clarified the subset of timestamps I'm interested in influencing, but there may be more (such as the symlinks), so we can continue to shuffle things, if that makes sense.

> Move all logic that checks os.LookupEnv() into the RunE functions of cobra

This would be my preference as well, and ideally the `SourceDateEpoch` field would signal this in some way, but the static constructor `New` for `build.Context` is reading this directly and not leaving much in the way of breadcrumbs to determine the original user intent.  I think ideally this would use `*time.Time` where `nil` -> unspecified, but I think that's a bigger refactor that'd (ideally) clean all of this behavior up.


Hopefully the new structure and the comments help to clarify some of the earlier confusion, and these changes should definitely make it easier to pull this into `tf-apko`, so LMK if you see anything else we can improve. 🙏

deitch:
If I get this correctly, you:

1. Take a starting BDE
2. As you build each arch-specific image, find the highest buildtime for any package in that image
3. If the highest one is higher than our most recent, save that
4. Use the now-absolute-highest (i.e. most recent) of any package in any arch as the new BDE
5. Use that new BDE on the image-specific artifacts (SBoMs, etc.)

Is that it?

deitch:
If so, then the only purpose of [Context.SetBuildDateEpoch()](https://github.com/chainguard-dev/apko/pull/675/files#diff-e3115b8ca9fdf611ae3dc61dc235e39aeeddbca61036b803ba3c0826f25cc3c9R91) is to change the var, so it can be read later? That modification of something not for its own purpose, but for some external reason, strikes me as a bit risky. Someone might see it, use it before build, and be surprised when it doesn't do what is expected.

Assuming I understood it correctly, we could change it to this:

```go
func (bc *Context) GetLatestEpoch() (epoch time.Time, err error) {
	pl, err := bc.impl.InstalledPackages(bc.fs, &bc.Options)
	if err != nil {
		return fmt.Errorf("failed to determine installed packages: %w", err)
	}
	for _, p := range pl {
		if p.BuildTime.After(epoch) {
			epoch = p.BuildTime
		}
	}
	return epoch, nil
}
```

Then use it as:

```go
			bde, err := bc,GetLatestEpoch()
                        if err != nil {
				return fmt.Errorf("failed to determine build date epoch: %w", err)
			}
			if bde.After(multiArchBDE) {
				multiArchBDE = bde
			}
```

mattmoor:
> If I get this correctly

Almost.  We use the per-image BDE for each image's config/SBOM, and compute the `MAX(BDE per image)` for the index's SBOM.

> Then use it as:

We can certainly refactor things similar to this (a draft on the way here did exactly this!), but we need to set the `bc.Options.SourceDateEpoch` for each image, so that when the config/SBOM are generated and they carry the correct timestamp.

As I said about the accumulation of the max is just for the index's SBOM.

mattmoor:
Ok I removed the side-effect and adjusted things so the caller owns that and it is just a `Get` call now.

mattmoor:
I'd love to have a better test case for this, perhaps something that leverages fully pinned packages and expects a fixed timestamp as a result?  Open to ideas here 🤔

deitch:
Looks better. 

> I'd love to have a better test case for this, perhaps something that leverages fully pinned packages and expects a fixed timestamp as a result? Open to ideas here 

No great ideas. Maybe building a few packages, saving them locally, run a build and then ensure that the timestamp is respected?

deitch (approved):

deitch:
LGTM. I am holding on clicking "merge" in case there is more you area ready to add.

PR #681, opened 2023-05-16
adds warning in case of unknown distro in etc/os-release

### LINKED ISSUE #447: Warn users when producing an image with an unknown distro
Problem: Today, when apko builds an image, if none of the packages it adds to the image include an `/etc/os-release` file, the image will end up with an `/etc/os-release` file that gets created on-the-fly, which looks like this:

```text
ID=unknown
NAME="apko-generated image"
PRETTY_NAME="apko-generated image"
VERSION_ID=unknown
HOME_URL="https://github.com/chainguard-dev/apko"
```

This can be problematic. In particular, if someone runs a vulnerability scanner on the image, the scanner will not be able to correctly identify the distro, and thus the scanner won't be able to locate the appropriate advisory/fix data source to use during the scan — which means the scan results will be invalid.

In most cases, apko users probably don't _intend_ to have an unidentified distro like this. It's more likely that they just didn't realize how to add a proper distro `/etc/os-release` file, which today is done by including a package in the config file that contains an `/etc/os-release` file.

Therefore, apko should check to see if any package has added an `/etc/os-release` file and ensure the user sees a warning about the fact that the distro won't be identifiable and how that can lead to problems.

mritunjaysharma394:
Tries to fix #447 

Currently, with my little knowledge of the code being a new contributor here, I could find that mutation of `/etc/os-release` files happens while validation and `ID` tells distro name usually in most `/etc/os-release` files. 

Therefore while checking for `ID`, if it is `""`, we set it to `unknown` in the generated file, this happens in the [Validate function](https://github.com/chainguard-dev/apko/blob/main/pkg/build/types/image_configuration.go#L111). Therefore I added a warning for this there in the check for empty ID. This should probably support what we desired. Thanks!

### DIFF
--- a/pkg/build/os-release.go
+++ b/pkg/build/os-release.go
@@ -79,6 +79,9 @@ func (bc *Context) GenerateOSRelease() error {
 	defer w.Close()
 
 	if osr.ID != "" {
+		if osr.ID == "unknown" {
+			bc.Logger().Warnf("distro ID not specified and /etc/os-release does not already exist")
+		}
 		_, err := fmt.Fprintf(w, "ID=%s\n", osr.ID)
 		if err != nil {
 			return err


kaniini (changes requested):
The validation needs to happen at a later state, when `/etc/os-release` is actually generated, as it may be present due to a package providing it.

mritunjaysharma394:
Thanks for the review @kaniini, based on my understanding, I attempted it another way, looking for your feedback to improve it further if it needs!

pkg/build/os-release.go
```diff
@@ -82,6 +82,9 @@ func (di *defaultBuildImplementation) GenerateOSRelease(
 	defer w.Close()
 
 	if ic.OSRelease.ID != "" {
+		if ic.OSRelease.ID == "unknown" {
+			o.Logger().Warnf("unidentifiable distro, may cause failure in vulnerability scanning results to correctly identify distro")
```
kaniini, line 86:
```suggestion
			o.Logger().Warnf("distro ID not specified and /etc/os-release does not already exist")
```

mritunjaysharma394: Thanks, improved it :))

mritunjaysharma394: Let me know if it needs any other changes @kaniini, thanks a lot for the review

luhring (approved):

mritunjaysharma394:
Thanks a lot for the review @luhring, not sure how the build and lint happened as last I checked it worked but maybe I am new to apko code base and its guidelines and erred somewhere, I will take a look at this quickly and push the required fixes, thanks again for the review @luhring :pray:

mritunjaysharma394:
I don't know why I am on a spree of typos while committing these  days, the same happened with https://github.com/wolfi-dev/os/pull/2308 

Missed a `}` :sweat_smile:

luhring:
LGTM. @kaniini please review

mritunjaysharma394:
Hi @kaniini, just fixed the merge conflicts due to recent changes, feel free to review and let me know if further improvements required, thanks for the help!

kaniini (approved):

PR #690, opened 2023-05-22
Fix annotations

amouat:
This PR adds the `annotation` flag to `apko build` for equivalence with `publish`.

It also fixes a bug where the annotations from the yaml config were deleted.

This PR could do with some tests - I'd appreciate any advice on where/how to add these.

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -48,6 +48,7 @@ func buildCmd() *cobra.Command {
 	var extraRepos []string
 	var buildOptions []string
 	var logPolicy []string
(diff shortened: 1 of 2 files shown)


imjasonh (commented):
Looks good, can we get a test as well?

cpanato (approved):
lgtm

amouat:
> Looks good, can we get a test as well?

Yeah, I wanted to do this, but wasn't quite sure where to put it! I was hoping there was a similar test I could crib from.

PR #696, opened 2023-05-23
Stop using tarball.LayerFromFile

jonjohnsonjr:
We spend a ton of CPU compressing a targz, then immediately hand it over to tarball.LayerFromFile to gunzip it in order to compute its sha256.

Instead, compute the diffid and digest when we generate the targz so we can avoid recomputing it and go a bit faster.

Supersedes https://github.com/chainguard-dev/apko/pull/693

Bumps go-apk to https://github.com/chainguard-dev/go-apk/commit/27a2b06830e3c6e2384a911aa5c001a4297a4ef1


deitch (approved):
This looks good. Going through it, I see some cleanup we can do to simplify some parts around it, but not this itself, and it is not worth holding up.

I like how you simplified the multiple calls to set the media type. Much simpler

PR #700, opened 2023-05-24
Remove indirection for apk implementation

jonjohnsonjr:
This is just making it more difficult to follow the control flow, and we are mostly just testing the control flow, which is the compiler's job.


deitch (approved):
We might be able to go one step further and remove `type APK` from here entirely. With your change, it is a very lightweight wrapper around `apkimpl.APK`, a `FullFS` instance, and some options.

But one step at a time. This helps.

PR #711, opened 2023-05-31
extend summarize to provide the rest of the options

deitch:
We regularly use `BuildContext.Summarize(logger)` to report the build options, but it only reported 5 of them, missing the rest.

This adds the rest, which should make understand builds (and debugging) easier.


luhring (approved):

deitch:
Updated to print options as json as suggested by @jonjohnsonjr

Issue #716, opened 2023-06-01
Add smoke tests to multi-platform image generation


jonjohnsonjr:
I had to revert some seemingly harmless changes in https://github.com/chainguard-dev/apko/pull/714

I _believe_ I introduced a bug that caused the layer platform for every image to be GOARCH, so to prevent that regression (and feel safe in refactoring), I'd like to add some tests that check the contents of each platform's filesystem to ensure they have the expected platform-specific files.

PR #719, opened 2023-06-01
Check that images have the correct layers

jonjohnsonjr:
Fixes https://github.com/chainguard-dev/apko/issues/716


jonjohnsonjr:
Testing against the known-bad commits:

```
$ crane export --platform linux/amd64 "${REF}" | tar -Ox etc/apk/arch | grep x86_64
$ echo $?
1
```

Issue #717, opened 2023-06-01
Get test coverage of internal/cli


jonjohnsonjr:
We added https://github.com/chainguard-dev/apko/pull/715 for race detection, which would have helped with https://github.com/chainguard-dev/apko/pull/714, but I believe there are a handful of race conditions in internal/cli that it also would have caught if we had go test coverage in those packages.

We don't need extensive tests, just some amount of coverage to let the race detector do its thing.

PR #720, opened 2023-06-01
Add a no-op test for PublishCmd

jonjohnsonjr:
Adding this test as-is so that the race-detection fixing stuff can be reviewed separately from the actual test code.

I am unsure how this test passes.

Fixes https://github.com/chainguard-dev/apko/issues/717


imjasonh:
> I am unsure how this test passes.

This is not very encouraging.

Is the test there just to exercise things that invoke the logger, to ensure race detector will tell us if logging is racey?

It looks like `PublishCmd` doesn't do anything because it ranges over `archs` and there are none. Does it make sense to have this test with 2+ `archs`?

jonjohnsonjr:
> Does it make sense to have this test with 2+ archs?

I just wanted this PR with the race fixes to be small and easy to review, this follows up with a real test: https://github.com/chainguard-dev/apko/pull/721

imjasonh (approved):

PR #724, opened 2023-06-02
unify publish and build commands

deitch:
This changes the implementation of `apko build` and `apko publish` CLI commands (primarily the CLI parts) to unify the common areas. There was a _ton_ of duplication beforehand, which this mostly eliminates.

There still is duplication in the cobra part, but that can wait for a later PR.

In doing so, it restructures the build process in a few ways. Note that the underlying build part - laying down a filesystem from apk files, and turning those into layer tars - is completely unchanged. What is changed

* Before, we had a workdir for building images (subdirectories per arch) and someplace else that sboms and other such would go. This makes everything under the workdir, with the artifacts in `<workdir>/image/`. This makes it much easier to reason about where things are, find artifacts, and clean up
* Before, the build would do its thing, and publish its thing, with lots of overlap but lots different. With this PR, the build phase is entirely separate, in its own func. After it is done, the `build` command combines the various artifacts into `outputTarGZ`, while publish publishes those artifacts. This is much easier to reason about, as well as add capabilities.
* With a single "build image layers and sboms and manifests and index" function, we now can pass the results of those around, rather than constantly calculating where the various artifacts (like sboms) should be and what they should be named.
* Some of the `oci.*` utilities have been simplified. For example, whereas `publishImage` used to take a layer and then publish it, duplicating a lot of what other things do, now they have more of a single focused job.
* Separating these allowed me to separate build options from publish options. That also makes it easier to understand what is going on.

I have tested several build cases, and the CI does a really good job testing various scenarios. I also checked sbom creation etc.

There is a publish test, courtesy of @jonjohnsonjr , which it does pass. However, I would like quite a bit more publish testing, especially around the sboms, as I am weaker in that area.

Notes of some things I did _not_ address, on purpose:

* publish of sbom does not let you attach multiple formats. I didn't change that. We could (should?) but it is beyond the scope of this PR; this contains enough
* there still is a lot of that "indirection so we can generate fakes" that @jonjohnsonjr started to clean up. This does _not_ remove any of that in any serious sense. It just is out of scope.


pkg/build/options.go
jonjohnsonjr, line 1: Genuine question, not rhetorical:

A bunch of functional options go away in this file, does it also make sense for those fields to get dropped in `Options`, or are those fields set/read by other mechanisms (e.g. config file)? I would like to get rid of vestigial struct fields if we can, but I'm not sure if it's safe.

deitch: I got rid of some of them already, removing anything I could to PublishOptions. I didn't hunt for any more.

pkg/build/oci/oci.go
```diff
@@ -56,10 +56,10 @@ const (
 )
 
 func BuildImageFromLayer(layer v1.Layer, ic types.ImageConfiguration, logger log.Logger, opts options.Options) (oci.SignedImage, error) {
```
imjasonh, line 58: oci.go is hundreds of lines of critically important code, with no tests, and this is a pretty thorough refactor of this code.

I don't want to bog this change down by making you add tests for all of this code, but if there's anything you're adding in this PR that you think warrants a test, I'd really love to start adding them. 🙏

deitch: @imjasonh I appreciate not wanting to bog this down, yet I agree that we really need them, as oct.go _is_ hundreds of lines of critically important code.

In that vein, I started by restructuring the oci _package_ into 6 separate files, each with many fewer functions. That should make it much easier to write tests. I did it as a separate commit, to make it easier to view separately. It doesn't change any functionality, just moves the single overly large file into dedicated ones for building images, building indexes, sbom, publishing types and constants.

deitch: And I cleaned up some duplication, and I added inline documentation to pretty much every func in `package oci`, and created the 4 `*_test.go` files, and added the empty skeletons, so every primary func is defined.

We need a lot of those tests, but I am not convinced we can get that done in here. This was a big step forward., though

internal/cli/options.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func WithPackageVersionTagPrefix(packageVersionTagPrefix string) PublishOption {
+	return func(p *publishOpt) error {
+		p.packageVersionTagPrefix = packageVersionTagPrefix
+		return nil
+	}
+}
+
+// withTagSuffix sets a tag suffix to use, e.g. `-glibc`.
```
jonjohnsonjr, line 40: A bunch of these have lowercase comments but uppercase names.

deitch: Fixed

pkg/build/types/types.go
```diff
@@ -345,3 +345,10 @@ func ParseArchitectures(in []string) []Architecture {
 	})
 	return archs
 }
+
+type Sbom struct {
```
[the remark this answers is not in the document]
deitch: Fixed

pkg/build/build_implementation.go
```diff
@@ -347,6 +360,27 @@ func buildImage(
 	return nil
 }
 
+// WriteIndex saves the index file from the given image configuration.
+func (di *defaultBuildImplementation) WriteIndex(o *options.Options, idx coci.SignedImageIndex) (string, int64, error) {
```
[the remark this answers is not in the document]
deitch: I think it is ok. The only consideration I would have is if we are publishing to docker daemon. If we use `apko publish`, it already has its own code paths for that. So the only consideration is for `apko build` followed by `docker load`, which tested correctly when I ran it. And I believe our CI does as well

jonjohnsonjr (approved):

Issue #692, opened 2023-05-22
Copy Annotations into config.json (Docker Labels)


amouat:
Apko currently supports adding annotations as per the OCI Image spec: https://github.com/opencontainers/image-spec/blob/main/annotations.md

Docker's support for annotations seems a bit mixed, and in particular `docker inspect` will not display annotations. For this reason, it might be worth adding annotations into the [config](https://github.com/opencontainers/image-spec/blob/main/config.md), which will get surfaced in `docker inspect` and easily found/searched/scripted etc.

mritunjaysharma394:
Oh that makes sense, thanks a bunch @amouat,  I think so that more or less I have understood the assignment here: update Labels which are a map so annotations can be hopefully easily added here but I suppose we already have [Annotations in Manifest struct](https://github.com/google/go-containerregistry/blob/main/pkg/v1/manifest.go#L30) so that should be handleable too? 

If that sounds okay, maybe I can try to get my hands dirty and create a PR/Draft PR to implement this? Thanks again :pray:

amouat:
That would be great @mritunjaysharma394! However, I would like @kaniini or someone to verify that we do want this functionality first. 

Assuming we do, it's hopefully just a case of copying the annotations struct into the labels struct for the config.

Issue #481, opened 2023-01-27
Annotations in the apko yaml file aren't picked up.


found-it:
It seems that annotations inside an apko yaml file aren't picked up when the image is built. My manifest is 

```yaml
contents:
  keyring:
    - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
  repositories:
    - https://packages.wolfi.dev/os
  packages:
    - wolfi-baselayout
    - busybox

accounts:
  groups:
    - groupname: inky
      gid: 1001
  users:
    - username: inky
      uid: 1001
      gid: 1001
  run-as: inky

environment:
  VERSION: v0.0.1

entrypoint:
  command: sh

archs:
  - x86_64

annotations:
  foo: bar
```

When I run `apko build` it seems to get picked up (by looking at the log). However when I look at the image manifest, the annotations don't exist. I tried loading the image and pushing it to a registry then using `crane manifest` to check the manifest and still no annotations.

<details>
  <summary>apko build log</summary>

```
Jan 27 22:29:48.807 [INFO] loading config file: /work/configs/latest.apko.yaml
Jan 27 22:29:48.808 [DEBUG] [arch:x86_64] failed to probe VCS URL: opening git repository: directory /work/configs is not in a git repository
Jan 27 22:29:48.808 [INFO] [arch:x86_64] WARNING: ignoring archs in config, only building for current arch (amd64)
Jan 27 22:29:48.808 [INFO] [arch:x86_64] building image 'myreg/inky:build'
Jan 27 22:29:48.808 [INFO] [arch:x86_64] build context:
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   working directory: /tmp/apko-799989857
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   tarball path:
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   use proot: false
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   source date: 1970-01-01 00:00:00 +0000 UTC
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   Docker mediatypes: false
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   SBOM output path: /work
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   arch: x86_64
Jan 27 22:29:48.808 [INFO] [arch:x86_64] image configuration:
Jan 27 22:29:48.808 [INFO] [arch:x86_64]   contents:
[...]

PR #727, opened 2023-06-04
add annotations to index manifest

imjasonh:
This adds any annotations specified in the config file, or determined from VCS info if enabled, to the index manifest in addition to constituent image manifests.

This has always worked, and still does:
```
crane manifest --platform=linux/amd64 $(go run ./ publish examples/nginx.yaml ttl.sh/jason) | jq .annotations
{
  "org.opencontainers.image.revision": "d16c9050918e6a5da5bf575d625be7f615b3b966",
  "org.opencontainers.image.source": "git+ssh://github.com/chainguard-dev/apko.git"
}
```

This didn't work, but does now:
```
crane manifest $(go run ./ publish examples/nginx.yaml ttl.sh/jason) | jq .annotations
{
  "org.opencontainers.image.revision": "d16c9050918e6a5da5bf575d625be7f615b3b966",
  "org.opencontainers.image.source": "git+ssh://github.com/chainguard-dev/apko.git"
}
```

(previously the annotations were empty and `jq` would return `null`)

I'm not proud of all I had to do to get this wired up.

#481 
Fixes https://github.com/chainguard-dev/apko/issues/692

### DIFF
--- a/pkg/build/oci/index.go
+++ b/pkg/build/oci/index.go
@@ -54,8 +54,27 @@ func GenerateDockerIndex(ctx context.Context, ic types.ImageConfiguration, imgs 
 
 // generateIndexWithMediaType generates an index or docker manifest list from the given imgs. The index type
 // is provided by the `mediaType` parameter.
-func generateIndexWithMediaType(mediaType ggcrtypes.MediaType, _ types.ImageConfiguration, imgs map[types.Architecture]oci.SignedImage) (name.Digest, oci.SignedImageIndex, error) {
-	idx := signed.ImageIndex(mutate.IndexMediaType(empty.Index, mediaType))
+func generateIndexWithMediaType(mediaType ggcrtypes.MediaType, ic types.ImageConfiguration, imgs map[types.Architecture]oci.SignedImage) (name.Digest, oci.SignedImageIndex, error) {
+	// If annotations are set and we're using the OCI mediaType, set annotations on the index.
+	annotations := map[string]string{}
+	if mediaType == ggcrtypes.OCIImageIndex {
+		annotations = ic.Annotations
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		if ic.VCSUrl != "" {
+			if url, hash, ok := strings.Cut(ic.VCSUrl, "@"); ok {
+				annotations["org.opencontainers.image.source"] = url
+				annotations["org.opencontainers.image.revision"] = hash
+			}
+		}
+	}
+
+	idx := signed.ImageIndex(
+		mutate.IndexMediaType(
+			mutate.Annotations(empty.Index, annotations).(v1.ImageIndex),
+			mediaType),
+	)
 	archs := make([]types.Architecture, 0, len(imgs))
 	for arch := range imgs {
 		archs = append(archs, arch)
--- a/.github/workflows/build-samples.yml
+++ b/.github/workflows/build-samples.yml
@@ -142,3 +142,30 @@ jobs:
               exit 1
             fi
           done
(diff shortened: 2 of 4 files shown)


imjasonh:
While I'm here, and since I noticed, I also knocked out https://github.com/chainguard-dev/apko/issues/692

If you push at all, I'll add tests for this stuff, or drop the index annotations stuff as too gross.

kaniini:
I wonder if we should focus on landing #724 first.  It seems like it might introduce some conflicts with this PR.

deitch:
I suspect @kaniini is correct, it might introduce conflicts with #724. I would like to try and get that one in, if we can.

imjasonh:
> I wonder if we should focus on landing #724 first. It seems like it might introduce some conflicts with this PR.

I'm fine with that. I'll come back to this after #724.

pkg/build/oci/oci.go
```diff
@@ -460,10 +461,23 @@ func publishIndexWithMediaType(ctx context.Context, mediaType ggcrtypes.MediaTyp
...
+	// oci.SignedImageIndex, so we do a lot of work to avoid having to reimplement
+	// mutate.Annotations in ocimutate.
+	annotations := ic.Annotations
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	if ic.VCSUrl != "" {
+		if url, hash, ok := strings.Cut(ic.VCSUrl, "@"); ok {
+			annotations["org.opencontainers.image.source"] = url
+			annotations["org.opencontainers.image.revision"] = hash
```
jonjohnsonjr, line 475: Didn't we want to drop this due to the churn it caused?

deitch:
@imjasonh #724 is in, which includes that restructuring of `pkg/build/oci/`, so go for this

amouat:
Oh, just saw this! Brings mucho joy to me and I'm currently writing a blog on annotations/labels for the fun of it.

deitch (approved):
LGTM but needs conflicts fixed.

pkg/build/oci/index.go
```diff
@@ -54,8 +54,27 @@ func GenerateDockerIndex(ctx context.Context, ic types.ImageConfiguration, imgs
...
+	annotations := map[string]string{}
+	if mediaType == ggcrtypes.OCIImageIndex {
+		annotations = ic.Annotations
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		if ic.VCSUrl != "" {
+			if url, hash, ok := strings.Cut(ic.VCSUrl, "@"); ok {
+				annotations["org.opencontainers.image.source"] = url
+				annotations["org.opencontainers.image.revision"] = hash
```
imjasonh, line 68: Resurfacing Jon's question about how this makes the image churn.

amouat: I don't follow this - if the hash changes hasn't the image changed?

imjasonh: If you change the README in the repo you'll get a new source commit, without changing anything about the image. If you have a repo with dozens of apko YAMLs you don't want to have to rebuild all of them when you change any of them, etc.

I think we're fine with this as-is because `VCSUrl` can be disabled, and is disabled in chainguard-images/images, but I think it's worth considering what we want the default to be for regular apko users.

I'm also fine punting on the question to land this PR.

jonjohnsonjr: > I think we're fine with this as-is because VCSUrl can be disabled, and is disabled in chainguard-images/images

This addresses my concern.

jonjohnsonjr (approved):

PR #731, opened 2023-06-06
Split publishing and loading

jonjohnsonjr:
This untangles daemon from remote when publishing a bit more, which will make it easier to move towards lazily producing these images, which will make it easier to avoid building things entirely.

### DIFF
--- a/pkg/build/oci/publish.go
+++ b/pkg/build/oci/publish.go
@@ -19,10 +19,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"sync"
-	"time"
 
-	"github.com/avast/retry-go"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/daemon"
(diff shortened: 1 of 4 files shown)


deitch:
I like the idea of decoupling. The question is, should we do it even one step further? Maybe it shouldn't be `apko publish` and `apko publish --local`, but rather `apko publish` (for OCI registries) and `apko load` or similar (for docker)? I think that might make it even clearer

imjasonh (approved):

imjasonh (approved):

pkg/build/oci/publish.go
```diff
@@ -33,257 +30,236 @@ import (
...
 	}
-	dig, err := PublishImage(ctx, v1Image, local, shouldPushTags, logger, tags, remoteOpts...)
-	return dig, v1Image, err
+	logger.Debugf("docker daemon response: %s", strings.ReplaceAll(resp, "\n", "\\n"))
+	for _, tag := range tags {
+		localDstTag, err := name.NewTag(tag)
+		if err != nil {
+			return name.Digest{}, err
+		}
+		if strings.HasPrefix(localSrcTag.Name(), fmt.Sprintf("%s/", LocalDomain)) {
```
markkrj, line 114: Will not this always be true? Line 97 has `localSrcTagStr := fmt.Sprintf("%s/%s:%s", LocalDomain...`
I was expecting that `publish` actually loaded the tags to my Docker daemon. This prevents it.

PR #733, opened 2023-06-07
safe rename

deitch:
When moving the sbom files around to `--sbom-path` as part of the `apko build` command, we use `os.Rename()`. That was somewhat shortsighted, as that only works when it can move inodes (or their equivalents) around, i.e. on the same filesystem partition. If it is across partitions, filesystems, networks, bind-mounts, etc., it fails, rather than falling back to copy-delete (which is what `mv` does).

This changes it to try `os.Rename()` first, and if it fails _because of cross-partition_, it falls back to copy/delete.

It is difficult to make a test for this specific function, because you need to be able to set up two distinct partitions and have writes across them, something that is challenging to do inside CI. I am open to ideas.

In the meantime, I did create a small test program and run it manually, feeding it both local and cross-partition.

```go
package main

import (
	"bytes"
	"errors"
	"fmt"
	"io"
	"log"
	"os"

	"golang.org/x/sys/unix"
)

// rename just like os.Rename, but does a copy and delete if the rename fails
func rename(from, to string) error {
	err := os.Rename(from, to)
	if err == nil {
		log.Printf("%s -> %s: os.Rename worked", from, to)
		return nil
	}
	// we can handle cross-device rename errors
	if !errors.Is(err, unix.EXDEV) {
		return err
	}
	f1, err := os.Open(from)
	if err != nil {
		return err
	}
	defer f1.Close()
	f2, err := os.Create(to)
	if err != nil {
		return err
	}
	defer f2.Close()
	_, err = io.Copy(f2, f1)
	if err != nil {
		return err
	}
	log.Printf("%s -> %s: copy/delete worked", from, to)
	return os.Remove(from)
}

func doCompare(from, to string) error {
	content := []byte("abc")
	if err := os.WriteFile(from, content, 0o644); err != nil {
		return fmt.Errorf("could not create %s: %v", from, err)
	}
	if err := rename(from, to); err != nil {
		return fmt.Errorf("could not do rename: %v", err)
	}
	b, err := os.ReadFile(to)
	if err != nil {
		return fmt.Errorf("could not read target file: %v", err)
	}
	if !bytes.Equal(b, content) {
		return fmt.Errorf("target file %s had mismatched content, %s vs %s", to, b, content)
	}
	return os.Remove(to)
}

func main() {
	from := os.Args[1]
	toLocal := os.Args[2]
	toRemote := os.Args[3]
	if err := doCompare(from, toLocal); err != nil {
		log.Fatalf("error local: %v", err)
	}
	if err := doCompare(from, toRemote); err != nil {
		log.Fatalf("error remote: %v", err)
	}
	log.Printf("success!\n")
}
```

I ran it for local between `/tmp` and `/tmp`, and for remote between `/tmp` and `/run`, which is mounted separately. Output was:

```sh
# go run . /tmp/from /tmp/to /run/foo
2023/06/07 08:24:01 /tmp/from -> /tmp/to: os.Rename worked
2023/06/07 08:24:01 /tmp/from -> /run/foo: copy/delete worked
2023/06/07 08:24:01 success!
```

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -16,7 +16,9 @@ package cli
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"sync"
@@ -25,6 +27,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sys/unix"
 
 	"chainguard.dev/apko/pkg/build"
 	"chainguard.dev/apko/pkg/build/oci"
@@ -159,7 +162,8 @@ func BuildCmd(ctx context.Context, imageRef, outputTarGZ string, archs []types.A
 
 	// copy sboms over to the sbomPath target directory
 	for _, sbom := range sboms {
-		if err := os.Rename(sbom.Path, filepath.Join(sbomPath, filepath.Base(sbom.Path))); err != nil {
+		// because os.Rename fails across partitions, we do our own
+		if err := rename(sbom.Path, filepath.Join(sbomPath, filepath.Base(sbom.Path))); err != nil {
 			return fmt.Errorf("moving sbom: %w", err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


deitch:
> Is there a way to be more confident in large refactors like this?

It is a fair question. We already are _much_ more confident, due to a lot of the integration-type testing that we have added, e.g. we build all of the examples, and we have more complex builds and publishes that @jdolitsky and @imjasonh and @jonjohnsonjr have added, and we continue to add more.

There always will be some thing that slip through. This case was a good example. We would have had to think to have a test case where `--sbom-path` is on a different filesystem than the `/tmp` where we build. I think the chances of even thinking to test that are extremely small.

> It's nice that our presubmits to images catch this now with a digest bump. 

I kind of view that as a success. We cannot foresee every possible variant at upstream tool build time, so before we adopt an update, run it through a pre-merge test (in this case, presubmits) and catch errors. Errors in `main` = bad; errors anywhere else = good catch.

> I'm imagining a way for you to build a version of the apko image with the changes in a PR and then run that apko build through the presubmit testing in the images repo, before a merge

I think you are suggesting we expand the integration-type testing (can we just call it "integration"?) that we are doing. We have the "build all of the examples",. which is good, and now we have the stuff in `hack/ci/` originally from @jdolitsky . You are looking at expanding that part. But does it mean apko PR CI dips into downstream images? Or that we expand the test cases in apko repo?

dlorenc:
I'm just suggesting something manual for riskier changes - where you can kickoff a test build of apko then run it through presubmits in the chainguard-images repo, that seems to be the most meaningful validation of the larger changes to this repo. We're still averaging 1-2 breakages a week, which is better than before but still too high.

kaniini:
Adding the ability to run the presubmit tests would be easy enough, but I am waiting to see how the build system refactor plays out first.

PR #735, opened 2023-06-07
Remove build implementation interface

jonjohnsonjr:
This just swaps the interface for the actual implementation that's currently being used. The tests do nothing, so they are stubs for now, but I will add actual tests once we have the ability to do real dependency injection into the builder.


imjasonh (approved):
I'd love a 👍 from @deitch or @kaniini but this looks overall like a good simplifying change.

deitch (commented):
I like this approach, cleaning up more of the indirection.

The only issue is that we lose the tests. `TestBuildImage()` and `TestBuildLayer()` in `build_test.go` now are empty. Were the tests of value? Do we replace them with anything?

jonjohnsonjr:
The unit tests only test that a method fails if the methods it is supposed to call fail. This doesn't really test anything other than "yep this code was executed", which is kinda useful for dynamic languages, but is mostly just a no-op. We have real coverage from internal/cli tests and the presubmit.

deitch:
In that case, go for it. We are no worse off, and arguably better without the unnecessary indirection.

PR #751, opened 2023-06-15
go-apk with support for pinned pre-existing as deps

deitch:
Pulls in `go-apk` with support for pinned repos as dependencies when already installed.


deitch:
That is mighty strange. It is reporting a test failure with a race condition. I cannot see that locally. Exact same command.

Going to rerun job and hope it is transient.

deitch:
Yup, it was just transient.

kaniini (approved):

PR #755, opened 2023-06-21
add optional oci volumes field to resulting image config

joshrwolf:
this is to support images that leverage a container runtime that uses the OCI spec `volumes` field ([ref](https://github.com/opencontainers/image-spec/blob/main/config.md#properties)) to set up runtime volume mounts. it is a noop for all other runtimes

this supports the relatively small subset of cases where an image is built to run with `docker`, and needs certain mount configs to function. more concretely, the `k3s` container running in docker (ie: `k3d`). this runs `containerd` in docker by default with the `overlayfs` snapshotter, and requires a non `overlayfs` mount point to work. this is achieved upstream with a host volume mount ([ref](https://github.com/k3s-io/k3s/blob/master/package/Dockerfile#LL16C1-L17C1)).

this change lets us mimic this behavior with something like:

```yaml
...
volumes:
  - /var/lib/rancher/k3s
```

this only affects runtimes that actually parse the `Volumes` field (like docker).

related: https://github.com/wolfi-dev/os/pull/2970 https://github.com/chainguard-images/images/pull/877

### DIFF
--- a/pkg/build/oci/image.go
+++ b/pkg/build/oci/image.go
@@ -132,6 +132,13 @@ func BuildImageFromLayer(layer v1.Layer, ic types.ImageConfiguration, created ti
 		cfg.Config.WorkingDir = ic.WorkDir
 	}
 
+	if ic.Volumes != nil {
+		cfg.Config.Volumes = make(map[string]struct{})
+		for _, v := range ic.Volumes {
+			cfg.Config.Volumes[v] = struct{}{}
+		}
+	}
+
 	// Set these environment variables if they are not already set.
 	if ic.Environment == nil {
 		ic.Environment = map[string]string{}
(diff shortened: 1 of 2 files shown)


imjasonh (approved):

imjasonh (commented):
Can you add something to https://github.com/chainguard-dev/apko/blob/main/docs/apko_file.md ?

joshrwolf:
> Can you add something to https://github.com/chainguard-dev/apko/blob/main/docs/apko_file.md ?

this sounds like a cop out, but I almost prefer not to, the use case here is so few and far between I'm worried it'll confuse end users between `path`/`volume`.

I inline commented the struct, so in the future when we have a json schema for autocomplete it'll show up, is that good enough?

imjasonh:
> > Can you add something to https://github.com/chainguard-dev/apko/blob/main/docs/apko_file.md ?
> 
> this sounds like a cop out, but I almost prefer not to, the use case here is so few and far between I'm worried it'll confuse end users between `path`/`volume`.
> 
> I inline commented the struct, so in the future when we have a json schema for autocomplete it'll show up, is that good enough?

Yeah, that sounds fine. We might want to add something later and just surround it by big scary "you probably do not want this" warnings.

PR #761, opened 2023-06-25
Always pass WithLogger first

jonjohnsonjr:
Some options require the use of a logger (e.g. WithConfig), but because we're passing WithLogger after WithConfig, we use the Default logger, which logs to io.Discard.


imjasonh (approved):
Should we stash the logger in the context.Context like we do in Knative?

jonjohnsonjr:
> Should we stash the logger in the context.Context like we do in Knative?

Wouldn't really help in this case because we don't have a context either: https://github.com/chainguard-dev/apko/blob/e6788cf5bbf7a075e21a2a69477c96999df4dd30/pkg/build/options.go#L32

PR #790, opened 2023-07-06
Bump go-apk to pick up faster installs

jonjohnsonjr:


jonjohnsonjr:
Nice, found new races. Makes sense.

imjasonh (approved):

PR #810, opened 2023-07-18
fix: tag-suffix not applied to package version stems

joemiller:
Apply `--tag-suffix` to versions discovered through package-version-stemming (`--package-version-tag-stem true`).

`--tag-suffix string : suffix to use for automatically generated tags`

I am assuming that the `--tag-suffix` flag to `publish` is meant to apply to stemmed tags but the suffix is not being applied. It is only being applied to the full package version tag discovered via `--package-version-tag`.

Fix: apply suffix-tag to stemmed tagas. This is particularly useful with multiple flavors of a package. Consider a case of two variants of a ruby image, regular (no suffix) and `-dev`:

```console
$ apk publish --package-version-tag ruby-3.2 --package-version-tag-stem ./apko.yaml reg/ruby:latest
```
Would produce tags:
- `reg/ruby:3.2.2-r1`
- `reg/ruby:3.2.2`
- `reg/ruby:3.2`
- `reg/ruby:3`

And the `-dev` variant:
```console
$ apk publish --package-version-tag ruby-3.2 --package-version-tag-stem --tag-suffix "-dev" ./dev.apko.yaml reg/ruby:latest-dev
```
Would produce tags:
- `reg/ruby:3.2.2-r1-dev`
- `reg/ruby:3.2.2-r1`
- `reg/ruby:3.2`
- `reg/ruby:3`

This is not ideal because the -dev variant's tags such as `ruby:3` would overwrite the non-dev variant's tag.

With this patch all of the stemmed tags are suffixed, eg:

- `reg/ruby:3.2.2-r1-dev`
- `reg/ruby:3.2.2-dev`
- `reg/ruby:3.2-dev`
- `reg/ruby:3-dev`

### DIFF
--- a/pkg/apk/apk.go
+++ b/pkg/apk/apk.go
@@ -47,10 +47,6 @@ func AdditionalTags(fsys fs.FS, opts options.Options) ([]string, error) {
 			opts.Log.Warnf("Version for package %s is empty", pkg.Name)
 			continue
 		}
-		if opts.TagSuffix != "" {
-			version += opts.TagSuffix
-		}
-		opts.Log.Debugf("Found version, images will be tagged with %s", version)
 
 		additionalTags, err := appendTag(opts, fmt.Sprintf("%s%s", opts.PackageVersionTagPrefix, version))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Will need a rebase -- apologies for stomping all over this file 😅

joemiller:
@jonjohnsonjr thanks for the heads up! rebased

joemiller:
The fix has been incorporated into #825 , closing

PR #819, opened 2023-07-19
Add a test to catch SBOM changes

jonjohnsonjr:
We don't want to accidentally change SBOMs, so this will break if we do.

The wrapper image that cosign generates for SBOMs uses time.Now(), so we have to pull out the layer and test that digest, which is unfortunate, but it's easier than fixing cosign.


kaniini (approved):

imjasonh (approved):

jonjohnsonjr:
> There is a test failure :/

Perhaps it's already working!

PR #820, opened 2023-07-19
Stop exposing build.Context fields

jonjohnsonjr:
I'm not opposed to exposing fields like this in general, but reusing and mutating a build.Context externally has led to me breaking prod, so I'm making these private.


luhring (approved):
Smaller scopes. Higher hopes. ™️

imjasonh (approved):

PR #823, opened 2023-07-20
Remove sbom generator indirection

jonjohnsonjr:
This removes the fake unit tests and the indirection around generated fakes, which exposed some bugs in the impl, which this also fixes.

With that gone, we can remove our dependency on counterfeiter!


kaniini (approved):

internal/tools.go
```diff
@@ -18,7 +18,3 @@
 // This is used to import things required by build scripts, to force `go mod` to see them as dependencies
 
 package internal
```
imjasonh, line 20: Do we need this at all now?

imjasonh (approved):

internal/tools.go
jonjohnsonjr: I'm ok with it existing until the official go blessed way of doing it lands, in case we want to put stuff here.

PR #825, opened 2023-07-23
restore handling of packageTag CLI flags for publish

### LINKED ISSUE #808: some publish options (eg --package-version-tag) broken in 0.9.0
/I believe something happened between 0.8.0 and 0.9.0 that caused some `apko publish` flags to stop working. The first one I noticed is `--package-version-tag` but I believe others might be impacted such as `--package-version-tag-stem`

I am still digging into the code but right now I am assuming this may have happened in or around the build/publish unification PR https://github.com/chainguard-dev/apko/pull/724 when much of the Opts handling in PublishCmd() changed.

Reproduce:
```yaml
contents:
  keyring:
  repositories:
  packages:
    - git
    - git-lfs
    - openssh-client

entrypoint:
  command: /usr/bin/git

work-dir: /home/git

accounts:
  groups:
    - groupname: git
      gid: 65532
  users:
    - username: git
      uid: 65532
      gid: 65532

archs:
  - x86_64
  - aarch64
```

Using both 0.8.0 and 0.9.0, observe that 0.8.0 adds additional tags from the `git` package and 0.9.0 does not:

```console
$ apko publish --debug --package-version-tag=git ./apko.yaml  registry/git:latest --arch host
```

In 0.8.0 this will be seen in the output:
```
❕           | Found version, images will be tagged with 2.41.0-r0
ℹ️            | Returning additional tags [registry/git:2.41.0-r0]
```

I believe other functionality related to `additionalTags` is likely broken too but I have not tested, likely `--package-version-tag-stem`.

I am still digging into the code but since I am unfamiliar with it I am not sure the route to resolve this issue yet. It does seem like this new `additionalTags` var is initialized but never populated  in the PublishCmd func though, and that's where I am starting my dive: https://github.com/chainguard-dev/apko/blob/main/internal/cli/publish.go#L219

~Also, I have noticed the `--sbom` flag to `publish` seems to be broken on 0.8.0 and 0.9.0 as neither of them will produce separate SBOM files on disk (.sbom layers are built and published) and report `WantSBOMS = false` in the debug output.~

deitch:
Fixes #808 

This does the following:

1. Pulls the `additionalTags()` discovery out of the `buildImage` section
2. Has `buildImage()` report back on all installed packages, which is a good artifact relevant to build stage: "I built this image for this arch, here are the packages I installed"
3. Runs through those packages by arch, and checks that the versions are identical, per [this comment](https://github.com/chainguard-dev/apko/issues/808#issuecomment-1646332265), else error
4. Use those now-verified-as-identical tags for the additional tags

In doing so, it also moved some funcs that used to just rely on passing `build.Context.Options` and knowing what they were, to explicit vars. This makes the signature a bit longer, but easier to understand, and no longer dependent on some build context that isn't relevant to that func. I think there is a minimal amount of this after Jon Johnson's restructure last week.

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"sync"
 
+	"github.com/chainguard-dev/go-apk/pkg/apk"
 	coci "github.com/sigstore/cosign/v2/pkg/oci"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -152,7 +153,7 @@ func BuildCmd(ctx context.Context, imageRef, outputTarGZ string, archs []types.A
 	defer os.RemoveAll(wd)
 
 	// build all of the components in the working directory
-	idx, sboms, err := buildImageComponents(ctx, wd, archs, opts...)
+	idx, sboms, _, err := buildImageComponents(ctx, wd, archs, opts...)
 	if err != nil {
 		return err
 	}
@@ -177,13 +178,13 @@ func BuildCmd(ctx context.Context, imageRef, outputTarGZ string, archs []types.A
 
 // buildImage build all of the components of an image in a single working directory.
 // Each layer is a separate file, as are config, manifests, index and sbom.
-func buildImageComponents(ctx context.Context, wd string, archs []types.Architecture, opts ...build.Option) (idx coci.SignedImageIndex, sboms []types.SBOM, err error) {
+func buildImageComponents(ctx context.Context, wd string, archs []types.Architecture, opts ...build.Option) (idx coci.SignedImageIndex, sboms []types.SBOM, pkgs map[types.Architecture][]*apk.InstalledPackage, err error) {
 	ctx, span := otel.Tracer("apko").Start(ctx, "buildImageComponents")
 	defer span.End()
 
 	o, ic, err := build.NewOptions(wd, opts...)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	// cases:
@@ -217,7 +218,7 @@ func buildImageComponents(ctx context.Context, wd string, archs []types.Architec
 	workDir := wd
 	imageDir := filepath.Join(workDir, "image")
 	if err := os.MkdirAll(imageDir, 0755); err != nil {
-		return nil, nil, fmt.Errorf("unable to create working image directory %s: %w", imageDir, err)
+		return nil, nil, nil, fmt.Errorf("unable to create working image directory %s: %w", imageDir, err)
 	}
 
 	imgs := map[types.Architecture]coci.SignedImage{}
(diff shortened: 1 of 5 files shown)


joemiller:
@deitch Looking pretty good. I'm testing it with `git` and `ruby-3.2` packages.

```console
$ apko publish --debug --package-version-tag=ruby-3.2 --package-version-tag-stem ./images/ruby/3.2-dev.apko.yaml localhost:5000/ruby:latest

❕           | Found version, images will be tagged with 3.2.2-r1
❕           | Adding stemmed version tags
ℹ️            | Returning additional tags [localhost:5000/ruby:3.2.2-r1 localhost:5000/ruby:3.2.2 localhost:5000/ruby:3.2 localhost:5000/ruby:3]
❕           | Found version, images will be tagged with 3.2.2-r1
❕           | Adding stemmed version tags
ℹ️            | Returning additional tags [localhost:5000/ruby:3.2.2-r1 localhost:5000/ruby:3.2.2 localhost:5000/ruby:3.2 localhost:5000/ruby:3]
```

```console
$ apko publish --debug --package-version-tag=git --package-version-tag-stem  ./images/git/latest-root.apko.yaml localhost:5000/git:latest

❕           | Found version, images will be tagged with 2.41.0-r0
❕           | Adding stemmed version tags
ℹ️            | Returning additional tags [localhost:5000/git:2.41.0-r0 localhost:5000/git:2.41.0 localhost:5000/git:2.41 localhost:5000/git:2]
❕           | Found version, images will be tagged with 2.41.0-r0
❕           | Adding stemmed version tags
ℹ️            | Returning additional tags [localhost:5000/git:2.41.0-r0 localhost:5000/git:2.41.0 localhost:5000/git:2.41 localhost:5000/git:2]
```

The only remaining issue, which was not mentioned in the original report in #808, is when `--tag-suffix` is specified. ex:

```console
# --tag-suffix -dev

❕           | Found version, images will be tagged with 3.2.2-r1-dev
❕           | Adding stemmed version tags
ℹ️            | Returning additional tags [localhost:5000/ruby:3.2.2-r1-dev localhost:5000/ruby:3.2.2-r1 localhost:5000/ruby:3.2 localhost:5000/ruby:3]
❕           | Found version, images will be tagged with 3.2.2-r1-dev
❕           | Adding stemmed version tags
ℹ️            | Returning additional tags [localhost:5000/ruby:3.2.2-r1-dev localhost:5000/ruby:3.2.2-r1 localhost:5000/ruby:3.2 localhost:5000/ruby:3]
```

I have a PR that should fix this in #810 . The fix is fairly simple in just a few lines by moving the appending of tagSuffix to the end of the package stemming logic. It could be added to this PR or #810 could be rebased after this merges.

deitch:
@joemiller thanks for that. I worked your updates from #810 into this PR, so it should do all you need. Thank you indeed for that, and especially the clear tests.

jonjohnsonjr (approved):

PR #830, opened 2023-07-27
fix --workdir

### LINKED ISSUE #787: `apko build` ignores `--workdir` flag
I spotted that 'apko build' is working outside of directory given using `--workdir` flag.

I suspect this piece of code: https://github.com/chainguard-dev/apko/blame/1e5f9dbfd40406c0219c30ffa9ee34007814a438/internal/cli/build.go#L146
```
func BuildCmd(ctx context.Context, imageRef, outputTarGZ string, archs []types.Architecture, tags []string, wantSBOM bool, sbomPath string, logger log.Logger, opts ...build.Option) error {
	wd, err := os.MkdirTemp("", "apko-*")
```

is Ignoring https://github.com/chainguard-dev/apko/blame/1e5f9dbfd40406c0219c30ffa9ee34007814a438/internal/cli/commands.go#L55C2-L55C2:
```
cmd.PersistentFlags().StringVarP(&workDir, "workdir", "C", cwd, "working dir (default is current dir where executed)")
```

deitch:
Fixes #787 

It was ignoring `--workdir`.

### DIFF
--- a/internal/cli/commands.go
+++ b/internal/cli/commands.go
@@ -36,9 +36,9 @@ func New() *cobra.Command {
 		SilenceUsage:      true,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			http.DefaultTransport = userAgentTransport{http.DefaultTransport}
-			if cwd != "" {
-				if err := os.Chdir(cwd); err != nil {
-					fmt.Printf("failed to change dir to %s: %v\n", cwd, err)
+			if workDir != "" {
+				if err := os.Chdir(workDir); err != nil {
+					fmt.Printf("failed to change dir to %s: %v\n", workDir, err)
 				}
 			}
 		},
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -263,7 +263,11 @@ func buildImageComponents(ctx context.Context, wd string, archs []types.Architec
 			if err != nil {
 				return fmt.Errorf("failed to get installed packages for %q: %w", arch, err)
 			}
+			// this should be unnecessary, as the arch list already is unique,
+			// but the race detector complains, so we can use it anyways
+			mtx.Lock()
 			pkgs[arch] = installed
+			mtx.Unlock()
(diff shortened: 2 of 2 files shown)


internal/cli/commands.go
```diff
@@ -36,8 +36,8 @@ func New() *cobra.Command {
 		SilenceUsage:      true,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			http.DefaultTransport = userAgentTransport{http.DefaultTransport}
-			if cwd != "" {
-				if err := os.Chdir(cwd); err != nil {
+			if workDir != "" {
+				if err := os.Chdir(workDir); err != nil {
 					fmt.Printf("failed to change dir to %s: %v\n", cwd, err)
```
jonjohnsonjr, line 41:
```suggestion
					fmt.Printf("failed to change dir to %s: %v\n", workDir, err)
```

We should also drop `cwd` above? Or do something else with it? I'm not sure what the original intent was.

deitch: Cwd is used as the default for workdir in the flag

deitch:
It is interesting that it is complaining about a race condition in `go test -race`, specifically [here](https://github.com/chainguard-dev/apko/actions/runs/5687717633/job/15416514630#step:4:139):

```
      /Users/adeitcher/Documents/Development/go/src/github.com/chainguard-dev/apko/internal/cli/build.go:266 +0x264
```

That line is part of:

```go
	for _, arch := range archs {
...
		errg.Go(func() error {
...
			pkgs[arch] = installed    // <--- the line with the race condition
```

I guess it is theoretically possibly if the same `arch` appears twice in `archs`, but we already ensure it is unique in `types.ParseArchitectures(archstrs)`. So while it is theoretically possible, it never can happen.

I guess we can put a lock on it, just to ensure this doesn't happen. 🤷‍♂️

deitch:
That should fix it.

deitch:
It is happening now.

kaniini (approved):

PR #847, opened 2023-08-16
fix: publish --stage-tags missing generated tags

joemiller:
The `--stage-tags FILE` flag is only including the tags explicitly set as positional args to `apko publish`. Any tags generated by `--package-version-tag/stem` are not included.

The `additionalTags` slice is empty when this code is reached. Instead we can use the `allTags` slice generated just before this step instead of shadowing it. `allTags` includes a de-duped set of tags that includes both explicit tags from the command line and any generated/discovered tags that come from package-version and stemming.


joemiller:
ping

kaniini (approved):

PR #851, opened 2023-08-18
Don't call build.New for index SBOM

jonjohnsonjr:
We need to call a few methods that happen to hang off of build.Context, but they don't actually depend on the dirfs that build.New generates, which is expensive if we've written a large number of files to disk.

This skips that entirely and uses build.NewOptions to get the data we need to generate the index SBOM.

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -311,17 +311,17 @@ func buildImageComponents(ctx context.Context, wd string, archs []types.Architec
 		build.WithSBOM(imageDir),
 	)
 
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
Before:

```
apko publish --arch amd64 --keyring-append  --repository-append     8.38s user 1.58s system 200% cpu 4.964 total```
```

After:

```
apko publish --arch amd64 --keyring-append  --repository-append     8.28s user 1.55s system 203% cpu 4.832 total
```

imjasonh (approved):
Nice.

PR #853, opened 2023-08-23
fix: publish cmd --sbom-path not writing files

### LINKED ISSUE #804: `apko-publish` doesn't copy SBOM to local `sbomPath`
During the usage of [`apko-snapshot`](https://github.com/chainguard-images/actions/tree/main/apko-snapshot) action I experienced a problem at SBOM attestation step:

```console
...
Run function attest_sbom {
Unable to find SBOM on disk: sbom-index.spdx.json (arch: index). Exiting.
Error: Process completed with exit code 1.
```
I've reproduced it tweaking [`test-scan-apk.yaml`](https://github.com/andros21/actions/commit/df845b4a0bfe046d933ec6f71c5a529f75d0457e) workflow, [here its run with full logs](https://github.com/andros21/actions/actions/runs/5517957909).

It seems that since `apko` started to generate also SBOM in tempdir `apko-*`, SBOM are correctly copied in the requested `sbomPath` when using `apko-build`, thanks to this snip of code:

https://github.com/chainguard-dev/apko/blob/37b82f3a5bd85bde214df382fa974bc8dc88f692/internal/cli/build.go#L164-L170

but not when using `apko-publish`, I was wondering if this is the wanted behavior.
Thanks

joemiller:
fixes #804

quick fix based on the `--sbom-path` logic in the `build` cmd

### DIFF
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
@@ -122,6 +123,7 @@ in a keychain.`,
 			remoteOpts = append(remoteOpts, remote.Reuse(puller))
 
 			if err := PublishCmd(cmd.Context(), imageRefs, archs, remoteOpts,
+				sbomPath,
 				[]build.Option{
 					build.WithLogger(logger),
 					build.WithConfig(args[0]),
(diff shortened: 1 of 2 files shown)


kaniini (approved):

kaniini:
Tests need to be fixed

joemiller:
@kaniini oops. sorry i missed that completely 🤦 

fixed the method sig in the test and also added a basic assertion that sbom files are actually written to the path

kaniini (approved):

PR #861, opened 2023-08-28
Add an internal tarfs implementation

jonjohnsonjr:
This is a fork of go-apk's memfs but with the ability to "write" tar headers and their offsets. We use the memfs implementation mostly for tracking file metadata (directories, permissions, users, and such) and when we actually modify or create files.

For files that are sourced from APKs, we just seek to the file's offset and read from that position without having to actually write anything to disk. This speeds up installation tremendously when dealing with a large number of files.


jonjohnsonjr:
Feedback from code:
* Unit tests for overwriting a file (failures, success).
* Unit tests for etc/group and etc/passwd mutation.
* In general, comments for why we're doing weird stuff.

PR #864, opened 2023-08-31
add overlay-file path mutation

lpcalisi:
`overlay-file` path mutation allows add files to images.

- Some times we need to copy scripts or files inside images to use in CI.
- Exclude ELF files, because these must be provisioned by apks.

### DIFF
--- a/docs/apko_file.md
+++ b/docs/apko_file.md
@@ -191,7 +191,7 @@ will set the environment variable named "FOO" to the value "bar".
 ### Paths
 
 `paths` defines filesystem operations that can be applied to the image. This includes
-setting permissions on files or directories as well as creating empty files, directories and links.
+setting permissions on files or directories as well as creating empty files, directories, links and copy configuration files.
 
 The `paths` element contains the following children:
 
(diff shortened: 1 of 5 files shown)


lpcalisi:
@imjasonh @jonjohnsonjr @kaniini @dlorenc

imjasonh:
We've discussed this before, and I've definitely had cases where I've wanted this, but the reason we haven't added it previously (which I think I'm coming around on) is that files in the image should have some provenance to an APK, which has been built and signed.

We have a number of packages in Wolfi that only contain a file or two, but having those packaged as apks means we know more about where they came from, who put them there, when, etc.

If we added something like this at the apko layer, we should also take care to include as much provenance data as we can about those files in the image's SBOM, even if that's just to say "someone specified this file in the apko.yaml, not in one of the constituent apk packages".

Otherwise, files added this way end up as [dark matter](https://www.chainguard.dev/unchained/software-dark-matter-is-the-enemy-of-software-transparency), being contents of the image that aren't attributable in the SBOM.

lpcalisi:
Right, make sense. Definitely we should have provenance for this files and include it in SBOM. Maybe sign this files?
For future, do you have any idea to extend something like to applications? In order to have provenance and transparency not only in base images also in final images. Use melange to build aplications and install it as proven apk.

kaniini:
Extend what to applications?  You can already build your application with Melange today...

PR #867, opened 2023-08-31
Use tarfs implementation for publish/build

jonjohnsonjr:
Actually opt into the optimized tarfs path.

This cuts the amount of time we spend on I/O dramatically.

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -24,7 +24,6 @@ import (
 	"sync"
 
 	"github.com/chainguard-dev/go-apk/pkg/apk"
-	apkfs "github.com/chainguard-dev/go-apk/pkg/fs"
 	coci "github.com/sigstore/cosign/v2/pkg/oci"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
> 2023/08/31 22:04:11 error during command execution: failed to build layer image for "386": installing apk packages: installing packages: installing alpine-keys: unable to install files for pkg alpine-keys: writing header for "etc/apk/keys/alpine-devel@lists.alpinelinux.org-61666e3f.rsa.pub": conflicting file for "etc/apk/keys/alpine-devel@lists.alpinelinux.org-61666e3f.rsa.pub" has no tar entry

I love having tests.

go.mod
```diff
@@ -2,6 +2,8 @@ module chainguard.dev/apko
 
 go 1.20
 
+replace github.com/chainguard-dev/go-apk => ../go-apk
```
imjasonh, line 5: We remove this right?

imjasonh (approved):

go.mod
jonjohnsonjr: Blah accidentally pushed this while I was grabbing perf numbers.

Issue #74, opened 2022-03-08
OCI Layout support


imjasonh:
apko should support `--oci-layout-path`, like `ko` and Kaniko, to also push OCI image/index contents to a local path. This hooks into other workflows that can (for example) do some scan locally before pushing in a subsequent step. Having a local layout option means there can be a `--push` (default true) flag, since writing the layout might mean you don't want to push.

See https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/layout#Write

dlorenc:
whoops, i was playing with some github project automation and screwed something up.

vito:
Would very much like this! There are certain scenarios where passing around a layout directory rather than a tarball makes a huge performance difference: https://github.com/dagger/dagger/pull/5363

I tried passing `/dev/stdout` as the tarball path and piping to `tar -xf -` to avoid hitting the disk twice, but the resulting directory isn't actually a valid OCI layout because all the blobs are at the root level instead of under `blobs/`. I suppose that means the output tarball isn't actually valid OCI either.

If I find some time I'll try to put a PR together. Seems like it shouldn't be difficult.

PR #882, opened 2023-09-11
Write index as layout if target is a directory

jonjohnsonjr:
This allows us to write out images in a much more efficient format for other tools to consume.

Just writing the whole index as index.json is the simplest thing we could do, and we end up dropping the image tag, but I think this is a fine place to start.

Fixes https://github.com/chainguard-dev/apko/issues/74
Closes https://github.com/chainguard-dev/apko/pull/769

@vito apologies for not seeing your outstanding PR, just folded it into this and rebased.

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 
 	"github.com/chainguard-dev/go-apk/pkg/apk"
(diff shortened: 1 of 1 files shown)


thesayyn (approved):
lgtm

imjasonh (approved):

PR #894, opened 2023-09-18
Add apko dot command

jonjohnsonjr:
This is not 100% great because it relies on some pretty naive heuristics to figure out the dependency graph; however, it does produce some pretty useful graphs without touching any existing go-apk or apko code paths, so I think it's a reasonable place to start.

Ideally, we would have ResolveWorld (or something like it) return a graph with metadata about why things resolved the way they did, but that's a more invasive change.

As is, this is great for understanding what packages would get pulled in by apko and what they depend on and/or provide.

### DIFF
--- a/internal/cli/dot.go
+++ b/internal/cli/dot.go
@@ -1,0 +1,317 @@
+// Copyright 2022, 2023 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"context"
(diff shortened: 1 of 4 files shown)


jonjohnsonjr:
Here's for example [`bazel`](https://github.com/chainguard-images/images/tree/main/images/bazel) from chainguard-images/images:

```
$ apko dot -S bazel/configs/latest.apko.yaml --keyring-append https://packages.wolfi.dev/os/wolfi-signing.rsa.pub --repository-append https://packages.wolfi.dev/os --arch aarch64 | dot -Tpng > graph.png
```

(The `-S` flag makes it look incomplete by omitting redundant edges but mostly that's so it's not too busy to understand.)

image

internal/cli/dot.go
```diff
@@ -0,0 +1,317 @@
...
+	// save the final set we will build
+	archs = ic.Archs
+	o.Logger().Infof("Determining packages for %d architectures: %+v", len(ic.Archs), ic.Archs)
+
+	// The build context options is sometimes copied in the next functions. Ensure
+	// we have the directory defined and created by invoking the function early.
+	defer os.RemoveAll(o.TempDir())
+
+	// TODO: Perhaps we want to support multiple architectures?
+	arch := archs[0]
```
imjasonh, line 114: I don't think we need to, and could probably default to x86_64 and get the same results.

There's no way for an apk to conditionally depend on something based on arch.

imjasonh (approved):

PR #902, opened 2023-09-28
feat: implement resolve command

thesayyn:


go.mod
```diff
@@ -144,3 +144,5 @@ require (
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gotest.tools/v3 v3.1.0 // indirect
 )
+
+replace github.com/chainguard-dev/go-apk v0.0.0-20230906161245-0728258ab917 => github.com/thesayyn/go-apk v0.0.0-20230928231027-0ed5c9ad96a8
```
imjasonh, line 148: Is there a go-apk PR that corresponds to this?

internal/cli/resolve.go
```diff
@@ -0,0 +1,254 @@
...
+	var logPolicy []string
+	var debugEnabled bool
+	var quietEnabled bool
+
+	cmd := &cobra.Command{
+		Use:     "resolve",
+		Short:   "TODO",
+		Long:    `TODO`,
+		Example: `apko resolve <config.yaml>`,
+		Args:    cobra.MinimumNArgs(1),
```
imjasonh, line 87: We can also mark this as hidden, wdyt?

internal/cli/resolve.go
```diff
@@ -0,0 +1,254 @@
...
+	var archstrs []string
+	var output string
+
+	var logPolicy []string
+	var debugEnabled bool
+	var quietEnabled bool
+
+	cmd := &cobra.Command{
+		Use:     "resolve",
+		Short:   "TODO",
```
imjasonh, line 84: I'll suggest an edit here later unless someone beats me to it.

internal/cli/resolve.go
```diff
@@ -0,0 +1,254 @@
...
+			return err
+		}
+
+		resolvedPkgs, err := bc.Resolve(ctx)
+
+		if err != nil {
+			return fmt.Errorf("failed to get package list for image: %w", err)
+		}
+
+		for _, rpkg := range resolvedPkgs {
```
imjasonh, line 202: I think we want to fail if the archs produce inconsistent resolved configs. That's what we do in tf-apko and it generally helps prevent confusion later.

imjasonh (commented):
Nice! Just a few small comments, this looks really good.

Sorry for the delay in reviewing it.

go.mod
thesayyn: oh yes, https://github.com/chainguard-dev/go-apk/pull/125

internal/cli/resolve.go
thesayyn: totally. no need to advertise it in the help.

internal/cli/resolve.go
thesayyn: do you have a link I can look at? I am not sure about what to do here.

imjasonh: https://github.com/chainguard-dev/terraform-provider-apko/blob/002f3cb0e3d06f356357e0e484224a54bbc667f6/internal/provider/config_data_source.go#L342 is where we align these in tf-apko -- it's pretty gnarly unfortunately.

Basically if two archs of the same image can't agree about what packages are in there (e.g., because arm builds are ahead of x86) then we fail rather than produce an image with inconsistent versions.

We could do the same here, possibly by resolving off x86 and checking that arm is the same. `apko` itself won't complain about this, but maybe it should.

It's not a blocker, but something we might want a TODO about in case this causes confusion in the future.

internal/cli/resolve.go
```diff
@@ -0,0 +1,251 @@
...
+	repositories := map[string]bool{}
+
+	for _, keyring := range ic.Contents.Keyring {
+		lock.Contents.Keyrings = append(lock.Contents.Keyrings, lockKeyring{
+			Name: stripURLScheme(keyring),
+			URL:  keyring,
+		})
+	}
+
+	for _, arch := range archs {
```
imjasonh, line 185:
```suggestion
	// TODO: If the archs can't agree on package versions (e.g., arm builds are ahead of x86) then we should fail instead of producing inconsistent locks.
	for _, arch := range archs {
```

go.mod
```diff
@@ -144,3 +144,5 @@ require (
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gotest.tools/v3 v3.1.0 // indirect
 )
+
+replace github.com/chainguard-dev/go-apk => github.com/thesayyn/go-apk v0.0.0-20231011191853-e0653a72ad44
```
imjasonh, line 148: We should remove this replace before merging.

thesayyn: done

thesayyn:
@imjasonh  could you kick the CI?

thesayyn:
fixed the lint. need the CI run again.

thesayyn:
@imjasonh could you please run the ci again?

imjasonh (approved):

PR #903, opened 2023-09-29
Improve publish cmd error & usage msg.

### LINKED ISSUE #817: Apko publish does not specify what argument is missing
````
apko publish images/ingress-nginx-controller/configs/latest.apko.yaml -r ~/Documents/code/go/src/github.com/wolfi-dev/os/packages -k ~/Documents/code/go/src/github.com/wolfi-dev/os/local-melange.rsa.pub --local --arch $(uname -m) --debug
Error: requires at least 2 arg(s), only received 1
2023/07/19 09:07:12 error during command execution: requires at least 2 arg(s), only received 1
````

Also there are no markdown doc like there is for melange, https://github.com/chainguard-dev/melange/tree/main/docs/md . That might help some but the cli should let the user know which argument is missing.

### PR DESCRIPTION
**Pull request by debasishbsws**
Fixes #817 

Though the command help has a description of which arguments we need.
I have added a custom error msg. (It may not require as the command --help is very clear about the args, suggestions please) 

I am seeing almost all of the command `Usage:` are written in the `Example:` sections.
I have change it for the publish command.

Current help:
image

### DIFF
--- a/internal/cli/publish.go
+++ b/internal/cli/publish.go
@@ -70,14 +70,13 @@ func publish() *cobra.Command {
 	var offline bool
 
 	cmd := &cobra.Command{
-		Use:   "publish",
+		Use:   "publish <config.yaml> <tag...>",
 		Short: "Build and publish an image",
 		Long: `Publish a built image from a YAML configuration file.
 
 It is assumed that you have used "docker login" to store credentials
 in a keychain.`,
-		Example: `  apko publish <config.yaml> <tag...>`,
-		Args:    cobra.MinimumNArgs(2),
+		Example: `  apko publish hello-world.yaml hello:v1.0.0 --quiet`,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: internal/cli/publish.go [hunk 1]
```diff
@@ -87,6 +86,10 @@ in a keychain.`,
 				}
 			}
 
+			if len(args) < 2 {
```
mattmoor, line 89: We expect exactly two positional arguments

mattmoor: Maybe that's not true, lemme dig in and refresh my memory 🤔

mattmoor: Nevermind, you are right.  TIL you can push to N tags at once

imjasonh (approved):
Looks great! Thanks for this contribution 🙌

PR #908, opened 2023-10-02
Bug Fix: calling the function in `show-config` cmd

debasishbsws:
Looks like we have some missing parenthesis here😅 and never calling the function.


Elizafox (approved):

Elizafox:
LGTM 👍

PR #932, opened 2023-10-26
Pass UID and GID mapping to the tarball writer

epsilon-phase:
Required for https://github.com/chainguard-dev/melange/pull/784


Elizafox (approved):

Elizafox:
This depends on chainguard-dev/go-apk#145 and once that's done and the dep is updated this should work AIUI?

Elizafox (approved):

PR #934, opened 2023-10-26
Update NEWS.md for 0.11.1

Elizafox:
We should cut a release to facilitate a fix for chainguard-dev/melange#501 in downstream providers.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+# Changes from 0.11.0 to 0.11.1
+
+* Add JSON tags to ImageConfiguration types: https://github.com/chainguard-dev/apko/pull/933
+
+* Pass UID and GID mapping to the tarball writer: https://github.com/chainguard-dev/apko/pull/932
+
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -1,3 +1,11 @@
+# Changes from 0.11.0 to 0.11.1
+
+* Add JSON tags to ImageConfiguration types: https://github.com/chainguard-dev/apko/pull/933
+
+* Pass UID and GID mapping to the tarball writer: https://github.com/chainguard-dev/apko/pull/932
+
+Full diff: https://github.com/chainguard-dev/apko/compare/v0.11.0..main
```
imjasonh, line 7: This is only going to be correct until the next commit. Can this be either the current head commit or 0.11.1 in preparation?

Elizafox: I just went off of what the previous news entry sent but I did think about that. I could make it the tag?

imjasonh (approved):

PR #942, opened 2023-10-31
cleanup: remove unused flags

imjasonh:
In our move toward Terraform-based image builds, we no longer use a number of flags we had previously used -- mainly around tagging.

This PR drops some flags we don't use anymore, in an attempt to delay the implacable advance of entropy that lies at the heart of the human experience.

- `--stage-tags` used to skip tagging -- now we just invoke the build and tag later with [`oci_tag`](https://registry.terraform.io/providers/chainguard-dev/oci/latest/docs/resources/tag)
- `--use-docker-mediatypes` used to use Docker-style media types -- now we just always publish OCI images
- `--package-version-tag`, `--package-version-tag-stem`, `--package-version-tag-prefix`, `--tag-suffix` all used to control version-tagging logic in apko; now we do this version string extraction and tagging in [`apko_tags`](https://registry.terraform.io/providers/chainguard-dev/apko/latest/docs/data-sources/tags) and `oci_tag`

This also unexports `GenerateOSRelease` which was only ever called from the same package.

No tests were harmed in the removal of this behavior. 🙃

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"sync"
 
-	"github.com/chainguard-dev/go-apk/pkg/apk"
 	"github.com/google/go-containerregistry/pkg/v1/layout"
 	coci "github.com/sigstore/cosign/v2/pkg/oci"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


cpanato (approved):

joemiller:
I was using the --skip-tags and --package-version-* flags. :(

PR #961, opened 2023-11-17
drop support for CycloneDX SBOMs

imjasonh:
Despite CycloneDX being described as one of the default SBOM formats, when publishing, we pick the first one (which is `spdx` by default) and publish that only.

https://github.com/chainguard-dev/apko/blob/4f9a4c617d079adc838c47303154d90a5ee4455e/pkg/build/oci/sbom.go#L153-L157

As a result, this code path is effectively never used (by Chainguard, at least), and since there are no tests or validation we have no idea whether this works or will continue to. Instead of having effectively dead code, let's just remove it.

TF-apko already only supports SPDX ([ref](https://github.com/chainguard-dev/terraform-provider-apko/pull/43#discussion_r1179557634))


imjasonh:
On second thought, let's not go to Camelot. 'Tis a silly place.

PR #962, opened 2023-11-17
WIP: fix and continuously validate SBOMs

imjasonh:


pkg/sbom/generator/spdx/spdx.go
```diff
@@ -366,6 +364,8 @@ func (sx *SPDX) imagePackage(opts *options.Options) (p *Package) {
 			"SPDXRef-Package-%s", opts.ImageInfo.ImageDigest,
 		)),
 		Name:             opts.ImageInfo.ImageDigest,
+		Version:          opts.ImageInfo.ImageDigest,
+		Supplier:         "Organization: Chainguard, Inc", // TODO
```
imjasonh, line 368: Should I pull this out of `/etc/os-release` if present? Isn't there somewhere else we do that?

jonjohnsonjr (approved):

.github/workflows/build-samples.yml
```diff
@@ -31,6 +33,49 @@ jobs:
...
+      - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v2.1.5
+        with:
+          go-version-file: 'go.mod'
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@68827325e0b33c7199eb31dd4e31fbe9023e06e3 # v3.0.0
+
+      - name: build
+        run: |
+          make apko
+          ./apko version
```
hectorj2f, line 62: Why do we need this?

hectorj2f (commented):
LGTM, 🆒

hectorj2f:
I can think of an attestation for the images with the output of the ntia conformance test. Wdyt?

.github/workflows/build-samples.yml
```diff
@@ -31,6 +33,49 @@ jobs:
         run: |
           ./apko build ./examples/nginx.yaml nginx:build /tmp/nginx-${{ matrix.arch }}.tar --debug --arch ${{ matrix.arch }}
 
+      - name: Check SBOM Conformance
+        run: |
+          set -euxo pipefail
+          for f in *.spdx.json; do
```
mattmoor, line 39: Here and elsewhere it would be good if the check failed if we were to stop emitting SBOMs here.  I'm not sure what this would do if the wildcard matches 0 things...

.github/workflows/build-samples.yml
```diff
@@ -31,6 +33,49 @@ jobs:
...
+    name: build-nginx-multiarch
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v2.1.5
+        with:
+          go-version-file: 'go.mod'
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@68827325e0b33c7199eb31dd4e31fbe9023e06e3 # v3.0.0
```
mattmoor, line 57: Why do we need QEMU?

mattmoor:
> I can think of an attestation for the images with the output of the ntia conformance test

I'm kinda 👎 on this, since it's something that should be a trivially verifiable property of the SBOM.

.github/workflows/build-samples.yml
imjasonh: Just a sanity check. I copied it from the other one.

PR #966, opened 2023-11-20
Fix packages with multiple Replaces

jonjohnsonjr:
We were ignoring multiple Replaces when installing files, which broke any packages that relied on replacing multiple packages. This fixes things for those packages, but also fixes the installed db file to include the replaces line that was missing.

### DIFF
--- a/pkg/tarfs/fs.go
+++ b/pkg/tarfs/fs.go
@@ -491,7 +491,14 @@ func (m *memFS) writeHeader(name string, te tarEntry) error {
 	}
 
 	sameOrigin := got.pkg.Origin == want.pkg.Origin
-	replaces := got.pkg.Name == want.pkg.Replaces
+
+	replaces := false
+	for _, replace := range want.pkg.Replaces {
+		if got.pkg.Name == replace {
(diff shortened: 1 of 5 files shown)


pkg/tarfs/fs.go
```diff
@@ -491,7 +491,14 @@ func (m *memFS) writeHeader(name string, te tarEntry) error {
 	}
 
 	sameOrigin := got.pkg.Origin == want.pkg.Origin
-	replaces := got.pkg.Name == want.pkg.Replaces
+
+	replaces := false
+	for _, replace := range want.pkg.Replaces {
```
found-it, line 496: Do we need to check both directions? It's valid if the package I'm installing replaces the one that exists, or if the package that's installed replaces the one I'm installing right?

hectorj2f (approved):
lgtm

pkg/tarfs/fs.go
jonjohnsonjr: I think you're right but I think that's a bit different from the current behavior, so I'm going to merge this and we can fix that later too.

PR #976, opened 2023-12-05
Allow run_as field to be a string or int for json/yaml validation.

wlynch:
To appease the linters


coreydaley-cg (changes requested):
Right fix, wrong type keyword. invopop emits `oneof_type` values verbatim, so this produces `"type": "int"`, which isn't a JSON Schema type. Both the struct tag and `schema.json` need `integer`. The YAML decoder already accepts an integer into the string field, so once the schema agrees this is complete.

PR #979, opened 2023-12-06
Support for locking packages to the versions from the 'apko.lock.json' file.

sfc-gh-ptabor:
This PR allows to constraint APK resolution to packages in the version locked in the .resolved.apko.json.

1. Apko is (already) able to produce a `lock` file with resolved SHA's, link & versions of artifacts. 
2.Apko is (already) able to work in 'offline` mode that requires all the files to be available locally. 
3. Rules_OCI uses the `apko.lock.json` to pre-download all needed artifacts...
4. But the  apko is not aware (and this PR changes it) -> what are the exact versions of packages that should be used in produced images).   In particular -> if APKINDEX get's a new versions of packages, the hermetic build will fail. 

Now apko can be called like that:
```
apko lock foo.apko.yaml  # that produces foo.apko.lock.json. 
apko build -lockfile=./foo.apko.lock.json  foo.apko.yaml foo foo.tar
```
and this will make sure the exact same packages are installed (and usage of APKINDEX will get entirely skipped)

The motivation is here: 
  https://github.com/chainguard-dev/rules_apko/issues/13#issuecomment-1839642859


jonjohnsonjr:
This looks pretty close to what I was toying around with.

One change in go-apk that would make this a little cleaner is here: https://github.com/chainguard-dev/go-apk/compare/main...jonjohnsonjr:go-apk:lock

And then we'd be able to skip over FixateWorld if we were given a lock file and just call InstallPackages directly with the packages in the lockfile, like this:

https://github.com/chainguard-dev/apko/compare/chainguard-dev:691fe51...jonjohnsonjr:5b34ad7#diff-2376e897ecf97b9e4ef42c3e1cfee979ffe76e2f346054a8cba586fffc481178R115-R128 

I'm not sure if this seems simpler to you than using your package filter approach? Interested in feedback.

sfc-gh-ptabor:
@jonjohnsonjr - Thank you for suggestions. Migrated this to InstallPackages approach and added tests. 
Ready for a review pass. 

The apk-go dependency is also ready in: 
https://github.com/chainguard-dev/go-apk/pull/172

thesayyn:
Now that its a lock file it think we should stop using the name .resolved.json. Instead it should be .lock.json

sfc-gh-ptabor:
> Now that its a lock file it think we should stop using the name .resolved.json. Instead it should be .lock.json

Happy to change if this the agreed direction where we prefer (incompatible) cleanup over incremental additions... 
(e.g. `apko resolve` would need to be renamed to `apko lock`). @jonjohnsonjr @imjasonh WDYT ?

imjasonh:
I'm personally fine breaking the CLI for `apko resolve` -> `apko lock`. The command is hidden for this reason, and the only user is rules_apko, which should be capable of handling this change if I understand correctly.

I don't think @jonjohnsonjr has a strong opinion either, and if @thesayyn thinks it makes more sense as .lock.json and `apko lock` I think we're all in rough agreement.

sfc-gh-ptabor:
@thesayyn -> I migrated the change to support `apko build --lockfile=foo.lock.json` as well as support `apko lock foo.apko.yaml`.  Please take a look.

sfc-gh-ptabor:
@jonjohnsonjr @thesayyn @imjasonh : Could you please take a look. Thank you.

thesayyn (commented):
LGTM

imjasonh (approved):

PR #996, opened 2023-12-21
feat: improve local include

lpcalisi:
- `include` in apko file now supports merge configurations for accounts, environments and path mutations.
- remove remote include
- add tests

related #1008

PR #1008, opened 2024-01-10
remove include: feature

imjasonh:
Opening for discussion.

We don't use this feature, and I don't know of anybody who does or would. It presents a potentially confusing or abuse-prone surface.

### DIFF
--- a/docs/apko_file.md
+++ b/docs/apko_file.md
@@ -208,9 +208,6 @@ The `paths` element contains the following children:
  - `permissions`: file permissions to set. Permissions should be specified in octal e.g. 0o755 (see `man chmod` for details).
  - `source`: used in `hardlink` and `symlink`, this represents the path to link to.
 
-
-### Includes
-
 `include` defines a path to a configuration file which should be used as the base configuration,
 the configuration data is layered on top of this base configuration.  By default, there is no
 base configuration used.
--- a/examples/include.yaml
+++ b/examples/include.yaml
@@ -1,1 +1,0 @@
-include: examples/alpine-base.yaml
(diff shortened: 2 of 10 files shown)


lpcalisi:
I was thinking about that include feature to incorporate some base configurations to all apko definitions, like inheritance. Users, nvironments variables, paths, or packages for tooling in some cases.

imjasonh:
> @imjasonh I was thinking about that include feature to incorporate some base configurations to all apko definitions, like inheritance. Users, nvironments variables, paths, or packages for tooling in some cases.

That seems reasonable. I would still like to remove remote includes, since that seems more likely to cause bad bugs than helpful experiences.

I'd also like to invest in better (any) tests for this feature if anyone is actually using it. There might be confusing corner cases involved in merging two configs, that we should at least document in a test.

If you're interested in adding those tests or using that behavior, let me know.

lpcalisi:
i could add some tests and improve documentation about how use this feature, if it still make sense for you.

A couple a weeks ago i made this #996 in order to extend definitions of include, i would try to work hard in this PR to add tests.

imjasonh:
Sorry I missed that PR. It looks good so far, more tests are always welcome.

Sounds like we should keep local includes, but not remote. Does that sound good?

lpcalisi:
@imjasonh i made the changes, could you take a look?

coreydaley-cg:
Closing as a stale draft: the author has moved on and the branch no longer applies. For the record, the remote-fetch half this was most concerned about is already gone (`pkg/fetch` no longer exists, though `docs/apko_file.md` still describes it). What's left is the local-file merge, still marked deprecated, with no config in our repos using it. Reopen if someone wants to finish the removal.

PR #1012, opened 2024-01-17
Store checksum of apko-config in the lock-file to detect changes in origin.

sfc-gh-ptabor:
Without this, it was easy to forgot updating lock-files, so got images based on the stale lock-file.

Now, if the `apko.yaml` (or one of the included files) is changed without refreshing the `apko.lock.json` file, the 
apko build will fail. 
The behavior is backward compatible, i.e. it triggers only when the new lock-files (with `config { deepChecksum: ...}} fields`) are used.

### DIFF
--- a/pkg/build/types/image_configuration.go
+++ b/pkg/build/types/image_configuration.go
@@ -17,6 +17,7 @@ package types
 import (
 	"context"
 	"fmt"
+	"hash"
 	"os"
 	"strings"
 
@@ -45,8 +46,9 @@ func (ic *ImageConfiguration) ProbeVCSUrl(ctx context.Context, imageConfigPath s
 }
 
 // Parse a configuration blob into an ImageConfiguration struct.
-func (ic *ImageConfiguration) parse(ctx context.Context, configData []byte) error {
+func (ic *ImageConfiguration) parse(ctx context.Context, configData []byte, configHasher hash.Hash) error {
 	log := clog.FromContext(ctx)
+	configHasher.Write(configData)
 	if err := yaml.Unmarshal(configData, ic); err != nil {
 		return fmt.Errorf("failed to parse image configuration: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


sfc-gh-ptabor:
@imjasonh - please take a look. It removes one of the surprises when using lock-files.

imjasonh (commented):
The API change to `build.NewOptions` is a bit unexpected, and might cause problems for some cases where we invoke apko as a library on image configuration contents in memory, without a file being present. The terraform provider for example mainly passes around configs in TF state, and not files on disk.

In general I'd like to have apko-as-a-library be less file-oriented, not more. Would it make sense to have `options.Options` have a method to return the `configPath` and `configChecksum`? And possibly fail if no file was passed in in the first place?

sfc-gh-ptabor:
> The API change to `build.NewOptions` is a bit unexpected, and might cause problems for some cases where we invoke apko as a library on image configuration contents in memory, without a file being present. The terraform provider for example mainly passes around configs in TF state, and not files on disk.
> 
> In general I'd like to have apko-as-a-library be less file-oriented, not more. Would it make sense to have `options.Options` have a method to return the `configPath` and `configChecksum`? And possibly fail if no file was passed in in the first place?

@imjasonh PTAL. I added this as explicit field in Options. It needs to be persisted somewhere... and in fact it's kind of an Option derived by the method 'Load' from the user's setting. 
In scenarios where there will be alternative source of Config to the file, the logic parsing it, has freedom to set the ConfigName and Checksum to whatever they want.

pkg/build/types/image_configuration.go
```diff
@@ -101,25 +103,28 @@ func (ic *ImageConfiguration) parse(ctx context.Context, configData []byte) erro
...
 
-	return ic.parse(ctx, data)
+	return ic.parse(ctx, data, configHasher)
 }
 
-// Loads an image configuration given a configuration file path.
-func (ic *ImageConfiguration) Load(ctx context.Context, imageConfigPath string) error {
+// Load - loads an image configuration given a configuration file path.
+// Returns checksum (sha256-) of the loaded config or error.
+func (ic *ImageConfiguration) Load(ctx context.Context, imageConfigPath string, configHasher hash.Hash) error {
```
imjasonh, line 117: What should callers pass here if they're not interested in locked/hashed configs?

sfc-gh-ptabor: Updated the comment about this method.

sfc-gh-ptabor:
@imjasonh : Rebased the change on top of recent changes. Please take a look.

imjasonh (approved):

PR #1028, opened 2024-02-05
use charm logger

imjasonh:
This moves away from apko and related tools having its own opinionated logger, leaning further into slog's general structured logging and letting callers dictate the format they want in their output.

In this change, the `apko` CLI itself will use [`charmbracelet/log`](https://pkg.go.dev/github.com/charmbracelet/log#section-readme) with the timestamp displayed, instead of promoting the arch to a log prefix.

Before:
Screenshot 2024-02-05 at 1 24 56 PM

After:
Screenshot 2024-02-05 at 1 24 35 PM


joshrwolf:
this is going to seriously mess with my fruedian developed dopamine hit looking at the apk summary logs

jdolitsky (approved):

PR #1034, opened 2024-02-07
move some logs to debug, avoid duplicate work/logs

imjasonh:
We were probing VCS URLs for each arch, which meant we were also logging it for each arch.

(In our own usage via tf-apko we don't use this at all, but it seemed silly to do it multiple times unnecessarily even for CLI users)

This also changes the default SBOM formats to SPDX only, since `apko publish` kept warning that we were only going to publish SPDX anyway, and it seemed silly for the default to warn you.


imjasonh:
Basic usage after this change:

Screenshot 2024-02-07 at 9 33 01 AM


Screenshot 2024-02-07 at 9 33 30 AM

You can get more logs with `--log-level=debug`

luhring (approved):

PR #1039, opened 2024-02-10
Drop cosign dep

jonjohnsonjr:
It's too far gone. We can't escape the dependency hell.

For our own purposes, we don't even use the SBOMs that get attached.


jonjohnsonjr:
Another way we could see similar benefit is by splitting this out into an optional thing for library dependents but still incorporate that into the CLI surface.

jonjohnsonjr:
Revived as https://github.com/chainguard-dev/apko/pull/1645

PR #1054, opened 2024-03-01
Base image mvp

sfc-gh-mhazy:
MVP for building APKO image on top of existing image. Multiple things related to metadata not working, but I ran the go binary on top of the created image.

**To build**
```
go run main.go build mhazy_test/glibc_openssl.yaml tag mhazy_test/out_on_top/ --arch x86_64 --lockfile mhazy_test/glibc_openssl.resolved.json
```

**Rough overview of changes**

* `existing_image.go` - utilities for parsing existing image.
* `build_implementation.go` - get list of packages from (hardcoded) base image. Set packages to install as `packages to install` \ `packages in base` 
* `image.go` - instead of starting from empty image, use base and add layer that we just built to it. 

**Topics to discuss / implement**
1. Files that we should overwrite from base (etc/apk/world , installed file etc)
2. Conflicts in packages 
3. Conflicts in files (what if package from upper layer would overwrite files of other package from previous layer? When building as a single layer apk would in such case)
4. User accounts
5. Environment variables ?
6. **Current MVP works only for limited set of cases**: The image that we build must use lock file and we must produce OCI image layout (for tar it doesn't work). Apko produces image index, so I also assumed base image is an index.
7. SBOM generation
8. API - how to expose the new option?


pkg/build/build_implementation.go
```diff
@@ -148,6 +167,7 @@ func (bc *Context) buildImage(ctx context.Context) error {
 		}
 	}
 
+	// TODO: What about accounts ?
```
sfc-gh-dszot, line 170: For MVP I would assume we inherit it from the base image.

A slightly better option would be to give the possibility to override all the accounts.

Giving to much flexibility by allowing both, adding and overwriting will make API more complex - we should consider if it's worth it.

PR #1082, opened 2024-04-05
Increase gzip compression level to 9 (BestCompression)

lyoung-confluent:
Currently when building an image apko uses `pgzip.NewWriter` which will use the `DefaultCompression` constant (`-1`) which is [becomes a compression level of 5](https://github.com/klauspost/compress/blob/1e2b27500c4eae851ba753ad116dc5a6524d2366/flate/deflate.go#L822). We can decrease the final image/layer size at the cost of small amount of additional CPU during build time by instead opting into the `BestCompression` level (9) via `pgzip.NewWriterLevel` instead. 

This won't necessarily result in a huge change in image size for most images, for example using `examples/alpine-base.yaml` it decreases from `27691520` bytes to `26865152` bytes (~3%).

### DIFF
--- a/pkg/build/build_implementation.go
+++ b/pkg/build/build_implementation.go
@@ -89,7 +89,7 @@ func (bc *Context) BuildTarball(ctx context.Context) (string, hash.Hash, hash.Ha
 	digest := sha256.New()
 
 	buf := bufio.NewWriterSize(outfile, 1<<22)
-	gzw := gzip.NewWriter(io.MultiWriter(digest, buf))
+	gzw := gzip.NewWriterLevel(io.MultiWriter(digest, buf), gzip.BestCompression)
 	if err := gzw.SetConcurrency(1<<20, pgzipThreads); err != nil {
 		return "", nil, nil, 0, fmt.Errorf("tried to set pgzip concurrency to %d: %w", pgzipThreads, err)
 	}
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
I think this would be great to add as an option/flag, but I'm reluctant to make it the default. We want building (and rebuilding) images to be fast during development so we can get a fast feedback loop. Since `melange` uses `apko` to produce the build environment for our package builds, the vast majority of apko-produced images are effectively "single use" so it would be wasteful to burn the additional CPU to make them smaller, but it's definitely unfortunate that we're leaving some gains on the table for the final release images.

(Arguably, it might make more sense to have 9 be the default and have a flag to go faster during development...)

I think klauspost's choice for default strikes a good balance (some older discussion [here](https://blog.klauspost.com/rebalancing-deflate-compression-levels/)), but I also spent a month last year chasing performance as much as possible, so I might have gone a little too far :)

Would you be okay with this as a flag?

xnox:
To win on both speed, and image size.... We should consider Zstd compression. But I have no idea if the ecosystem of where images are executed can support that.

jonjohnsonjr:
> But I have no idea if the ecosystem of where images are executed can support that.

I feel like the major runtimes (containerd, mostly) probably already support zstd, but I'd be a little worried about things like Cloud Run. Would be worth experimenting with.

xnox:
possibly we would need to do dual publication of gzip + zstd like fedora is pushing https://discussion.fedoraproject.org/t/switch-fedora-container-images-to-support-zstd-chunked-format-by-default/123712/44

coreydaley-cg:
Closing per the 2024 discussion: the default stays as it is for build-loop speed, and the thread's direction moved to zstd. #2376 adds a `--compression` flag, which is where a gzip level option belongs if someone wants one. Thanks for the measurement.

PR #1096, opened 2024-04-21
feat: allow creating new files

shyim:
Allow creating new files like config files when building an image with apko.

example config:

```
contents:
  repositories:
    - https://dl-cdn.alpinelinux.org/alpine/edge/main
    - https://dl-cdn.alpinelinux.org/alpine/edge/community
  packages:
    - alpine-base
    - php83

paths:
  - type: new-file
    path: /etc/php83/conf.d/memory.ini
    content: |
      memory_limit=2G

entrypoint:
  command: /usr/bin/php83 -i
```

and then run:

```
go run . build apko.yaml php-test php.tar
docker load < php.tar

docker run --rm php-test:latest-amd64 | grep memory
```

and you see:
```
Additional .ini files parsed => /etc/php83/conf.d/memory.ini
memory_limit => 2G => 2G
```

### DIFF
--- a/pkg/build/paths.go
+++ b/pkg/build/paths.go
@@ -33,6 +33,7 @@ var pathMutators = map[string]PathMutator{
 	"hardlink":    mutateHardLink,
 	"symlink":     mutateSymLink,
 	"permissions": mutatePermissions,
+	"new-file":    mutateNewFile,
 }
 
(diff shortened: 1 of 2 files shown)


lpcalisi:
we discuss this before https://github.com/chainguard-dev/apko/pull/864

PR #1127, opened 2024-05-16
spdx: allow specifying custom license

xnox:
If licenserefs are passed, include them in the SPDX document.

Related: https://github.com/chainguard-dev/melange/issues/1212

Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>

### DIFF
--- a/pkg/sbom/generator/spdx/spdx.go
+++ b/pkg/sbom/generator/spdx/spdx.go
@@ -470,7 +470,7 @@ type ExternalDocumentRef struct {
 
 // Can also contain name, comment, seeAlso
 type LicensingInfo struct {
-	LicenseId          string    `json:"licenseId"`
+	LicenseID          string    `json:"licenseId"`
 	ExtractedText      string    `json:"extractedText"`
 }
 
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

pkg/sbom/generator/spdx/spdx.go
```diff
@@ -467,6 +468,12 @@ type ExternalDocumentRef struct {
 	SPDXDocument       string   `json:"spdxDocument"`
 }
 
+// Can also contain name, comment, seeAlso
+type LicensingInfo struct {
+	LicenseId          string    `json:"licenseId"`
```
imjasonh, line 473:
```suggestion
	LicenseID          string    `json:"licenseID"`
```

xnox: @imjasonh 

lefthand side is internal; right hand side of the the spec is case insensitive, the tag is upper case https://spdx.github.io/spdx-spec/v2.3/other-licensing-information-detected/#101-license-identifier-field the example is camel-case Id, the sample json files in the tools-java https://github.com/spdx/tools-java/blob/d95178bbc5c96e9b2afac7b51f271e8ddfe7105e/testResources/SPDXJSONExample-v2.3-with-exception.spdx.json#L21 show it as camel case. 

May i please keep the right hand side as is? i.e. "licenseId" to ensure the output we produce matches the spdx sample files?

imjasonh: Oh absolutely, sorry about my overzealous capitalization

PR #1128, opened 2024-05-16
spdx: allow comment in ExternalRef

xnox:
Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>


xnox:
not interested.

PR #1133, opened 2024-05-20
sbom: fixup merging LicensingInfos during Image SBOM generation

xnox:
This fixes https://github.com/chainguard-dev/apko/pull/1127

With package SBOMs now containing custom license information (i.e. font-ubuntu), teach Image SBOM generator to merge those.

Error out if two packages use the same licence-ref but with different extracted license text, as that would be an invalid SBOM merge. This feels nicer than force renaming licenseIDs with multiple duplicate
texts.

Tested by generating apko image with just font-ubuntu package, and verifying that with this change, the image sbom now passes checkers.

This should fix:
> Unrecognized license reference: LicenseRef-ubuntu-font. license_expression must only use IDs from the license list or extracted licensing info, but is: LicenseRef-ubuntu-font

Error from the docker-selenium spdx json

### DIFF
--- a/pkg/sbom/generator/spdx/spdx.go
+++ b/pkg/sbom/generator/spdx/spdx.go
@@ -90,10 +90,11 @@ func (sx *SPDX) Generate(opts *options.Options, path string) error {
 			},
 			LicenseListVersion: "3.16",
 		},
-		DataLicense:   "CC0-1.0",
-		Namespace:     "https://spdx.org/spdxdocs/apko/",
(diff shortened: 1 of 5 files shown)


imjasonh (commented):
Could you add a simple test to cover this behavior and prevent future regressions?

xnox:
> Could you add a simple test to cover this behavior and prevent future regressions?

Added test case. please review. It is single package image sbom comparison.

imjasonh (approved):

PR #1144, opened 2024-05-30
This fixes the boolean logic to pass auth.

mattmoor:
As written we refuse to pass auth unless the `HTTP_AUTH` variable DOES NOT have 4 parts, but we really want the opposite.

This change makes things more strict and we will actually error out if `HTTP_AUTH` is specified but malformed.

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -98,7 +98,7 @@ Along the image, apko will generate SBOMs (software bill of materials) describin
 			} else if parts := strings.SplitN(auth, ":", 4); len(parts) != 4 {
 				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %d parts)", len(parts))
 			} else if parts[0] != "basic" {
-				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %q for first parth)", parts[0])
+				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %q for first part)", parts[0])
 			} else {
 				// NB: parts[1] is the realm, which we ignore.
 				user, pass = parts[2], parts[3]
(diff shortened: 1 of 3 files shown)


internal/cli/publish.go
```diff
@@ -100,7 +100,13 @@ in a keychain.`,
 			defer os.RemoveAll(tmp)
 
 			var user, pass string
-			if parts := strings.Split(os.Getenv("HTTP_AUTH"), ":"); len(parts) != 4 && parts[0] == "basic" {
+			if auth, ok := os.LookupEnv("HTTP_AUTH"); !ok {
+				// Fine, no auth.
+			} else if parts := strings.SplitN(auth, ":", 4); len(parts) != 4 {
+				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %d parts)", len(parts))
+			} else if parts[0] != "basic" {
+				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %q for first parth)", parts[0])
```
imjasonh, line 108:
```suggestion
				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %q for first part)", parts[0])
```

internal/cli/build.go
```diff
@@ -93,7 +93,13 @@ Along the image, apko will generate SBOMs (software bill of materials) describin
 			defer os.RemoveAll(tmp)
 
 			var user, pass string
-			if parts := strings.Split(os.Getenv("HTTP_AUTH"), ":"); len(parts) != 4 && parts[0] == "basic" {
+			if auth, ok := os.LookupEnv("HTTP_AUTH"); !ok {
+				// Fine, no auth.
+			} else if parts := strings.SplitN(auth, ":", 4); len(parts) != 4 {
+				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %d parts)", len(parts))
+			} else if parts[0] != "basic" {
+				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %q for first parth)", parts[0])
```
imjasonh, line 101:
```suggestion
				return fmt.Errorf("HTTP_AUTH must be in the form 'basic:REALM:USERNAME:PASSWORD' (got %q for first part)", parts[0])
```

imjasonh (approved):

PR #1146, opened 2024-05-31
plumb through HomeDir as optional build configuration

joshrwolf:
This is already supported in the `passwd` package, we just don't plumb it through to `ImageConfiguration` yet.

Turns out some software (jenkins.......................................) parses `/etc/passwd` instead of respecting common env var overrides (`$HOME`).


found-it:
image

imjasonh (approved):

PR #1182, opened 2024-06-28
auth: refactor into Authenticator interface

imjasonh:
Opening as draft to discuss.

This refactors our current auth handling -- reading `HTTP_AUTH` env var once at CLI startup, and passing through build options -- with a more cred-helper-centric form, with `EnvAuth` mimicking the current behavior, and `CGRAuth` providing an implementation that relies on `chainctl auth token`.

Go API consumers can choose to implement other auth helpers if they want, or change the behavior of the default if they want.

This should be backward-compatible with apko CLI, but breaks the Go API in ways that Melange and TF-Apko will have to adapt to.

### DIFF
--- a/pkg/apk/auth/auth.go
+++ b/pkg/apk/auth/auth.go
@@ -1,0 +1,77 @@
+package auth
+
+import (
+	"context"
+	"net/http"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/chainguard-dev/clog"
+)
+
+// DefaultAuthenciators is a list of authenticators that are used by default.
+var DefaultAuthenciators = multiAuthenticator{EnvAuth{}, CGRAuth{}}
+
+// Authenticator is an interface for types that can add HTTP basic auth to a
+// request.
+type Authenticator interface {
+	AddAuth(ctx context.Context, req *http.Request)
+}
+
+// MultiAuthenticator returns an Authenticator that tries each of the given
+// authenticators in order until one of them adds auth to the request.
+func MultiAuthenticator(auths ...Authenticator) Authenticator { return multiAuthenticator(auths) }
+
+type multiAuthenticator []Authenticator
+
+func (m multiAuthenticator) AddAuth(ctx context.Context, req *http.Request) {
+	for _, a := range m {
+		if _, _, ok := req.BasicAuth(); ok {
+			// The request has auth, so we can stop here.
+			return
+		}
+		a.AddAuth(ctx, req)
+	}
+}
+
+// EnvAuth adds HTTP basic auth to the request if the request URL matches the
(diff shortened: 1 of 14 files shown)


pkg/apk/auth/auth.go
```diff
@@ -0,0 +1,78 @@
...
+// apk.cgr.dev and the chainctl command is available.
+type CGRAuth struct{}
+
+func (c CGRAuth) AddAuth(ctx context.Context, req *http.Request) {
+	log := clog.FromContext(ctx)
+
+	host := req.Host
+	if host != "apk.cgr.dev" {
+		return
+	}
```
mattmoor, line 63: I'd love something a bit less hardcoded here.  We'll want that for staging, as well as dev environments where we will want to be able to leverage this with things like service principals.

pkg/apk/auth/auth.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// TODO(jason): Don't call this more than once per minute.
+	cmd := exec.CommandContext(ctx, "chainctl", "auth", "token", "--audience", host)
+	out, err := cmd.Output()
+	if err != nil {
+		log.Warnf("Error running `chainctl auth token`: %v", err)
+		return
+	}
+	req.SetBasicAuth("user", string(out))
```
mattmoor, line 77: I'd hope we have everything we need to do this via a library call in `chainguard.dev/sdk`

mattmoor: That's also make environmental concerns (above) simpler

imjasonh: Yep, that's the goal, this is just a lazy first step.

mattmoor: We also need to work through how this works with `--headless` style auth :-/

pkg/apk/auth/auth.go
imjasonh: wdyt about an `APKO_APK_HOST` env or something?

imjasonh: Could also run `chainctl config view` (or sdk equivalent) and get the APK host from that.

pkg/apk/auth/auth.go
imjasonh: Soon, `chainctl` will remember whether you dropped into headless auth and reuse headless mode the next time you ask for a token, so this should just work. Soon.

pkg/apk/auth/auth.go
imjasonh: I'm going to go with `APKO_APK_HOST` for now, since `chainctl config view` doesn't include the apk host for the env.

jonjohnsonjr (approved):

Issue #806, opened 2023-07-15
Question: Possible to build an OCI image from a base?


ChrisJBurns:
Currently, with Dockerfiles you can just use a base image and add more stuff that you want.

Example:
```dockerfile
FROM cgr.dev/chainguard/wolfi-base # can be any image

apk add custom-package
```

Is this possible via `apko`? Are we able to use images already as a base and just extend from them? Or is it required to build the images from scratch? Of course, this applied to any image.

jonjohnsonjr:
This isn't currently supported, but it wouldn't be impossible to implement... just a matter of priorities 😄

ChrisJBurns:
@jonjohnsonjr Well if I'm being completely honest, my issue was 100% a shot in the dark as I had a feeling it went against the fundamental principles of what `apko` is 😄 . However if it is actually something that folks agree would be a feature they're happy apko supports then (providing it's doable by a first time contributor) I don't mind tackling it as a first case? I realllllllllyy like apko, and want as many people to use it as possible, but one of the biggest pushbacks I'm getting from people at the moment is that you have to build an image completely from scratch with existing apk's. So if we could offer some sort of `base` functionality that allows us to build on top of an existing image, that would be great. - again, providing this is in-line with the Apko design principles?! 😄

jesseops:
@ChrisJBurns I came across this and the idea that came to mind was to use a remote `include` on the desired base apko configuration. I think this is intended to provide the solution you're looking for, where the common base can be inherited by multiple images.

I tried this out locally and it worked:

Create config inheriting from the "base", eg at `extend-wolfi-base.yaml`
```yaml

include: github.com/chainguard-dev/apko/examples/wolfi-base.yaml@main

contents:
  packages:
    - curl
```

Build, import, test:
```shell
apko build extend-wolfi-base.yaml local/test-extends-wolfi-base output.tar --arch=amd64 \
  && docker load < output.tar \
  && docker run --rm local/test-extends-wolfi-base:latest-amd64 curl -Ls https://httpbin.org/html
```
The output confirms it's working:

```
ℹ️            | loading config file: extend-wolfi-base.yaml
ℹ️  x86_64    | including github.com/chainguard-dev/apko/examples/wolfi-base.yaml@main for configuration
⚠️  x86_64    | remote configurations are an experimental feature and subject to change.
ℹ️  x86_64    | Building images for 1 architectures: [amd64]
[...]

PR #1184, opened 2024-07-02
Don't mutate accounts if base image is set

sfc-gh-mhazy:
The base image support is still in early development and most of top level components in config file are [banned](https://github.com/chainguard-dev/apko/blob/057f3e285595d4482f46558fdbe61aa703a0d4c5/pkg/build/types/image_configuration.go#L121)

However, Apko by default creates empty `/etc/passwd` and `/etc/group` files. With base image, this effectively wipes out any accounts set in base. Since we don't allow setting accounts in the "top" image yet, it should be enough to just not touch the above files in top layer for now. 

Long term there should be some merging logic from base, but we need to discuss how we want to retrieve the information from base image first.

Follow up for https://github.com/chainguard-dev/apko/issues/806


jonjohnsonjr (approved):

mattmoor:
We really need to rethink the approach being employed here.  The need to ban useful aspects of the configuration lest what's in the base image gets clobbered highlights my fundamental opposition to this approach on the original design.

The approach I suggested on the original doc of relying on the locked attestations and blending them with overlays (see the `MergeInto` method I just added) has none of these limitations as it remains fully declarative.

The reason this approach doesn't work is that you are trying to make a declarative tool imperative.  You are doing SCA to recover the installed package list, and because that SCA cannot recover the full base APKO specification the overlay ends up overwriting things, thus the list of banned functionality.  Conversely, starting from the original locked configuration has none of these problems because it is starting from what the SCA is trying and failing to recover.

sfc-gh-mhazy:
Hi, thanks for the comment. I will try to clarify some aspects here:

1. The difference between overlays and the base image feature is that overlays require that you can build the base as well. This means that we would not be able to build on top of images with private packages, for example [glibc-openssl-fips](https://images.chainguard.dev/directory/image/glibc-openssl-fips/versions). 
2. We don't look into anything from the base image, instead we provide the apkindex (which I hope can be considered as declarative attestation) in the [config](https://github.com/chainguard-dev/apko/blob/fee7d7cfad9b0e3c12334af4011c741b6eab8e2b/internal/cli/testdata/image_on_top.apko.yaml#L4)
3. The initial plan was to ban everything except the basic functionality (adding packages) and iteratively work on the rest of features. 
4. I didn't want to look up users in the base image, therefore added a quick "fix" that I think is aligned with users field in config being banned  when using base image. Once we have a clear plan for providing information about the base image, we can start supporting users in "top" image as well.

PR #1186, opened 2024-07-03
Simplify the resolution logic to use `expandapk.Split`

mattmoor:
The previous logic didn't properly handle unsigned APKs.

### DIFF
--- a/internal/cli/lock.go
+++ b/internal/cli/lock.go
@@ -174,11 +174,11 @@ func LockCmd(ctx context.Context, output string, archs []types.Architecture, opt
 				Architecture: rpkg.Package.Arch,
 				Version:      rpkg.Package.Version,
 				Control: pkglock.LockPkgRangeAndChecksum{
-					Range:    fmt.Sprintf("bytes=%d-%d", rpkg.SignatureSize, rpkg.ControlSize-1),
+					Range:    fmt.Sprintf("bytes=%d-%d", rpkg.SignatureSize, rpkg.SignatureSize+rpkg.ControlSize-1),
 					Checksum: "sha1-" + base64.StdEncoding.EncodeToString(rpkg.ControlHash),
 				},
 				Data: pkglock.LockPkgRangeAndChecksum{
-					Range:    fmt.Sprintf("bytes=%d-%d", rpkg.ControlSize, rpkg.DataSize),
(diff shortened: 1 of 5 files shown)


internal/cli/testdata/apko.lock.json
```diff
@@ -39,7 +39,7 @@
           "checksum": "sha1-cs+Hlyu5sY+1mmwKPedcRWj8E24="
         },
         "data": {
-          "range": "bytes=965-2899",
+          "range": "bytes=965-3864",
```
mattmoor, line 42: @jonjohnsonjr these you use a careful sanity check.  I have no idea how things worked before

jonjohnsonjr: Neither of these values looks right to me 🤔

jonjohnsonjr: I think this needs to be using Size instead of InstalledSize. Previously it was using InstalledSize. Yours is using InstalledSize + Data Offset 🤔

mattmoor: @jonjohnsonjr I believe this was the number you said on the call now, so PTAL and let me know what you think!

internal/cli/lock.go
```diff
@@ -174,11 +174,11 @@ func LockCmd(ctx context.Context, output string, archs []types.Architecture, opt
 				Architecture: rpkg.Package.Arch,
 				Version:      rpkg.Package.Version,
 				Control: pkglock.LockPkgRangeAndChecksum{
-					Range:    fmt.Sprintf("bytes=%d-%d", rpkg.SignatureSize, rpkg.ControlSize-1),
+					Range:    fmt.Sprintf("bytes=%d-%d", rpkg.SignatureSize, rpkg.SignatureSize+rpkg.ControlSize-1),
 					Checksum: "sha1-" + base64.StdEncoding.EncodeToString(rpkg.ControlHash),
 				},
 				Data: pkglock.LockPkgRangeAndChecksum{
-					Range:    fmt.Sprintf("bytes=%d-%d", rpkg.ControlSize, rpkg.DataSize),
+					Range:    fmt.Sprintf("bytes=%d-%d", rpkg.SignatureSize+rpkg.ControlSize, rpkg.SignatureSize+rpkg.ControlSize+rpkg.DataSize),
```
jonjohnsonjr, line 181: I think this should include a `-1` because the right hand size of a range header is inclusive.

jonjohnsonjr (approved):

PR #1192, opened 2024-07-03
Allow negative requirements in apko config

jonjohnsonjr:
Currently, adding !foo to the top-level package requirements of an apko config will not work. I'm not sure if this ever worked or if this is a regression, but I'd like to use it to work around an issue with the solver.


coreydaley-cg:
Closing as a stale draft: the author has moved on and the motivating solver workaround went with him. `constrain` now handles `!` prefixes, so part of this exists on `main`; whether the top-level path tolerates `!foo` end to end was not confirmed. If someone needs top-level negation, reopen or open fresh with a test.

PR #1215, opened 2024-07-23
copy annotations to config labels

imjasonh:


imjasonh:
https://github.com/chainguard-dev/apko/actions/runs/10062235445/job/27814248597 demonstrates that this change doesn't change images that don't have annotations (`TestPublish` passes without changes to the digest).

Images that specify annotations will produce different config blobs, which change image digests, which change index digests, and SBOM blobs.

imjasonh:
https://oci.dag.dev/?blob=ttl.sh/jason@sha256:d50aedcdb6fcb8c67fa2accaecf42a4084587a7cc904edfd40eec28c4478358e&mt=application%2Fvnd.oci.image.config.v1%2Bjson&size=556&manifest=ttl.sh/jason@sha256:25cc7bb368f83214f3ef52b8c780f38b720e92a5ebedb45a22739ac2ecb6fd93 is an example nginx image with the annotations (and now, labels)

jonjohnsonjr (approved):

PR #1216, opened 2024-07-23
Implement client-side APK discovery in `apko`

mattmoor:
The protocol follows a roughly similar handshake to OIDC discovery, hitting `{repository}/apk-configuration`, extracting the `jwks_uri` and then recording each pem-encoded key as `/etc/apk/keys/{kid}.rsa.pub`.

I have tested this locally with both public and private APK repositories against `apk.mattmoor.dev` with the changes in: https://github.com/chainguard-dev/mono/pull/18431

Once this lands, I can use this in e2e tests there, but those are somewhat 🐔  & 🥚

### DIFF
--- a/pkg/apk/apk/implementation.go
+++ b/pkg/apk/apk/implementation.go
@@ -19,9 +19,11 @@ import (
 	"bytes"
 	"context"
 	"crypto/rsa"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
@@ -37,7 +39,6 @@ import (
 	"time"
 
 	"github.com/go-jose/go-jose/v4"
-	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"go.lsp.dev/uri"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -874,6 +875,9 @@ func (a *APK) fetchChainguardKeys(ctx context.Context, repository string) error 
 	}
 
 	for _, key := range jwks.Keys {
+		if key.KeyID == "" {
+			return fmt.Errorf(`key missing "kid"`)
+		}
 		filename := filepath.Join(keysDirPath, key.KeyID+".rsa.pub")
 		f, err := a.fs.OpenFile(filename, os.O_CREATE|os.O_WRONLY, 0o644)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/apk/apk/implementation.go
```diff
@@ -794,6 +814,85 @@ func (a *APK) fetchAlpineKeys(ctx context.Context, alpineVersions []string) erro
...
+
+	for _, key := range jwks.Keys {
+		filename := filepath.Join(keysDirPath, key.KeyID+".rsa.pub")
+		f, err := a.fs.OpenFile(filename, os.O_CREATE|os.O_WRONLY, 0o644)
+		if err != nil {
+			return fmt.Errorf("failed to open key file %s: %w", filename, err)
+		}
+		defer f.Close()
+
+		pubBytes, err := cryptoutils.MarshalPublicKeyToPEM(key.Key.(*rsa.PublicKey))
```
imjasonh, line 884: cryptoutils is a kitchen sink. I know we already transitively depend on it via cosign, but I wish we didn't.

Can we just implement this ourselves and save the direct dep we'll have to sever later? It looks pretty simple.

https://github.com/sigstore/sigstore/blob/v1.8.7/pkg/cryptoutils/publickey.go#L76

pkg/apk/apk/implementation.go
```diff
@@ -794,6 +814,85 @@ func (a *APK) fetchAlpineKeys(ctx context.Context, alpineVersions []string) erro
...
+		return fmt.Errorf("failed to fetch JWKS: %v", jwksResponse.Status)
+	}
+
+	jwks := jose.JSONWebKeySet{}
+	if err := json.NewDecoder(jwksResponse.Body).Decode(&jwks); err != nil {
+		return fmt.Errorf("failed to unmarshal JWKS: %w", err)
+	}
+
+	for _, key := range jwks.Keys {
+		filename := filepath.Join(keysDirPath, key.KeyID+".rsa.pub")
```
imjasonh, line 877: Can we check that the key id is non-empty?

And possibly also that the key bytes is set, I'm not sure if that'll fail otherwise.

I want to avoid the situation where a non-JWKS JSON response is silently handled without failing.

imjasonh (commented):
Looks easier than I thought? Can we get some simple tests? 🙏

pkg/apk/apk/implementation.go
mattmoor: done

pkg/apk/apk/implementation.go
mattmoor: done

mattmoor:
@imjasonh I made it do a real key discovery to confirm things work against staging.  We can swap this to a prod group once we push.

imjasonh (approved):

PR #1218, opened 2024-07-24
begin a new APK client

luhring:
The idea here is to orient remote APK interactions around a **client**, in order to give consumers the ability to handle dependency injection properly (such as by injecting an HTTP client).

And the ultimate goal is to re-use this client whenever we need to access remote indexes and packages. It'd be great if we can refine this API enough where we can reuse this in all the places we need to interact with APK repositories, to replace the many disparate implementations of this logic we have in our tools today.

I haven't gone as far as to refactor other code in apko to use this client, but I think that's a reasonable next step. So far this implementation saves me from writing it again in an internal tool.

Feedback welcome. 😃 

cc: @imjasonh

### DIFF
--- a/pkg/apk/apk/client.go
+++ b/pkg/apk/apk/client.go
@@ -1,0 +1,65 @@
+package apk
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"chainguard.dev/apko/pkg/apk/auth"
+)
+
+const (
+	WolfiAPKRepo                = "https://packages.wolfi.dev/os"
+	ChainguardEnterpriseAPKRepo = "https://packages.cgr.dev/os"
+	ChainguardExtrasAPKRepo     = "https://packages.cgr.dev/extras"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


imjasonh (commented):
I might suggest a new package, so we can more clearly tell what still depends on the "plumbing" stuff in the-code-formerly-known-as-go-apk.

`pkg/apk/client` maybe? `client.New` is pretty overloaded, but hey, what's one more.

pkg/apk/apk/client.go
```diff
@@ -0,0 +1,65 @@
...
+	"chainguard.dev/apko/pkg/apk/auth"
+)
+
+const (
+	WolfiAPKRepo                = "https://packages.wolfi.dev/os"
+	ChainguardEnterpriseAPKRepo = "https://packages.cgr.dev/os"
+	ChainguardExtrasAPKRepo     = "https://packages.cgr.dev/extras"
+)
+
+const (
```
jdolitsky, line 18:
```suggestion

```

imjasonh (approved):

PR #1227, opened 2024-07-29
new command: `install-keys`

imjasonh:
```
$ apko install-keys
```

This discovers keys for all repositories configured in `/etc/apk/repositories`, and writes the public keys to `/etc/apk/keys/<ID>.rsa.pub`.


mattmoor:
This is so weird.  I just wrote myself a note to create a tiny CLI that does the install but based on discovery over /etc/apk/repositories.

imjasonh:
> This is so weird. I just wrote myself a note to create a tiny CLI that does the install but based on discovery over /etc/apk/repositories.

Oh that's a good idea! I can make this do that instead.

imjasonh:
Okay this is now `apko install-keys` and _only_ discovers and installs keys for all configured repositories.

mattmoor (approved):

PR #1256, opened 2024-08-10
rsa256

xnox:
- **test: Add test for RSA256 signed APKINDEX**
  This test case should fail, as apko has no support for RSA256
  signatures.
  
  And yet it passes.... as if the rsa256-signed/APKINDEX.tar.gz is not
  fetched nor verified.
  
  This in preparation to add support for RSA256 signed APK repositories.
  

- **apk,signature: add support to verify RSA256 signed Index**
  This adds support to correctly parse signature stream that contains
  multiple types of signatures. This also refactors signing to allow
  trivially changing signature type, and to have multiple if so desired.
  
  See code comments for lots of caveats.
  
  Once all tools upgrade to this version of the module, signing can be
  changed from RSA to RSA256.
  

Note Alpine apk-tools v2 has support for such signatures since 2017, Alpine abuild-sign support has been merged into mainline in june see https://gitlab.alpinelinux.org/alpine/abuild/-/merge_requests/290 

Note whilst apk-tools has had support for this signatures forever, it is not in active use, add a package check to apk-tools to be sure that RSA256 is double checked to be operational https://github.com/wolfi-dev/os/pull/27473


pkg/apk/signature/rsa.go
```diff
@@ -90,10 +92,15 @@ func RSAVerifySHA1Digest(sha1Digest, signature []byte, publicKey []byte) error {
...
 		return fmt.Errorf("verify PKCS1v15 signature: %w", err)
 	}
 
 	return nil
 }
+
+// Backwards compat
+func RSASignSHA1Digest(digest []byte, keyFile, passphrase string) ([]byte, error) {
+	return RSASignDigest(digest, crypto.SHA1, keyFile, passphrase)
+}
```
xnox, line 106: Need a VerifySHA1Digest compat function too.

jonjohnsonjr (approved):

PR #1275, opened 2024-08-18
Add LoongArch architecture definition

kaniini:
LoongArch is a newish ISA that was introduced by Loongson, who previously made MIPS64r2 processors.

LoongArch is essentially a fork of MIPS64r5 with some RISC-V influences added alongside Loongson's previous proprietary MIPS extensions.

The apk architecture name for this architecture is 'loongarch64', while the OCI architecture name is 'loong64'.

This change allows apko to build images for the Alpine loongarch64 port.

### DIFF
--- a/pkg/build/types/schema.json
+++ b/pkg/build/types/schema.json
@@ -93,7 +93,7 @@
             "type": "string"
           },
           "type": "array",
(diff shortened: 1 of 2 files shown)


kaniini:
bump ?

k4leung4 (approved):

PR #1277, opened 2024-08-18
fix(ci): mark GitHub releases as latest from prerelease

### LINKED ISSUE #639: Question: why do we create prereleases for apko?
As I saw from the [configuration in GoReleaser](https://github.com/chainguard-dev/apko/blob/889f9c1ed2f3e94736dafc58a3f42c26f4eaf659/.goreleaser.yaml#L60), apko releases marked as pre-release, is there any special reason of doing this? Thanks

/cc @cpanato

### LINKED ISSUE #1276: Mark GitHub releases as latest
Please could you mark releases as "latest" (i.e. non pre-release) since this is messing with some automation in various repositories. 
It's generally bad practice to use pre-release binaries in production, so therefore a lot of tooling is built to exclude these packages from automated updates and for selection during builds.

I know next to nothing about goreleaser, however I can see `prerelease: true` being set.
https://github.com/chainguard-dev/apko/blob/main/.goreleaser.yaml#L61

From reading older (closed) issues, there doesn't appear to be a real reason for these to be marked as pre-release, but please do let me know if there's a reason.

p5:
Closes #1276 
Closes #639 

By using pre-releases, new updates are not discoverable by users and tooling.  GitHub does not promote these as much, and tools like Renovate and Dependabot will exclude all pre-release packages by default.

This PR changes the Goreleaser configuration to no longer create pre-release packages, and instead create proper releases which will be available at the https://github.com/chainguard-dev/apko/releases/latest URL.

While I do not know anything about goreleaser, there is a `prerelease: true` line which does look to be the cause for this.

This appears to have been set when GoReleaser was [first introduced to this repository over two years ago](https://github.com/chainguard-dev/apko/commit/50a81ccf1e901ea54eb5326aa63a0f1a90aa0405), and I assume may have been something that has been overlooked.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -58,5 +58,5 @@ changelog:
 
 release:
   draft: false
-  prerelease: true
+  prerelease: false
   name_template: "Release {{ .Tag }}"


p5:
Do I need to setup a PGP key in order to commit to this repository?
I am already signing commits using my SSH key.

xnox:
a new release got cut and marked as latest; and there is no need to default to marking things as latest, as we can always edit a release after all the artifacts are built.

xnox:
and we keep forgetting to mark releases as released.

xnox (approved):

p5:
Thanks!

PR #1305, opened 2024-09-14
Keep apk modtime

jonjohnsonjr:
We were overwriting all the file timestamps with a configurable SOURCE_DATE_EPOCH, which was always the unix epoch. The only reason we were doing this is that we were also setting the modtime of everything to time.Now() elsewhere.

That's silly.

Instead, don't set anything to time.Now() and keep file modtimes when we add them to the filesystem. If the APKs change, nothing will be reproducible anyway, so keeping the data from the APK seems way better.

### DIFF
--- a/pkg/apk/tarball/write.go
+++ b/pkg/apk/tarball/write.go
@@ -149,9 +149,7 @@ func (c *Context) writeTar(ctx context.Context, tw *tar.Writer, fsys fs.FS, user
 		header.Name = path
 
 		// zero out timestamps for reproducibility
-		header.AccessTime = c.SourceDateEpoch
-		header.ModTime = c.SourceDateEpoch
-		header.ChangeTime = c.SourceDateEpoch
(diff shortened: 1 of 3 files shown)


pkg/apk/tarball/write.go
```diff
@@ -149,9 +149,7 @@ func (c *Context) writeTar(ctx context.Context, tw *tar.Writer, fsys fs.FS, user
 		header.Name = path
 
 		// zero out timestamps for reproducibility
-		header.AccessTime = c.SourceDateEpoch
-		header.ModTime = c.SourceDateEpoch
-		header.ChangeTime = c.SourceDateEpoch
+		header.ModTime = info.ModTime()
```
mattmoor, line 152: Update the comment?

mattmoor (approved):
nit about a comment, but otherwise this LGTM.

PR #1310, opened 2024-09-19
Restore SourceDateEpoch in tarball for melange

jonjohnsonjr:
We don't use this (and we don't want it) but melange uses the tarball package, so let's avoid breaking that for now until we can migrate stuff off of this tangled web of weird fs.FS implementations.


k4leung4 (approved):

k4leung4:
does it need unit tests?

PR #1317, opened 2024-09-23
Cache some more expensive operations

jonjohnsonjr:
Both of these caches are keyed off of a ~tree of the actual pointer values of indices. This makes it cheap to compare for cache hits, but also requires shuffling around a couple things to ensure that the APKINDEX parsing code returns the same pointer on every call and that we don't re-generate the []RepositoryPackage list when we call Packages().

These are both minor refactorings that help perf a little bit, but those enable us to memoize two expensive things:

The first is NewPkgResolver(), which takes a few hundred milliseconds for wolfi, which is negligible individually but in aggregate for thousands of images takes up a lot of time.

The second is the code for disqualifying packages that don't exist in every architecture. Similarly, this was relatively cheap for a single image build but significant in aggregate.

Since for both of these they (almost always) have the exact same inputs, memoizing these shows a substantial speedup. In my local testing, we're spending ~25% of our time in "terraform plan" doing these two things.


imjasonh (approved):

pkg/apk/apk/shameful_global_caches.go
joshrwolf, line 1: this is brilliant, if you put it in a separate file nobody will ever find it!

joshrwolf (approved):

PR #1321, opened 2024-09-26
Fail if APKINDEX has single-character lines

jonjohnsonjr:
We expect some blank lines, which delimit each entry. We expect every other line to have at least e.g. "C:", i.e. a single character followed by a colon.

I looked at our current APKINDEX to make sure I'm not missing anything:

```
$ awk '{print length}' APKINDEX | sort -n | uniq -c | head
83376 0
11137 2
7807 3
1214 4
18924 5
23738 6
49958 7
126520 8
62586 9
86165 10
```

We don't have any single-character lines, so this change should be fine.


cpanato (approved):
nice

PR #1327, opened 2024-10-02
Make etag checks optional

jonjohnsonjr:
This isn't perfect but I think it moves us closer in the right direction. Previously, we had a process-wide global etag cache that would prevent sending multiple HEAD requests to check the etag value of any URLs we fetched (keys and APKINDEX, mostly).

This is great for a local tool or even the terraform provider, but it's awful as a library used by a service because the lifetime of the process is liable to outlive the validity of the cach.

This hoists that global etag caching mechanism out into a little apk.Cache struct, which is where I'm hoping to put even more things that I'm ashamed of going forward.

Using the same apk.NewCache(true) instance across multiple builds will skip HEAD requests to the same URL and just assume that the tag hasn't changed. This was the previous default behavior.

Using the same apk.NewCache(false) instance across multiple builds (the default) will send a HEAD request prior to every key or index fetch in order to see if they've changed.

We still have some global caching going on for actually parsing the results, which keys off of etag values, so we shouldn't have any progressions re: the recent CPU savings changes I've made.

Another thing this introduces (because we dropped the more powerful etag-based caching) is a some singleflight instances for these outgoing requests. We don't want to cache the results indefinitely, but we do want to avoid sending redundant requests in every case, so apk.NewCache(false) will still coalesce multiple concurrent requests for the same URL, so there's still benefit to using this cache object even with the etag caching disabled.

One thing that's missing here is a distinction between ~request-scoped caching if e.g. I want to use an etag cache for any outgoing URLs for a single build, but I also want to use a global cache for coalescing requests. I'll probably split the singleflight stuff off from this struct at some point, but I want to measure first to see if that's really necessary.

### DIFF
--- a/pkg/apk/apk/index.go
+++ b/pkg/apk/apk/index.go
@@ -31,6 +31,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/chainguard-dev/clog"
 	"github.com/klauspost/compress/gzip"
 	"go.lsp.dev/uri"
 	"go.opentelemetry.io/otel"
@@ -47,7 +48,8 @@ var signatureFileRegex = regexp.MustCompile(`^\.SIGN\.(DSA|RSA|RSA256|RSA512)\.(
 // We just hold the parsed index in memory rather than re-parsing it every time,
 // which requires gunzipping, which is (somewhat) expensive.
 var globalIndexCache = &indexCache{
-	modtimes: map[string]time.Time{},
+	modtimes:  map[string]time.Time{},
+	urlToEtag: map[string]string{},
 }
 
 type indexResult struct {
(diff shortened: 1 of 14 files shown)


pkg/apk/apk/index.go
```diff
@@ -408,3 +452,19 @@ func WithIndexAuthenticator(a auth.Authenticator) IndexOption {
...
+	asURL, err := url.Parse(in)
+	if err != nil {
+		// Attempt to parse non-https elements into URI's so they are translated into
+		// file:// URLs allowing them to parse into a url.URL{}
+		asURL, err := url.Parse(string(uri.New(in)))
+		if err != nil {
+			return in
+		}
+
+		return asURL.Redacted()
```
imjasonh, line 466: Do we ever need to redact file paths? 🤔

imjasonh (approved):

pkg/apk/apk/index.go
jonjohnsonjr: I kinda gave up on this one and went with "keep the same behavior but make it a little less janky to read" and this let me just handle them all the same.

PR #1358, opened 2024-10-23
fix bug with triggers encoded in triggers file

imjasonh:


xnox (changes requested):
There is a test failure, and I am not sure the test case is valid.

Where is the test case definition and can we verify it against apk-tools created chroots?

xnox (approved):

xnox (changes requested):
let me do a local interractive test of rebuilt apko, and rebuilt melange and rebuilt everything.

xnox:
also to check oci terraform provider; and also to check that all existing triggers in wolfi are non-failing, as they will start to be executed for the first time.

xnox (approved):
Verified this fixes lots of bugs, everything works, and all existing scriplets are sane.

PR #1366, opened 2024-10-24
fix(sbom): deduplicate SBOM packages by ID

luhring:
In https://github.com/chainguard-dev/melange/pull/1474, we updated our SBOMs to have distinct SPDX packages for the APK itself, vs. the build config used to produce it, vs. the upstream source(s) pulled in for the build. This helps us produce more idiomatic and descriptive SPDX data.

But this caused a problem downstream in apko — when we aggregate APKs' SBOMs, and specifically their SPDX packages, into a single list of SPDX packages for the image, we end up with duplicate package IDs when multiple APKs were build from the same build config or the same upstream source.

This PR adds tests to catch this case, and it fixes the duplicate package issue by dropping subsequent instances of a given package ID.

It also further improves on the existing SBOM-related tests.


pkg/sbom/generator/spdx/spdx.go
```diff
@@ -137,6 +138,18 @@ func (sx *SPDX) Generate(opts *options.Options, path string) error {
 		}
 	}
 
+	dedupedPackages := make([]Package, 0, len(doc.Packages))
+	seenIDs := make(map[string]struct{})
```
jdolitsky, line 142: thank you

jdolitsky (approved):

PR #1371, opened 2024-10-25
remove support for docker-typed manifests

imjasonh:
We don't use anything but the OCI types, and supporting both adds complexity unnecessarily.

Draft until we've fixed the other blockers, so this doesn't cause more.


coreydaley-cg:
Closing as a stale draft: the author has moved on, the branch no longer applies, and it was self-blocked pending other work that never resumed. Dropping Docker-typed manifests is still a reasonable simplification if nobody depends on them, but that is a product call before it is a code change. Reopen if someone adopts it.

PR #1374, opened 2024-10-25
Make MergeInto threadsafe

jonjohnsonjr:
Just appending to maps or slices is not great in a concurrent universe, so my previous fix was in fact not fixing much of anything, whoops.


luhring (approved):

jonjohnsonjr:
Our tests against alpine are all failing because currently they only have riscv64 packages in edge/main.

PR #1404, opened 2024-11-14
Pass errors up the stack in CalculateWorld and InstallPackages

smoser:
This problem was found in melange
 https://github.com/chainguard-dev/customer-issues/issues/1822

Any time a download failed, we did not communicate the error up the stack.


luhring:
@jonjohnsonjr This looks up your alley 👀

pkg/apk/apk/implementation.go
```diff
@@ -524,9 +524,9 @@ func (a *APK) CalculateWorld(ctx context.Context, allpkgs []*RepositoryPackage)
 	resolved := make([]*APKResolved, len(allpkgs))
 
 	// A slice of pseudo-promises that get closed when expanded[i] is ready.
-	done := make([]chan struct{}, len(allpkgs))
+	done := make([]chan error, len(allpkgs))
```
jonjohnsonjr, line 527: We can actually just delete these because there's nothing that reads from these channels. I think this was a copy-paste leftover.

pkg/apk/apk/implementation.go
```diff
@@ -700,13 +707,13 @@ func (a *APK) InstallPackages(ctx context.Context, sourceDateEpoch *time.Time, a
 		g.Go(func() error {
 			exp, err := a.expandPackage(ctx, pkg)
 			if err != nil {
-				return fmt.Errorf("expanding %s: %w", pkg, err)
+				err = fmt.Errorf("expanding %s: %w", pkg, err)
```
jonjohnsonjr, line 710: ]

jonjohnsonjr, line 710: I would do something like:

```go
g.Go(func() (rerr error) {
  defer func() { done[i] <- rerr }()

  exp, err := a.expandPackage(ctx, pkg)
  if err != nil {
    return fmt.Errorf("expanding %s: %w", pkg, err)
  }

  expanded[i] = exp

  return nil
})
```

That creates a named return (yucky but useful for this error handling stuff) and closes `done[i]` with the returned error (which is nil on success).

jonjohnsonjr: I might not love this approach actually because now we need buffered channels instead of unbuffered if we're going to send 🤔

jonjohnsonjr (approved):

PR #1406, opened 2024-11-15
Consider already selected packages during solve

jonjohnsonjr:
We ran into a very very slow solve. I can't tell if it was actually slow
    or infinite, but it didn't seem to be convering very quickly. I realized
    that we essentially recurse the entire graph for every constraint, even
    for things we've already solved. This was surprising, because I thought
    we handled that already, but it seems like we didn't.
    
    I've added an example "abseil-regression.yaml" that took over a second
    to solve (as of this writing) before this change and takes about 1ms
    after this change. There's an open PR in wolfi that timed out because
    the abseil dep graph got even more gnarly and took several hours, so
    there was something (at least) quadratic going on here, oops.
    
    I've deleted a test that was exercising a scenario where packages were
    disallowed from providing themselves because there is no code path where
    that was actually possible (all invocations had "true" hard-coded).
    
    I also ran this against a several thousand image configs and they all
    produced the same results before and after this change, so I am fairly
    confident in it.

### DIFF
--- a/examples/abseil-regression.yaml
+++ b/examples/abseil-regression.yaml
@@ -1,0 +1,13 @@
+# This was very slow once.
+contents:
+  keyring:
+    - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+  repositories:
+    - https://packages.wolfi.dev/os
+
+  packages:
+    - abseil-cpp-dev
+    - pkgconf
+
+archs:
+- arm64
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
e2e tests are failing because a lot of our examples use alpine repos, and currently `musl` on x86_64 is at `1.2.5-r6` whereas `musl` on aarch64 is at `1.2.5-r5`

imjasonh (approved):

dannf:
I've tested this branch and it fixes the issue for me. Thanks! Do you mind referencing https://github.com/chainguard-dev/melange/issues/1651 in the commit log?

PR #1407, opened 2024-11-18
fix: Allow accounts to belong to GID 0

EyeCantCU:
Currently, we block new users from belonging to GID 0. While this is a great practice, some images may expect that the configured user belong to GID 0 so remove the limitation


imjasonh (approved):

imjasonh:
The change itself LGTM, the test errors seem to be due to Alpine dependencies. I'll retry some and see if we can get these passing.

PR #1415, opened 2024-11-22
Disqualify conflicts with self for packages

jonjohnsonjr:
Previously, we called disqualifyConflicts to disqualify any packages that would conflict with whatever package we selected. This made it possible to pick multiple packages with the same name, which we'd then deduplicate and end up with whatever the first package was. That's unexpected and produced solutions that were missing a package in a few corner cases, mostly caused by old packages (in wolfi) providing things accidentally that weren't withdrawn.

I've manually tested this change against our large corpus of images and nothing failed to solve.

### DIFF
--- a/pkg/apk/apk/repo.go
+++ b/pkg/apk/apk/repo.go
@@ -355,31 +355,37 @@ func (p *PkgResolver) conflictingVersion(constraint parsedConstraint, conflict *
 }
 
 // Disqualify anything that conflicts with the given pkg.
-func (p *PkgResolver) disqualifyConflicts(pkg *RepositoryPackage, dq map[*RepositoryPackage]string) {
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

coreydaley-cg:
Closing as stale: the author and approver have both moved on and the branch no longer applies. Recording the state so this is findable later. The silent-dedup bug this fixed can't happen on `main` anymore because `pick()` now rejects a same-name re-selection outright; what remains is a hard solve error, "selecting package X conflicts with Y on name," and no open issue reports it. If that error shows up with current data, the fix here, disqualifying same-name candidates so the solver never reaches the collision, is still the right shape and should fold into whatever resolver work is active at the time rather than reviving this branch.

PR #1427, opened 2024-12-11
Ignore scripts that aren't executable

jonjohnsonjr:
This should probably be an allowlist like apk does, but this is a smaller diff and I'm worried about breaking more things.

### DIFF
--- a/pkg/apk/apk/installed.go
+++ b/pkg/apk/apk/installed.go
@@ -154,6 +154,13 @@ func (a *APK) updateScriptsTar(pkg *Package, controlTarGz io.Reader, sourceDateE
 			continue
 		}
 
+		// Ignore files that aren't executable.
+		// This is mostly to ignore .melange.yaml files in the control section,
+		// but apk itself has hardcoded list of scripts that we might want to do too.
+		if header.FileInfo().Mode().Perm()&0555 != 0555 {
+			continue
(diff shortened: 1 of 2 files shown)


pkg/apk/apk/installed.go
```diff
@@ -154,6 +154,13 @@ func (a *APK) updateScriptsTar(pkg *Package, controlTarGz io.Reader, sourceDateE
 			continue
 		}
 
+		// Ignore files that aren't executable.
+		// This is mostly to ignore .melange.yaml files in the control section,
+		// but apk itself has hardcoded list of scripts that we might want to do too.
+		if header.FileInfo().Mode().Perm()&0555 != 0555 {
```
luhring, line 160: It's not immediately obvious to me from this why we're checking for these exact perm bits (r+x for all 3) but maybe that's just because I don't know this part of the domain

luhring (approved):

pkg/apk/apk/installed.go
jonjohnsonjr: FWIW I copied this from melange SCA that does similar checks for entirely different reasons, and it matched what I saw from sampling real-world scripts in alpine. If this ends up being brittle I'll come back and figure out exactly what `apk` does here and replicate it.

PR #1443, opened 2024-12-17
Add an explicit sort to squash diffs.

mattmoor:
This eliminates a subtle source of locked config diffs depending on whether we sort with or without the version string present.

### DIFF
--- a/pkg/build/lock.go
+++ b/pkg/build/lock.go
@@ -237,6 +237,10 @@ func unify(originals []string, inputs []resolved) (map[string][]string, map[stri
 	for _, pkg := range sets.List(acc.packages) {
 		pl = append(pl, fmt.Sprintf("%s=%s", pkg, acc.versions[pkg]))
 	}
+	// Sort the package list explicitly with the `=` included.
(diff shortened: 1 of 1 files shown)


mattmoor:
Here's an example of a sort that introduces very subtle differences: https://github.com/chainguard-dev/terraform-provider-apko/blob/07717440b3d45ba8c4408ec5f9ca0ecdc0b1376b/internal/provider/build.go#L43

jonjohnsonjr (approved):

vaikas:
Are there no tests for this method to begin with, or am I just missing them? Seems like this method is pretty large and sure would be nice to have tests for it.

mattmoor:
@vaikas we can add a test here.  It's arguably one of the best tested parts of apko 🙃 

https://github.com/chainguard-dev/apko/blob/00e43dd1e22ed5b78effe90f6a2e9d949787f5a1/pkg/build/lock_test.go#L25

vaikas:
@mattmoor knew I was missing them. The pkg/build threw me and I was looking at the pkg/build/build_test.go file incorrectly , blah... Thanks!

PR #1457, opened 2024-12-27
Fixing the url escape for name of the package URL

leoncamel:
This will fix the url-encoding for package URL, since for my object-storage(S3 protocol), it need safely escape URL for all path. And escaping the filename part for the URL is a general way for apk package fetching.

### DIFF
--- a/pkg/apk/apk/repository.go
+++ b/pkg/apk/apk/repository.go
@@ -2,6 +2,7 @@ package apk
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


leoncamel:
@imjasonh @jonjohnsonjr Can you give a quick review on this small patch? Thanks.

imjasonh (approved):

leoncamel:
@imjasonh Hi, Is there anything else I can do for this PR?

I saw it is approved, but this PR seems blocked here.

coreydaley-cg (changes requested):
This breaks local repositories: `fetchPackage` opens local packages with `os.Open(pkg.URL())`, so after this change a local `libstdc++-16.2.0-r1.apk` is looked up as `libstdc%2B%2B-16.2.0-r1.apk` and isn't found. Any local-repo build that installs `libstdc++` fails. The escape is right for the HTTP case, so it should happen where the request is constructed (the `http`/`https` branch in `fetchPackage`, and any other HTTP consumer of `URL()`), leaving `URL()` itself as the raw path. A test with a local repo containing a `+` in a filename would pin it.

PR #1449, opened 2024-12-19
Make GID its own type to prevent defaulting to 0

jdolitsky:
A follow up to https://github.com/chainguard-dev/apko/pull/1407 which had the unintended side effect of missing gid value defaulting to 0

PR #1466, opened 2025-01-07
Fix logging output of user.GID

jdolitsky:
Follow up to comment here: https://github.com/chainguard-dev/apko/pull/1449#issuecomment-2568110052

cc @pnasrat


cpanato (approved):

pnasrat (approved):
LGTM

PR #1468, opened 2025-01-07
Consider pinned repos when locking packages

jonjohnsonjr:


coreydaley-cg:
#1472 landed the pinned-suffix handling in `unify` a week after this opened, so this is already on `main`. Closing as superseded.

PR #1472, opened 2025-01-09
Fix package name handling to retain version and strip ‘@’ suffix

### LINKED ISSUE #1471: unable to lock packages to a consistent version: map[myapp@local:[ (amd64)]]
It was initially being discussed [here](https://github.com/chainguard-dev/apko/issues/1448) but with the new change the error message has changed and the build for local packages is broken.

Please see the error. Ref: https://github.com/chainguard-dev/apko/issues/1448#issuecomment-2575880302
```bash
...
2025/01/07 17:31:51 INFO deleting guest dir /tmp/melange-guest-633674898
2025/01/07 17:31:51 INFO deleting workspace dir /tmp/melange-workspace-2216291366
2025/01/07 17:31:51 INFO removing image path /tmp/melange-guest-1211196036
Building APKO ...
2025/01/07 17:31:53 INFO Building images for 1 architectures: [amd64]
2025/01/07 17:31:54 [DEBUG] GET https://dl-cdn.alpinelinux.org/alpine/edge/community/apk-configuration
2025/01/07 17:31:55 [DEBUG] GET https://dl-cdn.alpinelinux.org/alpine/edge/main/apk-configuration
2025/01/07 17:31:56 INFO setting apk repositories: [@local ./packages https://dl-cdn.alpinelinux.org/alpine/edge/community https://dl-cdn.alpinelinux.org/alpine/edge/main]
Error: locking config: unable to lock packages to a consistent version: map[myapp@local:[ (amd64)]]
2025/01/07 17:31:56 INFO error during command execution: locking config: unable to lock packages to a consistent version: map[myapp@local:[ (amd64)]]
...
```

/cc @jackwhelpton @jonjohnsonjr @Cloudnapsys

eminaktas:
Fix #1471

Dear @jonjohnsonjr,

Can you please review my change? I noticed your PR [here](https://github.com/chainguard-dev/apko/pull/1468). I was also looking in the issue and this change worked.

### DIFF
--- a/pkg/build/lock.go
+++ b/pkg/build/lock.go
@@ -109,6 +109,7 @@ type resolved struct {
 	arch     string
 	packages sets.Set[string]
 	versions map[string]string
+	pinned   map[string]string
 	provided map[string]sets.Set[string]
 }
 
@@ -124,21 +125,37 @@ func unify(originals []string, inputs []resolved) (map[string][]string, map[stri
 	originalPackages := resolved{
 		packages: make(sets.Set[string], len(originals)),
 		versions: make(map[string]string, len(originals)),
+		pinned:   make(map[string]string, len(originals)),
 	}
 
 	byArch := map[string][]string{}
 
 	for _, orig := range originals {
 		name := orig
+		version := ""
+		pinned := ""
+
 		// The function we want from go-apk is private, but these are all the
 		// special characters that delimit the package name from the cosntraint
 		// so lop off the package name and stick the rest of the constraint into
 		// the versions map.
 		if idx := strings.IndexAny(orig, "=<>~"); idx >= 0 {
 			name = orig[:idx]
+			version = orig[idx:]
 		}
+
+		// Extract pinned version if present
+		if idx := strings.IndexAny(orig, "@"); idx >= 0 {
+			pinned = orig[idx:]
+		}
+
+		// Remove pinned suffix from name and version
+		name = strings.TrimSuffix(name, pinned)
+		version = strings.TrimSuffix(version, pinned)
+
 		originalPackages.packages.Insert(name)
-		originalPackages.versions[name] = strings.TrimPrefix(orig, name)
+		originalPackages.versions[name] = version
+		originalPackages.pinned[name] = pinned
 	}
 
 	// Start accumulating using the first entry, and unify it with the other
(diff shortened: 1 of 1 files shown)


jonjohnsonjr (approved):

jonjohnsonjr:
Closed and re-opened to try to retrigger CI checks, didn't seem to work 🤔

pkg/build/lock.go
```diff
@@ -226,7 +243,11 @@ func unify(originals []string, inputs []resolved) (map[string][]string, map[stri
 	// package constraint including the operator.
 	for _, pkg := range sets.List(missing) {
 		if ver := originalPackages.versions[pkg]; ver != "" {
-			pl = append(pl, fmt.Sprintf("%s%s", pkg, ver))
+			if pin := originalPackages.versions[pkg]; pin != "" {
+				pl = append(pl, fmt.Sprintf("%s%s%s", pkg, ver, pin))
```
emilylange, line 247: This produces an invalid `world` when using tagged repositories like `@local` or `@edge`.

The order needs to be 
```diff
- pl = append(pl, fmt.Sprintf("%s%s%s", pkg, ver, pin))
+ pl = append(pl, fmt.Sprintf("%s%s%s", pkg, pin, ver))
```

as per [`apk-world(5)`](https://man.archlinux.org/man/apk-world.5.en):

> This is a plaintext file with one constraint using dependency notation per line. Each line has the format: `name{@tag}{[<>~=]version}`.

eminaktas: Hi @emilylange

When I tested it with `@local` package, the format you shared didn't work. This is the reason I kept that like that. However, changing the order doesn't break the local package build since it is not the case for local built packages.

With `@local` package repository below example config does not work.

```bash
contents:
  repositories:
    - '@local ./packages'
  keyring:
    - melange.rsa.pub
  packages:
    - world@local=7.0.0-r0
```

This is the error message. This error happens before the changes.

```bash
2025/01/24 19:25:15 DEBU got 1 indexes:
./packages/x86_64/APKINDEX.tar.gz
Error: locking config: resolving apk packages: for arch "amd64": solving "world@local=7.0.0-r0" constraint: nothing provides "world@local=7.0.0-r0"
2025/01/24 19:25:15 INFO error during command execution: locking config: resolving apk packages: for arch "amd64": solving "world@local=7.0.0-r0" constraint: nothing provides "world@local=7.0.0-r0"
```

However, this config works but it doesn't match with the format.

```bash
contents:
  repositories:
    - '@local ./packages'
  keyring:
    - melange.rsa.pub
  packages:
    - world=7.0.0-r0@local
```

Can you share the failing config?

eminaktas: @emilylange Thanks for the detailed explanation! I think this needs its own ticket and maybe even a discussion in the [#apko](https://kubernetes.slack.com/archives/C03MASRQP4M) Slack channel. Since this is a merged ticket, there’s a good chance the right people might not see it. Would be great to get more visibility on this.

PR #1488, opened 2025-01-21
Guard against os-release panic

jonjohnsonjr:


pkg/build/sbom.go
```diff
@@ -171,7 +172,18 @@ func readReleaseData(fsys fs.FS) (*osr.Data, error) {
...
 	}
 
-	return osr.Parse(string(osReleaseData)), nil
+	releaseStr := string(osReleaseData)
+
+	// osr.Parse can panic if given improper input, so error early instead.
+	for _, line := range strings.Split(releaseStr, "\n") {
+		if line != "" {
+			if len(strings.Split(line, "=")) < 2 {
+				return nil, fmt.Errorf("invalid os-release line: %q", line)
```
xnox, line 181: The format of this file is described at https://www.freedesktop.org/software/systemd/man/latest/os-release.html

> Lines beginning with "#" are treated as comments. Blank lines are permitted and ignored.

does above handle this requirement?

wlynch (approved):

pkg/build/sbom.go
xnox: Depending on needs it can be a broken symlink to unmounted /usr partition too.

> The file /etc/os-release takes precedence over /usr/lib/os-release. Applications should check for the former, and exclusively use its data if it exists, and only fall back to /usr/lib/os-release if that is missing.

wlynch: No - this seems like a likely culprit though. It might be worth making our own parser?

jonjohnsonjr: That's interesting. The `osr.Parse` code doesn't seem to handle comments, that's probably our root cause.

pkg/build/sbom.go
```diff
@@ -166,12 +173,38 @@ func readReleaseData(fsys fs.FS) (*osr.Data, error) {
...
+	scanner := bufio.NewScanner(f)
+
+	kv := map[string]string{}
+	for scanner.Scan() {
+		line := scanner.Text()
+		if line == "" {
+			continue
+		}
+
+		if strings.HasPrefix(line, "#") {
```
wlynch, line 186: Let's add a comment into the good test just to be safe.

wlynch (approved):

PR #1490, opened 2025-01-22
fix multi key support in APKINDEX verification

tcnghia:
The current logic uses the last .SIGN. file content as signature and verify that against all known keys. This PR fixes the logic to use matching signature to the right verification key.


pkg/apk/apk/index.go
```diff
@@ -366,61 +373,67 @@ func parseRepositoryIndex(ctx context.Context, u string, keys map[string][]byte,
...
-					verified = true
-					break
+		indexDigest := make(map[crypto.Hash][]byte, len(keys))
+		verified := false
+		for _, sig := range sigs {
+			// compute the digest if not already done
+			if _, hasDigest := indexDigest[sig.DigestAlgorithm]; !hasDigest {
+				digest, err := sign.HashData(indexData, sig.DigestAlgorithm)
+				if err != nil {
+					return nil, err
```
jonjohnsonjr, line 424: Can we annotate this error a little bit?

pkg/apk/apk/index.go
```diff
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/chainguard-dev/clog"
+	"github.com/charmbracelet/log"
```
jonjohnsonjr, line 35: Need to `clog.FromContext()` to avoid pulling this in.

tcnghia: Done

jonjohnsonjr (approved):

xnox:
@tcnghia thank you!!!!!!

PR #1495, opened 2025-01-24
Return an if fetching index fails

jonjohnsonjr:
We weren't actually checking the response code, so we get weird gzip errors instead of the HTTP status.


luhring:
"error"?

luhring (approved):

PR #1516, opened 2025-02-07
insert usrMerge. because you want to do that.

smoser:
add usrMerge to the list without any checks for now. and then when getting a list of packages, make sure that usrmerge ends up at the front.

See also https://github.com/wolfi-dev/os/pull/41575

### DIFF
--- a/examples/wolfi-base.yaml
+++ b/examples/wolfi-base.yaml
@@ -7,6 +7,7 @@ contents:
     - https://apk.cgr.dev/wolfi-presubmit/3bc259540e57442184d2bcd0dc091f190eaefea8
   packages:
     - wolfi-base
+    - bash-binsh
     - usrmerge-baselayout
 
 cmd: /bin/sh -l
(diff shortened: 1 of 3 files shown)


xnox:
In apko, we already have `--package-append` cmdline argument, and I would prefer to use that. Meaning opt-into usrmerge by specifying `--package-append=usrmerge-baselayout` instead of `bc.insertUsrMerge()`. Or like initialize package-append with usrmerge-baselayout always as the rest of the code picks that up correctly then.

Comparison with apk-tools, it has a famous package too. https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/2.14-stable/src/app_upgrade.c?ref_type=heads#L80 when performing upgrade, it resolves famous package name "apk-tools" and if update of that is present, a transaction is started to upgrade apk-tools itself and any new dependencies it needs, followed by re-exec and continuing upgrade. In Alpine, equivalent of usrmerge-baselayout would be then a runtime dependency of apk-tools to achieve similar behavior.

smoser:
> In apko, we already have --package-append cmdline argument, and I would prefer to use that

yes, this is just a proof to see if it actually works.
we shouldn't have apko on its own decide to shove a package in.  the other thing here is that it orders that early.  so we could use '--package-append', but we could also '--package-prepend' which is more what we actually want.

PR #1526, opened 2025-02-14
start using go 1.24

ajayk:


.github/workflows/verify.yaml
```diff
@@ -24,9 +24,9 @@ jobs:
           check-latest: true
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@051d91933864810ecd5e2ea2cfd98f6a5bca5347 # v6.3.2
+        uses: golangci/golangci-lint-action@0adbc47a5910e47adb692df88187ec8c73c76778 # v6.4.0
         with:
-          version: v1.63
+          version: v1.64.5
```
cpanato, line 29: keep 
```suggestion
          version: v1.64
```

then it will pick all patch and we dont need to update

go.mod
```diff
@@ -1,6 +1,6 @@
 module chainguard.dev/apko
 
-go 1.23.4
+go 1.24.0
```
cpanato, line 3: sure we want that?
this says that the min go is 1.24, and we force all to upgrade to go1.24

i am fine with that, just checking

cpanato:
we ca close this seems all is updated

PR #1540, opened 2025-02-25
Make LockImageConfiguration incorporate options

jonjohnsonjr:
We were missing repositories and keys that came in as options.

### DIFF
--- a/pkg/build/lock.go
+++ b/pkg/build/lock.go
@@ -34,13 +34,22 @@ import (
 // architecture that could not be locked. Using the "index" architecture is equivalent to what
 // this used to return prior to supporting per-arch locked configs.
 func LockImageConfiguration(ctx context.Context, ic types.ImageConfiguration, opts ...Option) (map[string]*types.ImageConfiguration, map[string][]string, error) {
-	mc, err := NewMultiArch(ctx, ic.Archs, append(opts, WithImageConfiguration(ic))...)
+	o, input, err := NewOptions(append(opts, WithImageConfiguration(ic))...)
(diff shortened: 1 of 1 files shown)


pkg/build/lock.go
```diff
@@ -34,13 +34,22 @@ import (
...
 func LockImageConfiguration(ctx context.Context, ic types.ImageConfiguration, opts ...Option) (map[string]*types.ImageConfiguration, map[string][]string, error) {
-	mc, err := NewMultiArch(ctx, ic.Archs, append(opts, WithImageConfiguration(ic))...)
+	o, input, err := NewOptions(append(opts, WithImageConfiguration(ic))...)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	archs := make([]resolved, 0, len(ic.Archs))
-	ics := make(map[string]*types.ImageConfiguration, len(ic.Archs)+1)
+	input.Contents.BuildRepositories = sets.List(sets.New(input.Contents.BuildRepositories...).Insert(o.ExtraBuildRepos...))
```
imjasonh, line 42:
```
	appendUniq := func(a, b []string) []string { return slices.Compact(slices.Sorted(slices.Values(append(a, b...)))) }
```

imjasonh: meh

imjasonh (approved):
Does this mean `archs` gets included too now? Or is that a different thing.

jonjohnsonjr:
> Does this mean archs gets included too now? Or is that a different thing.

I want to say yes but I think maybe it depends.

PR #1560, opened 2025-03-06
Add SubFS implementation for Melange

egibs:
Relates to: https://github.com/chainguard-dev/melange/pull/1804
Relates to: https://github.com/chainguard-dev/melange/issues/1731

This PR is part 1 of 2 (the second part is the Melange counterpart).

We need a SubFS implementation in order to correctly identify packages that are built by Melange when using the memFS filesystem.

Previously, we were doing `readlinkFS(pc.WorkspaceSubDir()` but now we need to maintain a FullFS type throughout the build. Using `fs.Sub` on the Melange side won't work because of this type assertion: https://github.com/chainguard-dev/apko/blob/3e63d01bb69480954b6040f5470197aead3f81da/pkg/apk/tarball/write.go#L250

If that assertion fails, then xattrs are not preserved.

If there's an easier way to orchestrate this outside of adding an entire SubFS implementation please let myself or @stevebeattie know!

### DIFF
--- a/pkg/apk/fs/sub.go
+++ b/pkg/apk/fs/sub.go
@@ -1,0 +1,148 @@
+package fs
+
+import (
+	"errors"
+	"io/fs"
+	"path/filepath"
+	"time"
+)
+
+type SubFS struct {
+	FS   FullFS
+	Root string
+}
+
+func (s *SubFS) Open(path string) (fs.File, error) {
+	if !fs.ValidPath(path) {
+		return nil, &fs.PathError{Op: "open", Path: path, Err: fs.ErrInvalid}
+	}
+	fullPath := filepath.Join(s.Root, path)
+	return s.FS.Open(fullPath)
+}
+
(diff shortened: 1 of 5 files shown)


egibs:
My other thought here was to handle xattrs differently for non-`apkfs.XattrFS` types.

pkg/apk/fs/sub.go
```diff
@@ -0,0 +1,149 @@
...
+	fullPath := filepath.Join(s.Root, path)
+	return s.FS.Chtimes(fullPath, atime, mtime)
+}
+
+func (s *SubFS) Symlink(oldname, newname string) error {
+	return s.FS.Symlink(oldname, newname)
+}
+func (s *SubFS) Link(oldname, newname string) error {
+	return s.FS.Link(oldname, newname)
+
```
imjasonh, line 90:
```suggestion
```

imjasonh (approved):

PR #1564, opened 2025-03-07
apko: make apk cache safer for multi-writers

tcnghia:
Currently the cached files are `mv` to the final destination. While this works for a single process of `apko`, if there are multiple `apko` processes running, an existing file read can be invalidated if the file is later replaced (with another file with the same content).

This PR changes to leverage symlink instead. Since file reads in Go by default will follow symlink at file open time, overwriting the symlink won't invalidate existing reads.

We also check for the existence of the destination file before attempting to advertise the cached file, and we clean up the cached file if we didn't create a symlink at all. This is safe because there should be no way anyone else could read our version of the cache file.

I've tested with both NFS and gcs-fuse FS.

### DIFF
--- a/pkg/apk/expandapk/expandapk.go
+++ b/pkg/apk/expandapk/expandapk.go
@@ -175,13 +175,12 @@ func (m *multiReadCloser) Close() error {
 }
 
 func (a *APKExpanded) Close() error {
-	errs := []error{}
-
-	if a.tempDir != "" {
-		errs = append(errs, os.RemoveAll(a.tempDir))
-	}
-
-	return errors.Join(errs...)
+	// We cant' remove the tempDir because our cached files are advertised
+	// by symlinks pointing into them.
+	//
+	// This is not a big deal because the temp files if not referred by
+	// a symlink will be cleaned up anyway.
+	return nil
 }
 
 // An implementation of io.Writer designed specifically for use in the expandApk() method.
--- a/pkg/apk/expandapk/expandapk.go
+++ b/pkg/apk/expandapk/expandapk.go
@@ -175,12 +175,13 @@ func (m *multiReadCloser) Close() error {
 }
 
 func (a *APKExpanded) Close() error {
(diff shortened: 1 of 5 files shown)


pkg/apk/expandapk/expandapk.go
```diff
@@ -175,13 +175,12 @@ func (m *multiReadCloser) Close() error {
...
-		errs = append(errs, os.RemoveAll(a.tempDir))
-	}
-
-	return errors.Join(errs...)
+	// We cant' remove the tempDir because our cached files are advertised
+	// by symlinks pointing into them.
+	//
+	// This is not a big deal because the temp files if not referred by
+	// a symlink will be cleaned up anyway.
+	return nil
```
jonjohnsonjr, line 183: There are other callers into this (I didn't expose this package) that do actually want to clean up the temp files, so I'd rather keep this the same (simplify to just `os.RemoveAll(a.tempDir)`) and remove the call to `Close()` inside of apko.

tcnghia: Done. Thanks

jonjohnsonjr (approved):

xnox:
@tcnghia @jonjohnsonjr these symlinks are absolute, but should be relative. As bind-mounting cache to a different root path is now broken; but was previously in use

```
lrwxrwxrwx  1 xnox xnox  136 Apr  1 13:50 51cfb0f5420c92c85b00e3d4e8d4fc1cb638646d.ctl.tar.gz -> /home/xnox/.cache/dev.chainguard.go-apk/https%3A%2F%2Fpackages.wolfi.dev%2Fos/x86_64/patch-2.7.6-r8/expand-apk3860169520/stream-0.tar.gz
```

It should be symlink like this

```
lrwxrwxrwx  1 xnox xnox  136 Apr  1 13:50 51cfb0f5420c92c85b00e3d4e8d4fc1cb638646d.ctl.tar.gz -> expand-apk3860169520/stream-0.tar.gz
```

As discovered by @joshrwolf

PR #1598, opened 2025-04-01
paths: prefer to use relative paths

xnox:
Using relative paths, makes the cache relocatable, which helps when
bind-mounting / moving it about.


imjasonh (approved):
Thank you!

PR #1606, opened 2025-04-08
Consider deps resolved if other resolved deps already provide them

dannf:


pkg/apk/apk/repo.go
```diff
@@ -787,6 +787,32 @@ func (p *PkgResolver) getPackageDependencies(ctx context.Context, pkg *Repositor
...
+				// selected satisfy this contraint.
+				satisfiedByProvide := false
+				for _, provide := range picked.Provides {
+					prostraint := cachedResolvePackageNameVersionPin(provide)
+					pname, pversion, pcompare := prostraint.name, prostraint.version, prostraint.dep
+					if pname != name {
+						continue
+					}
+					if pversion == "" {
+						satisfiedByProvide = true
```
jonjohnsonjr, line 800: Shouldn't this be false?

dannf: My thought was that, if package `real==1.0` provides `virt==2`, then it seems valid for it to satisfy a resolution for both `real` and `virt`, even though they have different versions. But this is a weird corner case. Is the answer to follow whatever `apk-utils` would do?

jonjohnsonjr: Let me know if I'm misunderstanding this:

Here we have a package `picked` that provides a `constraint`.

We know that the `constraint` has a version associated with it (or we would `continue` on line 778).

We have iterated over `picked.Provides` and found a `prostraint` that provides the `constraint`'s `name`.

If `pversion` has no version but `version` does have a version, I don't think that `picked` actually satisfies the constraint, right?

dannf: You're right. Let me fix that.

jonjohnsonjr (approved):

Issue #1475, opened 2025-01-10
Cache support for `apko lock`


sfc-gh-mhazy:
Hi, 
I would like to add cache capabilities for `apko lock`. 

**Background**

My usecase is that one is developing a large apko image (that must be built with lockfile - imposed by bazel [rules_apko](https://github.com/chainguard-dev/rules_apko)). When iteratively changing the config file, one needs to refresh the lockfile multiple times. For large images this takes quite some time. 

The assumption for the cache working would be that the package repository does not change the contents of package under given version (True for Wolfi, False for Alpine). But if this is not satisfied, one has larger problems with building with lockfile anyway and needs to use some artifactory proxy. 

**Proposed changes**

I have 2 proposals how this could be implemented:
1. Add a cache layer for [FetchPackage](https://github.com/chainguard-dev/apko/blob/bf582adbec1b9ea1b1c8147871d943381e1e17ad/pkg/apk/apk/implementation.go#L1220)  that is used for lock resolution. That cache could be optional and passed only from lock command to not add overhead to `apko build`
    * We could optimize this to cache hashes as well to not recalculate them every time (potentially useful for large packages)
2. Use already existing cache (done in ExpandPackage) for lock resolution. That would just require some changes to how we create  [ApkResolved](https://github.com/chainguard-dev/apko/blob/bf582adbec1b9ea1b1c8147871d943381e1e17ad/pkg/apk/apk/implementation.go#L536) - to make it from [ApkExpanded](https://github.com/chainguard-dev/apko/blob/bf582adbec1b9ea1b1c8147871d943381e1e17ad/pkg/apk/expandapk/expandapk.go#L33).

Please let me know if any of these sounds reasonable to you, happy to implement then (@mattmoor @jonjohnsonjr )

sfc-gh-ptabor:
Implemented: 

https://github.com/chainguard-dev/apko/pull/1612

PR #1612, opened 2025-04-10
CacheDir support for `apko lock`

sfc-gh-ptabor:
Speedup and code alignment of 'apko-lock'. 

Before this change 'apko lock' was downloading all the packages every time and not utilising cacheDir.
After this change the cache is working. Also the fix should speedup apko resolution of packages for regular `apko build` (not lockaed) as
now `ResolveWorld` and Installation are sharing downloaded packages (and code).

Fixes: https://github.com/chainguard-dev/apko/issues/1475

### DIFF
--- a/internal/cli/lock.go
+++ b/internal/cli/lock.go
@@ -71,7 +71,6 @@ func lockInternal(cmdName string, extension string, deprecated string) *cobra.Co
 	var includePaths []string
 	var ignoreSignatures bool
 	var cacheDir string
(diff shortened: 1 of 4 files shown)


sfc-gh-ptabor:
@jonjohnsonjr - could you, please take a look. 
This is pretty high impact PR for intensive iterations on the apko lock config.

jonjohnsonjr (approved):

PR #1613, opened 2025-04-10
Drop pkg/apk/tarball

jonjohnsonjr:
This is used in exactly one place within apko now that the complicated parts have been moved to melange, so we no longer expose the implementation and just move it into the package where it's used.


pkg/build/tarball.go
```diff
@@ -31,7 +31,26 @@ import (
...
+// The etc/passwd and etc/group file provide username and group name mappings for the tar.
+func writeTar(ctx context.Context, dst io.Writer, fsys apkfs.FullFS) error { //nolint:gocyclo
+	ctx, span := otel.Tracer("go-apk").Start(ctx, "WriteTar")
+	defer span.End()
+
+	tw := tar.NewWriter(dst)
+
+	// get the uname and gname maps
+	usersFile, _ := passwd.ReadUserFile(fsys, "etc/passwd")
+	groupsFile, _ := passwd.ReadGroupFile(fsys, "etc/group")
```
Illbjorn, line 44: Is it OK for us to drop these errors?

jonjohnsonjr: Yeah I think because those files might not exist, we don't care if it fails, and in practice this uses an in-memory implementation of the FS, so it can't fail.

But yeah I mostly just don't want any behavioral change here.

PR #1628, opened 2025-04-17
Undeprecate including other configurations

justinvreeland:
This was deprecated in https://github.com/chainguard-dev/apko/pull/1522

> deprecate some stuff we don't use and shouldn't recommend

I'd like to use it, and while yaml might not have a defined include it's useful for defining images that are the same with additional package or similiar.


justinvreeland:
Considering alternatives.

coreydaley-cg:
Closing as stale: no movement since it opened and no config in our repos uses `include`. The feature stays deprecated as it is on `main`. If you still want it undeprecated, reopen and it can go in alongside #2194.

PR #1633, opened 2025-04-22
Apko 0.26.0 regression: Build with base stopped working.

sfc-gh-ptabor:
Bring back the building with base functionality as well as integration test for this functionality (with script to regenerate the golden files.)

The bug was in: pkg/build/types/image_configuration.go file that ignored existenes of the `BaseImage` property.


sfc-gh-ptabor:
@jonjohnsonjr -> please take a look. This makes apko 0.26.0 not usable for us.

PR #1659, opened 2025-05-06
Upgrade go-containerregistry

xnox:
Some other projects use new go-containerregistry, lets upgrade it here
too.


xnox:
Not sure if we want this or not.

k4leung4 (approved):

PR #1664, opened 2025-05-07
Respect arch flags when using lockfiles

jonjohnsonjr:
This ~regressed in https://github.com/chainguard-dev/apko/pull/1654


jonjohnsonjr:
Added a tiny test.

zmarano (approved):

xnox:
note pre-existing test cases did not pass with this change upon rules_apko release; it has been released with this change in place and a release note stating that multi-arch image may or may not work correctly with rules_oci.

PR #1665, opened 2025-05-08
ldsocache snapshot/20250507

dannf:
- **LoadCacheFile(): use a ReadSeeker instead of a file path**
- **Move code from AddLDSOCacheEntriesForDir() into new getLibInfo()**
- **Move actual elf parsing code out into getElfInfo()**
- **Reimplement Test_GenerateCacheFile() using mock data**
- **Allow the same library to exist in multiple paths**
- **Add a test case for a .so file with no SONAME**
- **Add an ldconfig command for testing purposes**
- **Reverse sort entries to match ldconfig**
- **Add 2025 to Copyright string**
- **Fix sort to operate on file basename**
- **Test that we include libs of the same name in different dirs**
- **Only examine files that have valid library names**
- **test: mock_getElfInfo: don't crash reading empty files**
- **test: Add an empty directory**
- **test: Add some libdir symlinks as test cases**


internal/ldso-cache/ldsocache.go
```diff
@@ -129,6 +129,103 @@ type LDSOCacheFile struct {
...
+}
+
+func doGetElfInfo(libfReaderAt io.ReaderAt) (elfInfo, error) {
+	info := elfInfo{}
+
+	elflibf, err := elf.NewFile(libfReaderAt)
+	if err != nil {
+		return info, err
+	}
+	// FIXME: do we need to check for the ELF magic bytes?
```
xnox, line 144: Yes, it is security vulnerability otherwise. see how fun files were used to do remote code execution via unsuccessful ssh connections.

internal/ldso-cache/ldsocache.go
```diff
@@ -307,8 +340,28 @@ func AddLDSOCacheEntriesForDirs(fsys fs.FS, libdirs []string) ([]LDSOCacheEntry,
...
+		if err != nil {
+			return nil, err
+		}
+		allEntries = append(allEntries, entries...)
+	}
+
+	// ld expects entries to be reverse-sorted by name. Otherwise
+	// it may report "No such file or directory"
+	sort.Slice(allEntries, func(i, j int) bool {
+		return filepath.Base(allEntries[j].Name) < filepath.Base(allEntries[i].Name)
```
xnox, line 357: is this compatible with non-utf C locale that ld.so expects?

xnox (approved):

xnox:
i guess my comments need to be addressed in the underlying repo.

internal/ldso-cache/ldsocache.go
dannf: thanks for reminding me of this. I agree that this should be checked at some level, and I wondered if debug/elf already guaranteed this. I think we're good:
https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/debug/elf/file.go;drc=95b433eed428afbb4ab32f0f2541774e939989c7;l=294

internal/ldso-cache/ldsocache.go
dannf: Good question. Perhaps I should use text/collate w/ the C locale here.

PR #1690, opened 2025-05-28
Add an OperatingSystem package to our image SBOMs

sil2100:
We are missing some information in our image SBOMs that cause trivy to generate warnings. The missing piece is a package declaring what operating system our images are using.

### DIFF
--- a/pkg/sbom/generator/spdx/spdx.go
+++ b/pkg/sbom/generator/spdx/spdx.go
@@ -629,7 +629,7 @@ func (sx *SPDX) GenerateIndex(opts *options.Options, path string) error {
 // addOperatingSystem adds a package describing the operating system
 func addOperatingSystem(doc *Document, opts *options.Options) {
 	osPackage := Package{
-		ID:               fmt.Sprintf("SPDXRef-Package-%s", stringToIdentifier(opts.OS.ID)),
+		ID:               fmt.Sprintf("SPDXRef-OperatingSystem-%s", stringToIdentifier(opts.OS.ID)),
 		Name:             opts.OS.ID,
 		Version:          opts.OS.Version,
 		Supplier:         supplier(opts),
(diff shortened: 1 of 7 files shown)


pkg/sbom/generator/spdx/spdx.go
```diff
@@ -623,6 +626,22 @@ func (sx *SPDX) GenerateIndex(opts *options.Options, path string) error {
...
+func addOperatingSystem(doc *Document, opts *options.Options) {
+	osPackage := Package{
+		ID:               fmt.Sprintf("SPDXRef-OperatingSystem-%s", stringToIdentifier(opts.OS.ID)),
+		Name:             opts.OS.ID,
+		Version:          opts.OS.Version,
+		Supplier:         supplier(opts),
+		FilesAnalyzed:    false,
+		Description:      "Operating System",
+		DownloadLocation: NOASSERTION,
+		PrimaryPurpose:   "OPERATING-SYSTEM",
```
xnox, line 639: This SPDX Document is not valid due to:
	$.packages[2].primaryPackagePurpose: does not have a value in the enumeration ["OTHER", "INSTALL", "ARCHIVE", "FIRMWARE", "APPLICATION", "FRAMEWORK", "LIBRARY", "CONTAINER", "SOURCE", "DEVICE", "OPERATING_SYSTEM", "FILE"]

So i think you want OPERATING_SYSTEM but i think we need to double check multiple tools and specs as we have had issues with dashes and underscores before.

xnox (changes requested):

pkg/sbom/generator/spdx/spdx.go
xnox: v3.0.1 spec (which we don't use yet) has it as operatingSystem at https://spdx.github.io/spdx-spec/v3.0.1/model/Software/Vocabularies/SoftwarePurpose/

v2.3 spec has it as a dash https://spdx.github.io/spdx-spec/v2.3/package-information/#7241-description

@johnfosborneiii or @jspeed-meyers were you involved in spdx validation tools? i think when adding new one we really want to use dash as per spdx v2.3 do we need to fix any tooling for this?

sil2100: We currently define our standard as 2.3, so I'd stay with a dash for now?

xnox: please keep dash.

sil2100: Our change is approved upstream and I merged the new spdx-tools patch. I think we're good?

sil2100:
Tested against a locally built image: the sbom seems to be correct and trivy seems to be happy with it. Ship it!

PR #1692, opened 2025-05-29
DNM: Modify package SBOMs to add the OS package

sil2100:
Still in progress. This is a hack right now.


xnox:
this landed in melange instead

PR #1694, opened 2025-05-30
Retract v0.27.8

wlynch:
This tag was accidentally cut from a commit not from main. This tag has been deleted, but we are also explicitly retracting so that any caches that might have pulled it in already know not to use it.


sil2100:
Thank you! And sorry about that.

PR #1729, opened 2025-06-27
Pass pkg pins from meta package to installable package

### LINKED ISSUE #1541: Unable to build image with local packages with dependencies `not in indexes`
I've been following a few issues with local packages recently. Originally, I was hoping #1264 would solve my problems, but with newer versions, I've identified some other problems. Other issues like #1448 were resolved, but didn't solve all the issues for local packages.

As of `apko` 0.23.0+ I've identified a new problem where local packages that have dependencies from upstream cannot be built. Packages that do not have dependencies (such as statically compiled) seem to work fine with the latest (0.25.1).

Here is what the output looks like in 0.25.1:

```
% apko build --arch aarch64 image.yaml image test.tar --ignore-signatures
Error: building "arm64" layer: installing apk packages: error getting package dependencies: solving "py3.13-celery-prometheus-exporter-bundle=1.7.0-r0" constraint: not in indexes
2025/02/24 22:24:36 INFO error during command execution: building "arm64" layer: installing apk packages: error getting package dependencies: solving "py3.13-celery-prometheus-exporter-bundle=1.7.0-r0" constraint: not in indexes
```

Version 0.22.1 still works with local dependencies, but everything newer appears to fail.

mbrancato:
This fixes a problem introduced after `v0.22.1` installing local packages. The local package index is now a pin (specified using the `@` syntax). When a meta-package is specified with a pin, the resolved package that provides the meta-package does not specify the pin. This leads to `apko` only checking non-local indexes when the providing packages are also in the local index.

This PR passes the pin specified on the meta-package to the providing package to be installed.

Fixes #1541

### DIFF
--- a/pkg/build/lock.go
+++ b/pkg/build/lock.go
@@ -20,6 +20,7 @@ import (
 	"maps"
 	"reflect"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 
@@ -267,7 +268,7 @@ func unify(originals []string, inputs []resolved) (map[string][]string, map[stri
 
 	// Allocate a list sufficient for holding all of our locked package versions
 	// as well as the packages we were unable to lock.
-	pl := make([]string, 0, len(acc.versions)+missing.Len())
+	allPl := make([]string, 0, len(acc.versions)+missing.Len())
 
 	// Append any missing packages with their original constraints coming in.
 	// NOTE: the originalPackages "versions" includes the remainder of the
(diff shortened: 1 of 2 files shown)


xnox:
we also have a different regression fix for which was merged just now. Would want to get that out first, before investigating this.

sanmai-NL:
@xnox Which PR/merged commit is that, please?

mbrancato:
@xnox I tried the latest commit in `main` and still get the same error. I also tried my unit test in `main` and it also failed.

For reference, I simply tried a clone of the `py3-requests.yaml` file from `wolfi-os` with the following patch applied:
```patch
--- py3-requests.yaml   2025-07-07 17:57:33
+++ py3-requests-foo.yaml       2025-07-07 17:58:04
@@ -1,6 +1,6 @@
 # Generated from https://pypi.org/project/requests/
 package:
-  name: py3-requests
+  name: py3-requests-foo
   version: "2.32.4"
   epoch: 0
   description: Python HTTP for Humans.
@@ -21,7 +21,7 @@
       - py3-supported-wheel
 
 vars:
-  pypi-package: requests
+  pypi-package: requests-foo
 
 data:
   - name: py-versions
```

mkdir -p key
mkdir -p packages/build
melange keygen key/mykey.rsa


PKG_OUT_DIR=packages/build
SIGNING_KEY_PRIV=key/mykey.rsa
SIGNING_KEY_PUB=key/mykey.rsa.pub
PKG_FILE="py3-requests-foo.yaml"
ARCH=arm64

docker run --privileged -v "$PWD":"$PWD" -w "$PWD" \
  "ghcr.io/wolfi-dev/sdk:latest" -c "melange build --arch $ARCH \
  --out-dir '$PKG_OUT_DIR' \
  --pipeline-dir 'pipelines' \
  --signing-key '$SIGNING_KEY_PRIV' \
  --repository-append 'https://packages.wolfi.dev/os' \
  --keyring-append 'https://packages.wolfi.dev/os/wolfi-signing.rsa.pub' \
  '$PKG_FILE'"

Then with this test image file, I tried to build:
```yaml
contents:
  repositories:
    - "@local packages/build/"
  keyring:
    - key/mykey.rsa.pub

  packages:
    - py3-requests-foo@local
archs:
  - arm64
```

Resulting in the same failure.
```
% go install chainguard.dev/apko@ec7cd6819940c4069cc5f56edb416db9c1450fa9
% apko version                                                           
     _      ____    _  __   ___
    / \    |  _ \  | |/ /  / _ \
   / _ \   | |_) | | ' /  | | | |
  / ___ \  |  __/  | . \  | |_| |
 /_/   \_\ |_|     |_|\_\  \___/
apko

GitVersion:    v0.29.2
GitCommit:     unknown
GitTreeState:  unknown
BuildDate:     unknown
GoVersion:     go1.24.3
Compiler:      gc
Platform:      darwin/arm64

% apko build test.yaml image test.tar                                    
Error: building "arm64" layer: installing apk packages: error getting package dependencies: solving "py3.13-requests-foo=2.32.4-r0" constraint: not in indexes
2025/07/07 18:36:09 INFO error during command execution: building "arm64" layer: installing apk packages: error getting package dependencies: solving "py3.13-requests-foo=2.32.4-r0" constraint: not in indexes
```

coreydaley-cg:
Applied only the test onto current `main` and it fails, so the bug is real and the fix is still needed. It needs a signed commit, and CI has never run here; once it's signed I'll trigger the workflows.

PR #1737, opened 2025-07-01
github: copy release workflow from melange

xnox:
Enable cronned automatic releases as well as on demand release.

This removes need to manually create and push tags.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,19 +1,13 @@
-
-name: Create Release
+name: Release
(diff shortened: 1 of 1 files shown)


imjasonh (commented):
Is this the same as melange's setup with this change? This looks fine I just want it to be consistent.

luhring:
This should've copied the release documentation as well — e.g. https://github.com/chainguard-dev/melange/blob/main/RELEASE.md

PR #1752, opened 2025-07-11
Write a partial installed db to every layer

jonjohnsonjr:
Some scanners operate on a per-layer basis, so having our installed database only exist in the top-most layer leads to false positives.

By including the portion of the installed db that is relevant to each layer, we can appease those scanners at the cost of a small amount of duplication in each layer. I don't love this, but it's an option.

### DIFF
--- a/pkg/apk/apk/implementation.go
+++ b/pkg/apk/apk/implementation.go
@@ -653,7 +653,7 @@ func (a *APK) ResolveAndCalculateWorld(ctx context.Context) ([]*APKResolved, err
 }
 
 // FixateWorld force apk's resolver to re-resolve the requested dependencies in /etc/apk/world.
-func (a *APK) FixateWorld(ctx context.Context, sourceDateEpoch *time.Time) ([]*Package, error) {
+func (a *APK) FixateWorld(ctx context.Context, sourceDateEpoch *time.Time) ([]InstalledDiff, error) {
 	log := clog.FromContext(ctx)
 	/*
 		equivalent of: "apk fix --arch arch --root root"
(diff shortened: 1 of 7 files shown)


pkg/apk/apk/implementation.go
```diff
@@ -653,7 +653,7 @@ func (a *APK) ResolveAndCalculateWorld(ctx context.Context) ([]*APKResolved, err
 }
 
 // FixateWorld force apk's resolver to re-resolve the requested dependencies in /etc/apk/world.
-func (a *APK) FixateWorld(ctx context.Context, sourceDateEpoch *time.Time) ([]*Package, error) {
+func (a *APK) FixateWorld(ctx context.Context, sourceDateEpoch *time.Time) ([]*Package, [][]byte, error) {
```
imjasonh, line 656: Can you document in the doc comment what the `[][]byte` represents?

pkg/apk/apk/implementation.go
```diff
@@ -698,7 +698,7 @@ func (a *APK) FixateWorld(ctx context.Context, sourceDateEpoch *time.Time) ([]*P
 	return a.InstallPackages(ctx, sourceDateEpoch, allInstPkgs)
 }
 
-func (a *APK) InstallPackages(ctx context.Context, sourceDateEpoch *time.Time, allpkgs []InstallablePackage) ([]*Package, error) {
+func (a *APK) InstallPackages(ctx context.Context, sourceDateEpoch *time.Time, allpkgs []InstallablePackage) ([]*Package, [][]byte, error) {
```
imjasonh, line 701: And here

pkg/apk/apk/installed.go
```diff
@@ -48,11 +48,11 @@ func (a *APK) GetInstalled() ([]*InstalledPackage, error) {
 }
 
 // addInstalledPackage add a package to the list of installed packages
-func (a *APK) AddInstalledPackage(pkg *Package, files []tar.Header) error {
+func (a *APK) AddInstalledPackage(pkg *Package, files []tar.Header) ([]byte, error) {
```
imjasonh, line 51: And here

pkg/build/build_implementation.go
```diff
@@ -130,7 +130,7 @@ func newLayerWriter(out *os.File) *layerWriter {
 	}
 }
 
-func (bc *Context) buildImage(ctx context.Context) ([]*apk.Package, error) {
+func (bc *Context) buildImage(ctx context.Context) ([]*apk.Package, [][]byte, error) {
```
imjasonh, line 133: And here

pkg/build/build_implementation.go
```diff
@@ -141,39 +141,40 @@ func (bc *Context) buildImage(ctx context.Context) ([]*apk.Package, error) {
...
+				return nil, nil, err
 			}
 		}
 	}
 
 	var (
-		pkgs []*apk.Package
-		err  error
+		pkgs  []*apk.Package
+		diffs [][]byte
```
imjasonh, line 153: This is only ever going to be an additive diff right? This is "each layer's installed DB", and the solve that has already previously happened, so it won't ever be removing anything, only splitting the adds into chunks.

I don't know if this makes sense, or if there's some better word for "diff" that might make it clearer to read.

(Not blocking at all, just musing)

pdeslaur:
Do we think we'd be able to merge this today?

imjasonh (approved):

PR #1753, opened 2025-07-13
Use /usr/bin/s6-svscan if it exists

lxea:
/bin/s6-svscan has been moved to `/usr/bin/s6-svscan` since Alpine 3.21  which results in errors when trying to run containers using service bundles with any repos after 3.20 

```
Error: crun: executable file `/bin/s6-svscan` not found: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

If I set the alpine main repository to 3.20 it fixes the issue, however version mixing is unsupported


tstromberg (approved):

tstromberg (approved):
Still LGTM, but needs an approver with write access :(

PR #1756, opened 2025-07-15
fix: lower DirFS' new file permissions

vishal-chdhry:
## Explanation

Lower permissions of newly created file. 
- [os.Create](https://pkg.go.dev/os#Create) creates a file with `0666` permissions. This update lowers permission to `0644`
- DirFS is used to create files in APKs (by melange) and the image. If SBOMs are created using DirFS, they are world editable which violates compliance standards.

### DIFF
--- a/pkg/apk/fs/rwosfs.go
+++ b/pkg/apk/fs/rwosfs.go
@@ -352,7 +352,7 @@ func (f *dirFS) Create(name string) (File, error) {
 	if f.createOnDisk(name) {
 		// close the memory one
 		_ = file.Close()
-		file, err = os.Create(filepath.Join(f.base, name))
+		file, err = os.OpenFile(filepath.Join(f.base, name), os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
 		if err != nil {
 			return nil, err
 		}


pkg/apk/fs/rwosfs.go
```diff
@@ -352,7 +352,7 @@ func (f *dirFS) Create(name string) (File, error) {
 	if f.createOnDisk(name) {
 		// close the memory one
 		_ = file.Close()
-		file, err = os.Create(filepath.Join(f.base, name))
+		file, err = os.OpenFile(filepath.Join(f.base, name), os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
```
stevebeattie, line 355: This is a good change, but still leaves the `f.overrides.Create(name)` in the upper DirFS(). It's not necessarily clear when the permissions of the upper DirFS will be used, so we should verify that all of them have safe defaults. I created https://github.com/chainguard-dev/apko/pull/1783 (with some dumb tests) to fix the primary offenders I found.

stevebeattie: Ah right, looking more closely, a MemFS is used as the override layer within DirFS:
https://github.com/chainguard-dev/apko/blob/e34a2b580a7bfe4e1c7fcdf80b599a26479cbb89/pkg/apk/fs/rwosfs.go#L63-L74

and

https://github.com/chainguard-dev/apko/blob/e34a2b580a7bfe4e1c7fcdf80b599a26479cbb89/pkg/apk/fs/rwosfs.go#L126-L130

It was MemFS's `Create()` that had a hardcoded 0o666 in it (pre merge of https://github.com/chainguard-dev/apko/pull/1783):

https://github.com/chainguard-dev/apko/blob/5280de91fa96beabf3f9b67e7a8a9adbc152fb50/pkg/apk/fs/memfs.go#L394-L396

PR #1776, opened 2025-07-21
Ensure parent dirs exist for partial idb

jonjohnsonjr:
Some runtimes will get mad if the image is missing parent directories, this ensures we have them.


crane.json
```diff
@@ -0,0 +1,62 @@
+{
```
imjasonh, line 1: unintentional?

imjasonh (changes requested):

PR #1779, opened 2025-07-23
Fix parsing of apk installed for files in top level directory.

smoser:
In a /lib/db/apk/installed file, when there is a file in the top level directory, it will be preceeded by a 'F' record with no value ('F:').

That was confusing sanitizeArchivePath such that the returned InstalledPackages.Files would contain both a top level directory entry with Name == "", and also a file entry with Name == "".

The change here does
1. Fix Name for files in the top level directory
2. Avoids including an entry for top level directory.

'2' is the less obvious fix.  The other option would be to leave the directory entry in place.  I opted to remove it for consistency.  By leaving it in, a package that contained '/foo.txt' would then have an empty directory entry, but a package that contained '/var/foo.txt' would _not_ have that entry.

This was found with a naive apko-based implementation of 'apk info -L', which would show an empty file and an empty directory for the redis-operator-compat package.


smoser:
Just to be super clear here.
per https://wiki.alpinelinux.org/wiki/Apk_spec#Installed_Database_V2
> M: - directory ACL, only if different than the default of 0:0:0755

So my check for lastdir.Name == "" is limited to a 'M:' record. Meaning this is a non-755 permissions on the root directory.  We should not be creating such a thing, and if we _were_, I honestly can't imagine it would work as expected.

PR #1780, opened 2025-07-24
chore(publish): log os/arch by reading config file

hown3d:
if the goos and goarch don't match any manifest, the log output is wrong


coreydaley-cg:
Agreed the log is misleading: when nothing matches GOOS/GOARCH the code falls back to `Manifests[0]` and then logs the platform it didn't use. Rather than reading the config file, which turns a log line into a way for `LoadIndex` to fail, `useManifest.Platform` already has the OS and architecture of the image that was picked, with a nil check for manifests that lack one. Same accuracy, no new error path.

hown3d:
Refactored to use the manifests platform, please take a look :)

PR #1781, opened 2025-07-24
fix(publish): local tags

hown3d:
local tagging doesn't work because the tag used to check is incorrect. The condition is always false, because the src tag always contains the LocalDomain.


coreydaley-cg:
Superseded: `LoadImage` has checked `localDstTag` since #1880 (`51830a2e`), so this fix is already on `main`. Thanks for catching it.

PR #1795, opened 2025-08-01
Fix path traversal vulnerability in cache.go

antitree:
## Summary

- Fixed path traversal vulnerability in `pkg/apk/apk/cache.go`
- Replaced unsafe `filepath.Clean` + `strings.HasPrefix` validation with Go 1.24's `os.Root`
- Added protection against both directory traversal and symlink attacks

## Security Issue

The previous validation in `cachePathFromURL()` could be bypassed:

```go
// Vulnerable code
cacheFile = filepath.Clean(cacheFile)
cleanroot := filepath.Clean(root)
if \!strings.HasPrefix(cacheFile, cleanroot) {
    return "", fmt.Errorf("cache file %s is not within root %s", cacheFile, cleanroot)
}
```

**Attack example**: With root="/cache", a malicious path like "/cache-evil/file" would pass the prefix check.

## Solution

Implemented Go 1.24's `os.Root` which provides:
- Traversal-resistant path validation
- Protection against symlink attacks
- Platform-specific security optimizations
- Proper containment within the specified root directory

## Test Plan

- [x] All existing tests pass
- [x] Code builds successfully
- [x] go vet passes
- [x] Proper code formatting applied

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pkg/apk/apk/cache.go
+++ b/pkg/apk/apk/cache.go
@@ -493,12 +493,21 @@ func cachePathFromURL(root string, u url.URL) (string, error) {
 
 	// url encode it so it can be a single directory
 	repoDir = url.QueryEscape(u2.String())
-	cacheFile := filepath.Join(root, repoDir, dir, filename)
-	// validate it is within root
-	cacheFile = filepath.Clean(cacheFile)
-	cleanroot := filepath.Clean(root)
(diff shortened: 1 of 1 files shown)


tstromberg (changes requested):
To prevent a future regression, this should get a test.

PR #1805, opened 2025-08-08
Simplify the sorting of paths for apk db/installed file.

smoser:
The change here allows us to just use sort.Slice on []tar.Headers. It also makes it much easier to test, as we can simply test the comparison of 2 different entries and assume that sort.Slice is correct.

The thing that is still TODO here is that my changes are not dropping intermediate paths. There is a test in TestSortTarHeaders named intermediate_dirs_in_the_tree_should_be_required_to_preserve_children that is failing now.

Other TODO is a less clunky tarHeadersSort

### DIFF
--- a/pkg/apk/apk/installed.go
+++ b/pkg/apk/apk/installed.go
@@ -410,105 +410,76 @@ func parseInstalledPerms(permString string) (uid, gid int, perms int64, err erro
 	return
 }
 
+// TODO: This function is needed to make the test pass in
+//
+//	TestSortTarHeaders/intermediate_dirs_in_the_tree_should_be_required_to_preserve_children
+//
+// The comment there indicates that we want to drop files that are orphaned rather than
+// keep them.
+func removeOrphanedEntries(headers []tar.Header) {
+	return
+}
+
 // sortTarHeaders sorts tar headers by name. It ensures that all file children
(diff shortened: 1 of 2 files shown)


smoser:
The referenced test came from https://github.com/chainguard-dev/go-apk/pull/138.
Unfortunately there isn't any more context there and @luhring didn't have any more left in his brain.

I wonder if maybe @jonjohnsonjr does?  

Is it important to drop paths that are missing intermediate components like that tests wants to assert?

pkg/apk/apk/installed.go
```diff
@@ -410,105 +410,76 @@ func parseInstalledPerms(permString string) (uid, gid int, perms int64, err erro
...
-		if filepath.Dir(dir) == "." {
-			topLevelEntries = append(topLevelEntries, dir)
+	sep := fmt.Sprintf("%c", filepath.Separator)
+	aToks := strings.Split(aClean, sep)
+	bToks := strings.Split(bClean, sep)
+
+	for n = 0; n < len(aToks)-1 && n < len(bToks)-1; n++ {
+		result = strings.Compare(aToks[n], bToks[n])
+		if result != 0 {
+			return -1
```
smoser, line 453: this looks questionable. i think it should return result, not -1.

PR #1813, opened 2025-08-13
Don't separate cache by repos

markusthoemmes:
Our cache is fully content-addressable in that each package is cached by the control- and datahashes respectively. As such, we don't have to separate the cache by different repos and can instead rely on that content-addressability to ensure packages that are actually different are not polluting each other. This raises cache-hit rates when working with multiple repositories that all have the same packages.


markusthoemmes:
Yeah this isn't actionable right now as you're saying. I'll close this for now.

PR #1822, opened 2025-08-19
Enforce SBOM permissions at build time

egibs:
Similar to #1758 and #1761, we can ensure that SBOM permissions are correct at build time by running a recursive `fsys.Chmod` on the SBOM directory which will take relatively no time.

We also handle this on the Melange side but this gives us consistency and also acts as a failsafe.


egibs:
There's a better way to achieve this so I'll close this out.

PR #1852, opened 2025-09-09
use jonjohnsonjr/tarfs instead of internal/tarfs

imjasonh:


coreydaley-cg:
`internal/tarfs` moved to `pkg/apk/expandapk/tarfs` in #2022, and the replacement module this pointed at is no longer published. Closing as obsolete.

PR #1868, opened 2025-09-26
Add support for tagged repositories

marcofranssen:
This allows for adding tagged repositories. See https://wiki.alpinelinux.org/wiki/Repositories#Tagged_repository


pkg/build/types/types_test.go
```diff
@@ -23,6 +24,50 @@ import (
...
+			Repositories:      []string{"@testing " + alpineEdgeTesting},
+			BuildRepositories: []string{"@community " + alpineEdgeCommunity},
+		},
+		want: fmt.Sprintf("build_repositories:\n    - '@community %s'\nrepositories:\n    - '@testing %s'\n", alpineEdgeCommunity, alpineEdgeTesting),
+	}, {
+		desc: "tagged with creds",
+		in: ImageContents{
+			Repositories: []string{"@myorg " + alpineWithCreds},
+		},
+		want: fmt.Sprintf("repositories:\n    - '@myorg %s'\n", strings.Replace(alpineWithCreds, "pass", "xxxxx", 1)),
```
kevinmdavis: nit: I'd suggest dropping the `strings.Replace()` here and inlining the input and expected URLs. It makes the test intent more clear and is more robust against future test changes where `strings.Replace` could inadvertently not match the input string.

kevinmdavis (approved):

PR #1916, opened 2025-11-05
Upgrade golangci-lint to 2.6.1 and enable modernize linter

markusthoemmes:
The newest golangci-lint version introduced the modernize linter, which enables many of the checks that gopls has been running and pointing out in static linters. This enables that linter and fixes all issues with it. Out of caution, I kept the omitzero suggestions out though.

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
         with:
-          version: v2.6.1
+          version: v2.6
 
       - run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/verify.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
         with:
-          version: v2.2.1
+          version: v2.6.1
```
cpanato: i would drop just to 


```suggestion
          version: v2.6
```

so that will get the lastest version always and reduce the need for us to bump everytime

markusthoemmes: Done

PR #1938, opened 2025-11-17
apk: fix offline cached builds with alpine key discovery

xnox:
Even in offline case, calls to releases.json were performed without
caching. And caching them, appears to not work (possibly bug in
cachePathFromURL() as there are not enough sub-directories).

Offline cache is needed to avoid looping trying to download releases.json when one doesn't have
internet and is building with --offline flag.

Then cache discovery needed to be fixed up, as for the first time there is a possibility of a subdirectory in the cached dir. (thanks to @javacruft)

Also downgrade all Warnings around not able to fetch keys to Debug
level, as most of the time they are red-herring (alpine URLs do not have Chainguard key discovery, and most artifactory urls do not have Chainguard key discovery either potentially).

### DIFF
--- a/pkg/apk/apk/implementation.go
+++ b/pkg/apk/apk/implementation.go
@@ -323,11 +323,9 @@ func (a *APK) InitDB(ctx context.Context, buildRepos ...string) error {
 	for _, repo := range buildRepos {
 		if ver, ok := parseAlpineVersion(repo); ok {
 			if err := a.fetchAlpineKeys(ctx, ver); err != nil {
-				if a.cache != nil && !a.cache.offline {
-					var nokeysErr *NoKeysFoundError
-					if !errors.As(err, &nokeysErr) {
-						return fmt.Errorf("failed to fetch alpine-keys: %w", err)
-					}
+				var nokeysErr *NoKeysFoundError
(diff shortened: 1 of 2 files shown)


pkg/apk/apk/implementation.go
```diff
@@ -880,6 +882,11 @@ func (a *APK) fetchAlpineKeys(ctx context.Context, alpineVersions ...string) err
 
 	u := alpineReleasesURL
 	client := a.client
+	// NB: Does not seem to work, but at least offline builds are
```
imjasonh: I think we might want to just deprecate and remove `fetchAlpineKeys` and require apko users pulling Alpine packages to explicitly specify the keys they want to use to verify Alpine packages.

This change LGTM but I wouldn't want to keep investing in this code path without corresponding tests, and a commitment that we'll work to keep this working (and I don't think we have that)

xnox: I agree, but there are current workflow that is broken for a user. This is a minimal change to unbreak this existing feature.

Then we can look into ripping it out. I always am risk averse to rip out broken things. I somehow prefer to fix things up, prove the work, then rip them out.

PR #1940, opened 2025-11-17
Restructure cache directory to use subdirs to avoid issues with long URL's

javacruft:
Restructure the cache directory into top level domains, using the path component to create subdirectories.

This resolves an issue where long URL's generated with virtualapk is used in conjunction with other APK repo caching tools such as artifactory where the max filename length limit is hit for the top level cache directory name as it includes the full path.

This is not backwards compatible.

### DIFF
--- a/.github/workflows/build-samples.yml
+++ b/.github/workflows/build-samples.yml
@@ -90,6 +90,7 @@ jobs:
               ${build_script} ./apko
             else
               ./apko build ${cfg} ${name}:build /tmp/${name}.tar --arch amd64
+              ./apko build --offline ${cfg} ${name}:build /tmp/${name}.tar --arch amd64
             fi
           done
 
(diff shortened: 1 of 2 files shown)


xnox:
Landing the cache structure change will be good, but possibly want a feature flag for this new behaviour; and/or allow opt out; or coordinate simultanious change across:
- rules_apko
- terraform-provider-apko
- apkoaas

Because rules_apko has own cache implementation and yet it calls apko binary to do the builds in isolated environment.

Separately terraform-provider-apko uses apko as a library and does the cache, and then mounts that cache into apko container the binary to do reproducibility checks and in that case the cache needs to be understood by both.

Thus it would be nice if this new caching code, would also support looking up old cache paths, in addition to the new cache paths.

xnox:
> > Landing the cache structure change will be good, but possibly want a feature flag for this new behaviour; and/or allow opt out; or coordinate simultanious change across:
> 
> In that case previous PR should also be rolled back to not break it when not using the feature flag.
> 
> > Thus it would be nice if this new caching code, would also support looking up old cache paths, in addition to the new cache paths.
> 
> Or indeed fallback to old cache if the paths are empty.

There was no change to cache structure (i.e. using subdirs). We intentionally split it here.

PR #1967, opened 2025-12-05
`MergeInto` should favor the receiver value for `run-as`

mattmoor:
Right now this is backwards, so custom assembly overlays wouldn't appropriately override the base configuration's setting.

### DIFF
--- a/pkg/build/types/image_configuration.go
+++ b/pkg/build/types/image_configuration.go
@@ -154,7 +154,7 @@ func (ic *ImageConfiguration) MergeInto(target *ImageConfiguration) error {
 }
 
 func (a *ImageAccounts) MergeInto(target *ImageAccounts) error {
-	if target.RunAs == "" {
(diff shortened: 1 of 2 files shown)


pkg/build/types/image_configuration.go
imjasonh: Are these wrong too? Do we need to dedupe (with preference for `a`) if a user or group is defined in both?

pkg/build/types/image_configuration_test.go
```diff
@@ -208,7 +208,7 @@ func TestMergeInto(t *testing.T) {
 			StopSignal: "bar",
 			WorkDir:    "bar",
 			Accounts: types.ImageAccounts{
-				RunAs: "bar",
```
imjasonh: Why do `WorkDir`, `Cmd`, etc., all get `"bar"` when this should get `"foo"`?

This code is weird.

PR #1949, opened 2025-11-21
fix: prevent cache collisions when URLs return duplicate ETags

javacruft:
Fixes #1944 where multi-arch builds failed because Alpine Linux
returns the same ETag for different signing keys. The cache used
only ETags as keys, causing different files to collide.

Changes:
- Add URL hash to cache key: etag + sha256(url)[:4]
- Each unique URL gets unique cache entry regardless of ETag
- Add integration test simulating Alpine Linux duplicate ETags
- Add concurrency test for multi-arch build scenario
- Add backward compatibility test documenting cache format change
- Simplify and remove low-value unit tests

Test results:
- All existing tests pass
- Integration test verifies end-to-end cache behavior
- Concurrency test validates multi-arch scenario with goroutines

PR #1968, opened 2025-12-06
cache keys by name

xnox:
- **Examples: build multi-arch alpine-base**
  This will test key discovery with matching ETags.
  

- **Cache keyrings by full name**
  Cache keyrings by full name, as keys that are in the same location may
  have the same ETag.

- **Build samples for all the declared architectures**


xnox:
@javacruft this is smaller in size change than https://github.com/chainguard-dev/apko/pull/1949 yet critically addresses the current bad etag based caching of alpine keys.

PR #1976, opened 2025-12-10
cli/dot: Ignore Duplicate Node errors

IdlePhysicist:
Here I am explicitly ignoring duplicate node errors as it is a
frequently encountered error which we do not need to stop on.


imjasonh:
https://github.com/chainguard-dev/apko/pull/1910 is the relevant apko change

PR #1984, opened 2025-12-12
fix(fs): fix dirFS caching bug where ReadFile returns zeros

smoser:
This fixes a bug in dirFS where WriteFile would cache an empty buffer in the memory overlay even though it wrote the correct data to disk. This caused subsequent ReadFile operations to return zeros instead of the actual file content.

Root cause:
- WriteFile used conditional logic where memContent was left empty when createOnDisk(name) returned true
- The empty buffer was then stored in the overlay via f.overrides.WriteFile(name, memContent, mode)
- ReadFile would return this cached empty buffer when caseSensitiveOnDisk returned false

Impact:
- Any code that wrote a file and then read it back through dirFS could receive zeros instead of actual data
- This particularly affected updateScriptsTar in pkg/apk/apk/installed.go which would corrupt scripts.tar files during package installation

Solution:
- Always cache the actual data in the memory overlay, not an empty buffer
- Simplified WriteFile by removing memContent variable and conditional
- Updated comment to clarify the new behavior

Added comprehensive unit tests to verify:
- Basic write/read cycles work correctly
- Multiple writes/reads maintain data integrity
- Stat calls don't affect caching behavior
- OpenFile/Write/Close patterns work correctly
- The exact updateScriptsTar pattern preserves data

### DIFF
--- a/pkg/apk/fs/rwosfs.go
+++ b/pkg/apk/fs/rwosfs.go
@@ -423,21 +423,15 @@ func (f *dirFS) ReadFile(name string) ([]byte, error) {
 	return f.overrides.ReadFile(name)
 }
 func (f *dirFS) WriteFile(name string, b []byte, mode fs.FileMode) error {
-	var (
-		memContent []byte
-	)
 	if f.createOnDisk(name) {
 		if err := os.WriteFile(filepath.Join(f.base, name), b, mode); err != nil {
 			return err
 		}
-	} else {
-		memContent = b
 	}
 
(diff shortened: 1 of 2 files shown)


smoser:
Some background.... I'm trying to make a clone of `apk` using apko.
In that work, I hit a bug where writing /usr/lib/apk/scripts.tar was getting corrupted.

I'm not certain that claude root-caused this correctly, but the test cases it added seem valid.

PR #2005, opened 2026-01-08
sbom: Include predicate type as the output SBOM.

wlynch:
Currently we use the generator key as the format in the generated SBOM (e.g. "spdx"). It's more useful if we can get the predicate type in the output so we can know how to handle and parse it.

### DIFF
--- a/pkg/sbom/generator/spdx/spdx.go
+++ b/pkg/sbom/generator/spdx/spdx.go
@@ -66,6 +66,10 @@ func (sx *SPDX) Ext() string {
 	return "spdx.json"
 }
 
+func (sx *SPDX) PredicateType() string {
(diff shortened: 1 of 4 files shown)


pkg/sbom/generator/spdx/spdx.go
```diff
@@ -66,6 +66,10 @@ func (sx *SPDX) Ext() string {
 	return "spdx.json"
 }
 
+func (sx *SPDX) PredicateType() string {
+	return "https://spdx.dev/Document"
```
markusthoemmes: apko doesn't really know/care about this, does it? Attestations are written outside of the apko process IIRC, so this feels like a leaky abstraction.

PR #2015, opened 2026-01-13
Centralize all key fetching logic

inteon:
Introduces new `chainguard.dev/apko/pkg/apk/apk/keyring` package, centralizing all key fetching logic.

NOTE: when producing the lock file, we now also download all keys; this allows us to deduplicate these keys, we might want to include key hashes in the lock files in the future

Goal: centralize all key fetching logic, making refactors and bugfixes like https://github.com/chainguard-dev/apko/pull/1949 easier in the future.

### DIFF
--- a/internal/cli/install-keys.go
+++ b/internal/cli/install-keys.go
@@ -2,13 +2,11 @@ package cli
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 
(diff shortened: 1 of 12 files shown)


inteon:
@markusthoemmes could you make CI run on this PR?

What do you think about the general direction of this PR?

PR #2187, opened 2026-04-22
fs: Scope all DirFS operations through os.Root

markusthoemmes:
This runs all filesystem operations of DirFS through an `os.Root` instance to prevent any kind of path-traversal outside of that root, either through walking the paths directly or through following symlinks.

Behavior has largerly been preserved. Notable mentions are:

- `Mknode`: There's no `Mknodat` for non-linux targets but this likely was the effective behavior beforehand as well. We used to gracefully handle errors by writing empty files. That behavior is preserved across platforms.
- `Chmod`/`Chown`: We used to ignore all errors due to filesystem incompatibilities and the presence of the in-memory overlay. We're now only ignoring errors that point to those incompatibilities and are surfacing any other errors to the callers.

### DIFF
--- a/pkg/apk/apk/path_traversal_test.go
+++ b/pkg/apk/apk/path_traversal_test.go
@@ -88,6 +88,257 @@ func TestPathTraversalHardlink(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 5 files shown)


codysoyland (approved):
LGTM 🚢

PR #2193, opened 2026-04-23
apk: format install_if fields without brackets

officialasishkumar:
## Summary
- format APKINDEX and installed DB install_if fields as space-separated values instead of Go slice strings
- update generated image fixtures and publish digest expectations for the byte-level metadata change

Fixes #2059

## Test plan
- [x] go test ./pkg/apk/...
- [x] go test ./...
- [x] git diff --check
- [x] go mod tidy && git diff --exit-code -- go.mod go.sum


coreydaley-cg:
The fix is right and matches how `D:` and `p:` are already written. Two things to land it: rebase onto `main` and regenerate the golden fixtures, since the installed database bytes change, and sign the commit, which is the only red check. Say so if either is a hassle and a maintainer can re-land it with you as co-author.

PR #2210, opened 2026-05-04
retry package fetch+expand on transient errors

### LINKED ISSUE #1903: bug: apko should retry package installs
We have recently beeing seeing apko fail when installing packages with EOF's such as:

```
buildImage failed: installing apk packages: installing packages: expanding libbrotlienc1 (ver:1.2.0-r0 arch:x86_64): unexpected EOF
```

This happens randomly, on different packages, and is not at all consistent. But it now happening frequently.

Of course, this shouldn't happen but it is more and more frequent. At the very least, we should retry that package install.

a-crate:
Adds retry logic around the fetch→expand→verify pipeline in
getPackageImpl(). Random "unexpected EOF" errors during gzip/tar
decompression are now retried up to 3 times with linear backoff.

Fixes #1903

### DIFF
--- a/pkg/apk/apk/package_getter.go
+++ b/pkg/apk/apk/package_getter.go
@@ -20,12 +20,16 @@ import (
 	"crypto/sha1" //nolint:gosec // this is what apk tools is using
 	"encoding/base64"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
+	"time"
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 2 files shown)


a-crate:
`failed to fetch alpine releases: Get "https://alpinelinux.org/releases.json": Head "https://alpinelinux.org/releases.json": HEAD https://alpinelinux.org/releases.json giving up after 5 attempt(s): Head "https://alpinelinux.org/releases.json": dial tcp 213.219.36.190:443: i/o timeout`

Alpine appears to actually have having infrastructure issues, I don't think this is a bug.

pkg/apk/apk/package_getter.go
```diff
@@ -39,6 +43,45 @@ import (
...
+const (
+	// maxFetchRetries is the number of additional attempts after the first failure.
+	maxFetchRetries = 2
+	// retryBaseDelay is the base delay between retry attempts (scaled linearly by attempt number).
+	retryBaseDelay = 1 * time.Second
+)
+
+// isRetryableError reports whether err is a transient error that warrants retrying
+// the fetch+expand pipeline from scratch.
+func isRetryableError(err error) bool {
```
markusboehme, line 55: AFAICT this won't do anything for HTTP 5xx errors, where `fetchPackage` will return a string error `"unable to get package apk at [...]"`, but "server is out for lunch" seems very retry-able. Bringing that up since you mentioned the 500s today during our call.

pkg/apk/apk/package_getter_test.go
```diff
@@ -366,3 +370,111 @@ func TestGetPackage_HashVerification(t *testing.T) {
...
+
+	// Use a bad checksum — this is a permanent error that should NOT be retried.
+	ip := fakePackage(t, &Package{Name: "testpkg", Version: "1.0.0-r0"}, entries, "")
+	ip.checksum = "Q1" + base64.StdEncoding.EncodeToString(make([]byte, 20)) // wrong checksum
+
+	_, err := getter.GetPackage(ctx, ip)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "control hash mismatch")
+	// If it had retried, the error message would contain "after 3 attempts".
+	require.NotContains(t, err.Error(), "after 3 attempts")
```
markusboehme: nit: with the `truncatingTransport`, this could be an assertion on the mock (there's an infinite number of strings that also won't match "after 3 attempts").

markusboehme (dismissed):
Comments are not blocking.

markusboehme (approved):

PR #2216, opened 2026-05-05
cpio: add FromLayers for multi-layer CPIO archives

a-crate:
Extract recordsFromLayer helper and add FromLayers function that
converts multiple container layers into a single CPIO archive with
a shared dedup writer.

### DIFF
--- a/pkg/cpio/layer.go
+++ b/pkg/cpio/layer.go
@@ -26,6 +26,24 @@ import (
 
 // FromLayer converts a container layer to CPIO format.
 func FromLayer(layer v1.Layer, dest io.Writer) error {
+	return FromLayers([]v1.Layer{layer}, dest)
+}
+
+// FromLayers converts multiple container layers to a single CPIO archive.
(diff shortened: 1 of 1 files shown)


a-crate:
`cpio.FromLayer` can't really be called in a reasonable way in succession because it writes the trailer record. Callers would have to call it, drop the trailer record, and then call again. This just gives them a function that makes more sense.

pkg/cpio/layer.go
```diff
@@ -91,9 +104,9 @@ func FromLayer(layer v1.Layer, dest io.Writer) error {
 
 		default:
 			fmt.Printf("Unsupported TAR typeflag: %c for %s\n", header.Typeflag, header.Name)
-			continue // Skip unsupported types
```
justinvreeland, line 94: why'd we get rid of the comments?

justinvreeland (changes requested):

justinvreeland (approved):

PR #2220, opened 2026-05-08
fix: redact runtime repository credentials

immanuwell:
Tiny fix for YAML output: `runtime_repositories` was missing the same URL redaction path used by `repositories` and `build_repositories`.

Repro was pretty direct:

```sh
go test ./pkg/build/types -run TestYamlMarshallingRepositories -count=1
```

Before this, a runtime-only repo like `https://user:pass@...` marshalled with `pass` still visible. Kinda yikes. Now it comes out as `https://user:xxxxx@...`, same as the other repo fields.

Checked:

```sh
go test ./pkg/build/types -count=1
```

No related issue/PR found via `gh` search.


coreydaley-cg:
Mirrors the existing redaction for `Repositories` and `BuildRepositories`, with a test. The only red check is commit signing, which blocks the merge button. Can you sign the commit and force-push? If that's a hassle, say so and a maintainer can re-land it with you as co-author.

coreydaley-cg (approved):

PR #2227, opened 2026-05-12
build: generate SBOMs for minirootfs

officialasishkumar:
## Summary

- wire `build-minirootfs --sbom-path` to the default SPDX SBOM generator
- add a layer-level SBOM generation helper for minirootfs tarball output
- add coverage that verifies a minirootfs build writes an SPDX SBOM

Fixes #2084.

## Validation

- `go test ./internal/cli ./pkg/build ./pkg/sbom/generator/spdx -run 'TestBuildMinirootFSWritesSBOM|TestBuild$|TestBuildWithBase|TestGenerateImageSBOM' -count=1`\n- `go test ./internal/cli -count=1`\n- `go test ./pkg/build ./pkg/sbom/generator/spdx -count=1`\n

### DIFF
--- a/internal/cli/build-minirootfs.go
+++ b/internal/cli/build-minirootfs.go
@@ -27,6 +27,7 @@ import (
 	"chainguard.dev/apko/pkg/build"
 	"chainguard.dev/apko/pkg/build/types"
 	"chainguard.dev/apko/pkg/options"
+	"chainguard.dev/apko/pkg/sbom/generator"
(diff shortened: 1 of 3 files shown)


coreydaley-cg:
This makes `--sbom-path` actually do something, and the test proves it. Needs a signed commit. One non-blocking note: `GenerateLayerSBOM` copies most of `GenerateImageSBOM`; a shared helper for the release data, package list, and generator loop would be a good follow-up.

PR #2229, opened 2026-05-12
build: avoid chmod through path symlinks

### LINKED ISSUE #1270: can't create symlink to a file that does not exist
Now, for reasons (see the bottom) I had liked to create a symlink to a nonexistent file.

With a snippet like this,
```yaml
paths:
  - path: /bin/xyz
    type: symlink
    source: /ko-app/abc
```
I ended up getting
```
Error: failed to build layer image for "arm64": failed to mutate paths: symlink mutation on /bin/xyz: chmod "/bin/xyz": file does not exist
```

A workaround (suggested by @mattmoor, thanks!) is to add an `empty-file`, too:
```yaml
paths:
  - path: /ko-app/abc
    type: empty-file
  - path: /bin/xyz
    type: symlink
    source: /ko-app/abc
```

### Why?! 😮 

I wanted to use apko to build a base image for use with ko. But we also needed the ko-built binary to have a (few) different name(s), so symlinks had to be added. In a unix-y system, you would use `ln -sf` to create a symlink to some target, which may or may not exist. I thus expected apko to work the same.

This workaround works nicely, though -- when using ko with the apko-built base image, the target (empty) file is replaced, and everything falls into place ✨

### LINKED ISSUE #1565: apko `paths` fails to create a folder with a symlink
My use case here is I have an application that expects data at a folder `/var/lib/foo`, but there is an existing configuration script that uses the path `/foo` instead. So I want the folder `/var/lib/foo` to exist with the correct permissions, and a symlink to there from `/foo`. Sounds simple enough.

In a shell, that would be, roughly:
```shell
mkdir -p /var/lib/foo
ls -s /var/lib/foo /foo
```

The expected way:
```yaml
paths:
  - path: /var/lib/foo
    type: directory
    uid: 10000
    gid: 10000
    permissions: 0755
  - path: /foo
    type: symlink
    source: /var/lib/foo
```

This results in the `/foo` symlink existing, but the real path `/var/lib/foo` has all permissions removed.

```sh
$ ls -lad /foo /var/lib/foo
lrwxrwxrwx 1 root root   12 Jan  1  1970 /foo -> /var/lib/foo
d--------- 2 root root 4096 Jan  1  1970 /var/lib/foo
```

Reversing the order won't work:
```yaml
paths:
  - path: /foo
    type: symlink
    source: /var/lib/foo
  - path: /var/lib/foo
    type: directory
    uid: 10000
    gid: 10000
    permissions: 0755
```

Results, expectedly, in:
```shell
Error: building "arm64" layer: failed to mutate paths: symlink mutation on /foo: chmod "/foo": file does not exist
2025/03/11 11:51:59 INFO error during command execution: building "arm64" layer: failed to mutate paths: symlink mutation on /foo: chmod "/foo": file does not exist
```

The workaround - create the directory, create the symlink, set permissions on the directory. So the directory is listed twice:
```yaml
paths:
  - path: /var/lib/foo
    type: directory
  - path: /foo
    type: symlink
    source: /var/lib/foo
  - path: /var/lib/foo
    type: directory
    uid: 10000
    gid: 10000
    permissions: 0755
```

Produces the desired result:
```shell
$ ls -lad /foo /var/lib/foo
lrwxrwxrwx 1 root   root     12 Jan  1  1970 /foo -> /var/lib/foo
drwxr-xr-x 2 nobody nobody 4096 Jan  1  1970 /var/lib/foo
```

officialasishkumar:
## Summary

- skip the automatic chmod/chown pass after `paths` symlink mutations
- skip symlink entries during recursive directory permission walks
- add tests for symlinks to missing targets, preserving target directory permissions, and recursive walks over broken symlinks

Fixes #1270.
Fixes #1565.

## Validation

- `go test ./pkg/build -run 'TestMutatePathsSymlinkAllowsMissingTarget|TestMutatePathsSymlinkDoesNotChangeTargetPermissions|TestMutateDirectoryRecursiveSkipsSymlinks' -count=1`\n- `git diff --check`\n\nNote: I also attempted `go test ./pkg/build -count=1`, but the process stalled in the Go link step on this shared machine and was terminated to avoid leaving it running.\n

### DIFF
--- a/pkg/build/paths.go
+++ b/pkg/build/paths.go
@@ -64,6 +64,9 @@ func mutateDirectory(fsys apkfs.FullFS, o *options.Options, mut types.PathMutati
 			if err != nil {
 				return err
 			}
+			if d.Type()&fs.ModeSymlink != 0 {
+				return nil
+			}
 			if err := mutatePermissionsDirect(fsys, path, mut.Permissions, mut.UID, mut.GID); err != nil {
 				return fmt.Errorf("mutating permissions for path %q: %w", path, err)
 			}
@@ -143,7 +146,7 @@ func mutatePaths(fsys apkfs.FullFS, o *options.Options, ic *types.ImageConfigura
 			return fmt.Errorf("mutating path %q: %w", mut.Path, err)
 		}
 
-		if mut.Type != "permissions" {
+		if mut.Type != "permissions" && mut.Type != "symlink" {
 			if err := mutatePermissions(fsys, o, mut); err != nil {
 				return fmt.Errorf("%s mutation on %s: %w", mut.Type, mut.Path, err)
 			}
--- a/pkg/build/paths_test.go
+++ b/pkg/build/paths_test.go
@@ -1,0 +1,89 @@
+// Copyright 2026 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+import (
+	"io/fs"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


coreydaley-cg:
Both fixes are still needed on `main`: the chmod/chown pass still runs after `symlink` mutations and the recursive walk still follows symlinks. `paths.go` was reworked in #2274 and #2281, so this needs a rebase, plus a signed commit. Same offer as on your other PRs if that's a hassle: a maintainer can re-land it with you as co-author.

PR #2230, opened 2026-05-14
apk/installed: parse Z: (file checksum) token in ParseInstalled

Dev-X25874:
`AddInstalledPackage` writes a `Z:<checksum>` line for every regular
file that carries a checksum, and real installed databases are full of
them. `ParseInstalled` had no `case "Z":` handler, so those lines were
silently dropped on read-back — every reconstructed `tar.Header` came
back with a nil `PAXRecords` map and no checksum.

Add the missing case, mirroring the write path: allocate `PAXRecords`
on demand and store the value under `paxRecordsChecksumKey`, the same
key used by `installRegularFile` and `AddInstalledPackage`.

### DIFF
--- a/pkg/apk/apk/installed.go
+++ b/pkg/apk/apk/installed.go
@@ -338,27 +338,36 @@ func ParseInstalled(installed io.Reader) ([]*InstalledPackage, error) { //nolint
 				fullpath, _ = sanitizeArchivePath(lastDir.Name, val)
 			}
 			lastFile = &tar.Header{
-				Name: fullpath,
-				Mode: 0o644,
-				Uid:  0,
-				Gid:  0,
-			}
-			pkg.Files = append(pkg.Files, *lastFile)
-		case "a":
-			// file perms if not 0o644
-			if lastFile == nil {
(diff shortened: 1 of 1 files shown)


Dev-X25874:
Hey @a-crate — would you mind taking a look at this when you get a chance? This adds the missing `Z:` case in `ParseInstalled` to handle file checksums that were being silently dropped on read-back. Happy to answer any questions or make changes based on your feedback!

pkg/apk/apk/installed.go
a-crate, line 1: Unify the indentation here please, keep all `cases` statements at the same indentation and don't modify indentation from 341:344. `gofmt` should do this for you, I think. The diff and control flow is a bit confusing otherwise.

a-crate:
Sorry for delay. Looking.

a-crate (changes requested):
`Z:` is an APK V2 field: https://wiki.alpinelinux.org/wiki/Apk_spec

apko doesn't have support for APK V2 yet, and while it would be good to add it I don't think adding piecemeal support for individual APK V2 fields is the right way go about this. Even if we parse APK V2 fields in the installed DB we're not going to write new correct entries in V2 format.

Do you have a specific usecase that's broken without this change?

Issue #41, opened 2022-03-03
multiarch support


kaniini:
With the `proot` support landed, we can do multiarch builds, since `--no-scripts` is now a requirement.

From the apk side of things, this can be done by simply passing `--arch [arch]` when setting up the apkdb, e.g. in the `InitApkDb` step.

I'm told that we can possibly recruit @imjasonh to work on this.  Ideally we would probably build images for all of the archs at once, I suppose.

imjasonh:
> I'm told that we can possibly recruit @imjasonh to work on this.

Indeed! 🎉 

How do we want this to look to users? Is this something they should specify in the YAML (`archs: ['amd64', 'arm64', 'ppc64le']`) or in a flag (`apko build my.yaml --archs=amd64,arm64,ppc64le`) -- or both?

Is `--platforms=all` reasonable to support? If this is quick/reliable enough I kind of think we could just make it the default. It's not the default for `ko` because it does a ton of needless builds, but if that's not the case for `apko`...

imjasonh:
> We should probably want to use the Alpine architecture names here, and map them to the OCI ones, but I'm open to doing it the other way around.

I don't have a strong opinion either way. We could even accept both and just warn/fail if both flavors are specified. `--platform=amd64,x86_64` is technically a doable (though misguided) thing to request.

kaniini:
I don't either, do what you want to do with abandon of course :)

kaniini:
One thought I had is that OpenWrt and Adelie have different APK architecture names than Alpine does, so having the ability to override this in the YAML probably makes some sense.

panekj:
I would suggest doing what Docker does, which is accept different forms of arches and internally use whatever you like:
![image](https://user-images.githubusercontent.com/50457605/156644034-3af01352-3840-45db-a6a4-1f9fce491ac3.png)

*Less headache for users*

kaniini:
To be clear, we're talking about mapping OCI architecture names to the architecture names used by APK-ecosystem distributions, not what is actually published in registries, I would assume the registries do those mappings themselves.

imjasonh:
I've got a WIP change here: https://github.com/chainguard-dev/apko/compare/main...imjasonh:feature/multiarch

As of right now:
- `apko build` and `build-minirootfs` both only build tarballs for the current platform, ignoring whatever's in the YAML (with a warning)
[...]

PR #2249, opened 2026-05-26
Add erofs support to apko.

smoser:
Implement apko writing of erofs images according to the spec at
https://github.com/erofs/erofs-image-spec (`spec.md` on `main`; section numbers
in the code and docs refer to it). The spec is still in its draft phase — no
tagged release, and it says media types, annotation keys and the chunk-index
layout may change before the first stable one — so EROFS support here is
experimental.

The docs/erofs.md file in the pr explains the use and erofs image layout.

**Scope:** this PR adds the single-image writer (`--format=erofs`, one layer) and
`apko erofs ls`. Multi-layer splitting and `apko erofs mount`/`umount` were
descoped after review — see #2408 — and land next, in that order. Combining
`layering` with `format: erofs` is rejected during config validation rather than
silently producing a single layer.

I've opened a draft PR showing the value of erofs in images.
See https://github.com/chainguard-dev/melange/pull/2605 for example use of erofs support in apko by melange.

Related:
 * Compression is separate, at #2406.
 * Follow-up work is tracked in #2408.
 * Found one bug in go-erofs https://github.com/erofs/go-erofs/pull/41

### Review round 2 (2026-08-18)

Addressing @mattmoor's review at 46ac139e. Descoped, per his suggestion, the two
areas where the findings were concentrated; the rest is fixed here.

Fixed in this PR:

 * **Whiteouts now use the encoding the spec mandates.** `Stack` detected
   deletions by filename (`.wh.NAME`, `.wh..wh..opq`), which §8.1 item 9 forbids
   in EROFS images outright; §3.6 requires a char device with rdev 0 plus
   `trusted.overlay.opaque="y"`. The comment justifying the old encoding claimed
   it matched go-erofs `Writer.Merge`, but `Merge` *consumes* `.wh.` entries from
   its source and never writes them into an image — so no producer emits them and
   the reader was honoring an encoding that cannot conformantly exist. Whiteouts
   are interpreted only when the stack has two or more layers, since §7 step 6
   lets a lone layer be mounted directly, applying no overlay semantics.
 * **`ReadOCILayers` accepts the role shapes the spec allows.** Role-absent
   non-final layers and `overlay-lower` final layers are both conformant (§2.3,
   §3.8 rule 1, §7 step 1) and were both rejected. `device` and `overlay-data` are
   now refused as unsupported compositions rather than as malformed images.
 * **The build time is always stamped.** `WithBuildTime` was passed only for a
   non-zero timestamp, and go-erofs stamps `time.Now().Unix()` when it is absent —
   so a library caller passing a zero `time.Time` got a different digest on every
   build. It cannot be passed verbatim (a zero `time.Time` has a negative Unix
   seconds value that wraps through `uint64`), so zero and pre-epoch clamp to
   epoch 0.
 * **`os.features` reaches the index platform descriptor.** §5.4 is a MUST in two
   places and only the config half was implemented, so consumers filtering on the
   index — before fetching any config — never saw the signal.
 * **The default output is named `.erofs`**, not `apko-<arch>.tar.gz`.
 * **CI installs `erofs-utils`**, so the `fsck.erofs` cross-checks actually run.
   Without it every one degraded to a `t.Log` and `TestWriteErofs_FsckErofs`
   skipped, leaving go-erofs reading its own output as the only always-on
   validation.
 * Hardlink materialization and the go-erofs v0.3.1 caveats are written down: the
   pinned version predates erofs/go-erofs#41, so `FileInfo.Mode()` misreports
   setuid/setgid/sticky and callers must read `*erofs.Stat.Mode` off `Sys()`.
 * Dropped `erofsLayer.LayerPath()`, which had no callers. The exported
   state/driver surface went away with the mount descope, which also means the
   package no longer has a build-tag-dependent API — everything left is pure Go
   and builds everywhere.
 * Spec citations corrected: the "PR #1" pointer (merged 2026-06-01) now names
   `spec.md`, and four places that presented apko's producer convention as a spec
   rule now say which is which.

Moved to #2408, with the code preserved on the `feat/apko-erofs-full` branch of
smoser/apko so nothing is reconstructed from scratch:

 * Package routing never matching on the real tarfs path, and dir-only subtrees
   being dropped from every layer — plus the `tarfs`-backed test that would have
   caught the first, and the temp-file/fd leak on error paths.
 * State-file trust in `Unmount`, the rerun instruction that cannot work, the
   writable-by-default mount that discards `upper`, and a `Driver` seam so the
   orchestration is testable at all.

Not done here: a size measurement on a hardlink-heavy image. I had no such image
on hand, so `docs/erofs.md` states the arithmetic — one full copy per link,
rounded to the block size — rather than a number I did not measure.


smoser:
fyi @raharper, @rchincha

docs/apko_file.md
```diff
@@ -257,3 +257,16 @@ It contains the following children:
  - `budget`: The number of additional layers apko will use for layering.
 
 See [layering.md](layering.md) for more information.
+
+### Format (experimental)
+
+`format` selects the on-wire layer payload format. Two values are recognized:
+
+ - `tar` (default): standard gzip-compressed tar layers (`application/vnd.oci.image.layer.v1.tar+gzip`).
+ - `erofs`: uncompressed EROFS filesystem images (`application/vnd.erofs`), per the draft [erofs/erofs-image-spec](https://github.com/erofs/erofs-image-spec). EROFS layers advertise `erofs` in the image config's `os.features` so consumers that do not implement the spec can identify and skip them.
```
raharper: This mentions "uncompressed EROFS" which is a bit confusing.  EROFS by default uses LZ4 compression on data, unless you pass flags to to the erofs tool.  The spec mention talks about EROFS "raw" and EROFS+zstd compressed;  but, IIUC, what compression used by erofs is handled by erofs runtime mounting -- what do you mean here w.r.t use of compression, and where.

I'm not sure if the spec discussion of "raw" as EROFS (using lz4 default compression), or completely no compression mode of erofs.  And I'm trying to understand what the spec means by EROFS+ztsd -- does one zstd compress a "raw" EROFS uncompressed image or would that zstd compress default EROFS (lz4) image?

docs/erofs.md
```diff
@@ -0,0 +1,333 @@
...
+```sh
+sudo umount mnt/merged
+for d in mnt/lower*; do sudo umount "$d" 2>/dev/null || fusermount -u "$d"; done
+```
+
+Production runtimes (containerd's erofs snapshotter, podman/CRI-O with the erofs-aware plugin, etc.) automate this assembly; both `apko erofs mount` and the manual steps above are for verifying that an apko-built EROFS image really does compose into a valid rootfs.
+
+## Current limitations
+
+- **No compression.** apko emits raw `application/vnd.erofs` layers only. The draft spec defines `application/vnd.erofs+zstd` but neither apko's writer nor the underlying go-erofs library writes compressed images yet.
```
raharper, line 249: I see, since you're not using existing erofs-utils tools, you don't have compression yet.

docs/apko_file.md
raharper: I see further down you mention that 1) you're not using erofs-utils provided tools for creating erofs images 2) because of (1) there isn't any compression yet.

I'll caution that this sure looks a lot like the trouble that gzip compressed tar layers have in golang where the intermediate layers that transform compressed things can end up getting different hashes.  In stacker we ran into issues with docker recompressing stacker generated gzip blobs with different window.

Not directly relevant to this PR, but we should likely follow up on the erofs-image-spec about that scenario so we can ensure re-compression does not produce different sha256sums.

smoser: I hooked up use of 'mkfs.erofs'. and it will take that path when you ask for compression via `--format=` . i dont love it, but it is nice to not wait on go-erofs for it.

the one _completely_ missing feature is dmverity data now.

smoser:
I think this is ready for some eyes to review and some testing of it. See the description and linked PRs that i put together to show its usefulness.

pkg/build/build.go
```diff
@@ -185,8 +185,42 @@ func (bc *Context) ImageLayoutToLayer(ctx context.Context) (string, v1.Layer, er
...
+		if compressor := bc.ic.Format.Compressor(); compressor != "" {
+			// Compressed EROFS routes to mkfs.erofs (go-erofs can't write
+			// compressed images yet). Close our handle before invoking it so
+			// it can recreate the file; no deferred Close is set, so we own
+			// the lifetime here exactly once.
+			level, _ := bc.ic.Format.CompressionLevel()
+			outName := outfile.Name()
+			if err := outfile.Close(); err != nil {
+				return "", nil, fmt.Errorf("closing erofs outfile before mkfs.erofs: %w", err)
+			}
```
raharper: Below there's a `outfile.Sync()` after `.Close()`; here no sync is done.  Is that OK?  Not clear the need for Sync().  Maybe a comment on what the .Sync() addresses for the go-erofs writer vs the external compressor path?

pkg/build/erofs.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	if mt := info.ModTime(); !mt.IsZero() {
+		if err := w.Chtimes(absPath, time.Time{}, mt); err != nil {
+			return fmt.Errorf("chtimes %s: %w", absPath, err)
+		}
+	}
+
+	if mode.IsRegular() || mode.IsDir() {
+		xattrs, _ := fsys.ListXattrs(fsysPath)
```
raharper: Check for error? or does `ListXattrs` return an empty range on error?

pkg/build/erofs.go
```diff
@@ -0,0 +1,324 @@
...
+// internal compression). The compressed file's SHA-256 is Digest; the
+// uncompressed file's SHA-256 is DiffID and is also exposed as the
+// "org.erofs.uncompressed-digest" descriptor annotation per the draft
+// erofs/erofs-image-spec. extra entries are merged on top.
+func buildCompressedErofsLayerFromFiles(path, uncompressedPath string, extra map[string]string) (v1.Layer, error) {
+	digest, size, err := hashFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("hashing erofs layer %s: %w", path, err)
+	}
+	diffID, _, err := hashFile(uncompressedPath)
```
raharper: stacker annotates the uncompressed size in the layer information.  May be useful to do that as well.

pkg/build/erofs_layers.go
```diff
@@ -0,0 +1,285 @@
...
+// because the per-group installed db is a synthesized payload, not a
+// faithful copy of the source file.
+func writeErofsRegularBytes(w *erofs.Writer, absPath string, info fs.FileInfo, data []byte) error {
+	fout, err := w.Create(absPath)
+	if err != nil {
+		return fmt.Errorf("create %s: %w", absPath, err)
+	}
+	if len(data) > 0 {
+		if _, err := fout.Write(data); err != nil {
+			_ = fout.Close()
```
raharper: Shouldn't we verify we wrote `len(data)` ?

pkg/build/erofs_test.go
```diff
@@ -0,0 +1,530 @@
...
+		if err != nil {
+			return err
+		}
+		got[path] = info
+		return nil
+	}))
+
+	require.Contains(t, got, "a", "directory a missing from image")
+	require.Contains(t, got, "a/b", "file a/b missing from image")
+	require.Contains(t, got, "a/link", "symlink a/link missing from image")
```
raharper: May want to quote the path for easier parsing of the error:

e.g

```
directory 'a' missing from image
```

pkg/build/erofs_test.go
```diff
@@ -0,0 +1,530 @@
...
+		pkg1: []byte("pkg1 info\n"),
+		pkg2: []byte("pkg2 info\n"),
+	}
+
+	layers, err := splitErofsLayers(context.Background(), fsys, groups, pkgToDiff, t.TempDir(), epoch)
+	require.NoError(t, err)
+	require.Len(t, layers, 3, "expected 2 group layers + 1 top layer")
+
+	// All three layers should be valid EROFS images.
+	fsckBin, _ := lookFsckErofs()
```
raharper: check for error?

pkg/build/erofs_test.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+
+		// The image must parse via go-erofs.
+		f, err := os.Open(erl.path)
+		require.NoError(t, err)
+		_, err = erofs.Open(f)
+		_ = f.Close()
+		require.NoError(t, err, "layer[%d] is not a valid EROFS image", i)
+
+		if fsckBin != "" {
```
raharper: is `fsckBin` optional?

pkg/build/layers.go
```diff
@@ -85,6 +86,16 @@ func (bc *Context) buildLayers(ctx context.Context) ([]v1.Layer, error) {
 	}
 
 	// Then partition that single fs.FS into multiple layers based on our layering strategy.
+	if bc.ic.Format.Base() == types.LayerFormatErofs {
+		if compressor := bc.ic.Format.Compressor(); compressor != "" {
+			if err := preflightMkfsErofs(ctx, bc.ic.Format); err != nil {
+				return nil, err
+			}
+			level, _ := bc.ic.Format.CompressionLevel()
```
raharper: err check?

pkg/erofsmount/mount_linux.go
```diff
@@ -0,0 +1,306 @@
...
+		for _, c := range slices.Backward(cleanups) {
+			if err := c(); err != nil {
+				log.Warnf("cleanup on error: %v", err)
+			}
+		}
+	}()
+
+	// Single-layer read-only short-circuit: overlay buys nothing when there's
+	// one lower and no upper, and a lowerdir-only overlay over a single
+	// EROFS mount has been flaky across overlayfs versions. Mount the layer
```
raharper: This is very interesting.  Any further info, bug reports etc?

raharper: It may be unexpected to not have overlay if other non-erofs multi-layer OCI use overlayfs to compose a single layer OCI mount point.

pkg/erofsmount/source.go
```diff
@@ -0,0 +1,190 @@
...
+	// Tag, for KindOCIDir, optionally selects a manifest via the
+	// org.opencontainers.image.ref.name annotation in index.json. Empty means
+	// "use the sole manifest in the index".
+	Tag string
+	// Raw is the original spec the user passed, preserved for error messages.
+	Raw string
+}
+
+const (
+	prefixErofs  = "erofs:"
```
raharper, line 80: How is this used? and how does it differ from oci: with a path that points to an OCI image using erofs layers?  And what if one uses erofs: prefix to point an OCI that has no erofs layers?

raharper (commented):
Looks really solid and complete with extensive testing.  a couple comments/questions I'm interested in see feedback.

pkg/build/erofs.go
smoser: uncompressed size is less useful in erofs and also less meaningful than in a tarball.
less useful because you don't have to uncompress it to use it.  thats the big thing, stacker does that so the consumer can know the size that is otherwise unknown before unpacking.
less meaningful because, iiuc, ou can just have a mix of compressed and uncompressed, and thats perfectly fine.

I'd say if we find we need this we should push that up to the rfc proposal.

pkg/erofsmount/mount_linux.go
smoser: :-(. i lost any specifics i had there.  going to drop that comment. thanks for reading.

pkg/build/build.go
smoser: Good catch, and the `Sync()` turned out to be the part that didn't belong.

It was never load-bearing: `*os.File` does no userspace buffering, so once `Close()` returns, the fresh `Open()` in `buildErofsLayerFromFile` sees every byte. The real problem is the one you're pointing at — the *deferred* `Close`. Both paths hash the finished file, so the close has to happen before the hash **and** report its error; a deferred `Close` would swallow a write error that only surfaces at close time.

So both branches now close explicitly with an error check, no `Sync`, and there's a comment on the block explaining why.

pkg/build/erofs.go
smoser: Checked now — it wasn't returning an empty range on error, the error was just being dropped, so a failed lookup silently produced a file with no xattrs.

apko's `FullFS` implementations keep xattrs in memory for every node they know about, so an error here means the entry we just walked has gone missing, which is a bug worth failing on.

Worth noting that made the dirFS path load-bearing and nothing covered it: every existing erofs test writes from a memFS, while real builds write from a dirFS whose xattr lookups resolve against overrides seeded by walking the backing tree. Added `TestWriteErofs_DirFS` to pin that down.

pkg/build/erofs_layers.go
smoser: Yes — added. The `io.Writer` contract requires a short write to report an error, but go-erofs's writer is young enough that I'd rather not trust that silently. Now compares against `len(data)` and fails with `io.ErrShortWrite`.

pkg/build/erofs_test.go
smoser: Done, via `%q` so the quoting can't drift from the path.

pkg/build/erofs_test.go
smoser: Fixed — see the reply on `fsckBin` below; the helper no longer returns an error to discard.

pkg/build/erofs_test.go
smoser: It is optional, and you're right that nothing said so — three tests ran fsck only `if err == nil` and passed quietly when erofs-utils was absent.

Replaced the ad-hoc lookups (and the helper whose error was being discarded, per your other comment) with a single `optionalFsckErofs(t)` that returns `""` and logs when the binary is missing. Its doc comment states why it's optional: every caller has already parsed the image with go-erofs, so fsck is a second opinion from the C reference implementation rather than the only check.

`TestWriteErofs_FsckErofs` still skips outright, since the erofs-utils cross-check is that test's entire point.

pkg/build/layers.go
smoser: Not an error, though I agree it reads like one — the second return of `CompressionLevel()` reports whether `level=` was explicitly given, not whether parsing failed. `0` means "let mkfs.erofs pick its default".

Added a comment at both call sites saying so.

pkg/erofsmount/source.go
smoser: Documented at the prefix definitions, since this is exactly the confusion the doc was missing.

The prefixes select *how the path is interpreted*, not what the bytes at the far end turn out to be:

* `erofs:` names one EROFS filesystem image file — the single blob apko writes with `--format=erofs`, or a layer blob lifted out of a layout. There's no manifest, so there's nothing to compose; it mounts as a single layer.
* `oci:` / `oci-dir:` name an OCI *layout directory*. `OpenLayers` walks `index.json` and the selected manifest to collect every EROFS layer blob and its `org.erofs.role`, and composes all of them.

So they're not interchangeable even against the same image. Misuse in each direction:

* `erofs:` at an OCI layout directory fails at parse time with "not a regular file" — a directory isn't a blob.
* `erofs:` at a regular file that isn't an EROFS image parses fine, because the prefix asserts intent rather than content, and fails later at mount, since only the kernel (or go-erofs) can judge the superblock.

smoser:
I pulled the compression out. it is now separate at https://github.com/chainguard-dev/apko/pull/2406.
When this lands, we can look to re-add compression as shown that way or via https://github.com/erofs/go-erofs/pull/45

pkg/build/erofs_layers.go
```diff
@@ -0,0 +1,291 @@
...
+			dirInfo[absPath] = info
+			dirFsysPath[absPath] = fpath
+			return nil
+		}
+
+		// Default to the top layer.
+		owner := top
+
+		// If the file info exposes its owning package, route to that group.
+		if pkger, ok := info.Sys().(interface {
```
mattmoor: This assertion is on the wrong receiver, and as far as I can tell it can never succeed in a real build. Real builds reach this walk through tarfs (internal/cli/build.go passes `tarfs.New()` to `build.New`), and tarfs's `memFileInfo.Sys()` returns a fresh `*archive/tar.Header` (pkg/tarfs/fs.go:1012), which has no `Package` method. The `Package()` accessor is on the FileInfo itself (pkg/tarfs/fs.go:1031), which is what the tar split path asserts on (layers.go:323: `f.info.(interface{ Package() *apk.Package })`). `fs.WalkDir` hands that FileInfo back verbatim via `d.Info()` (stdlib `dirInfo` wrapper), so this should assert on `info`, not `info.Sys()`.

As written, `owner` stays `top` for every file: each group layer contains only its ancestor dirs plus the partial installed db (that special case keys on the path string at line 201, so it fires regardless), and the final layer carries the entire rootfs. Per-layer scanners then see a db claiming packages whose files are absent, which inverts the stated purpose.

TestSplitErofsLayers can't catch this because it drives the split through `apkfs.NewMemFS()`, which implements `Package()` nowhere, so the fixture has zero package-owned files by construction and the test passes identically with and without routing. A test that goes through `tarfs.New()` with package-owned files, asserting a file is present in its group layer and absent from top, fails on this line today and passes with the fix.

Related, worth fixing together: the tar path panics on a `packageToWriter` miss (layers.go:327) while this path silently falls back to `top` (lines 191-193). Once the assertion is fixed, that silent fallback can mask grouping bugs the tar path would surface loudly.

pkg/build/erofs_layers.go
```diff
@@ -0,0 +1,291 @@
...
+				// the source rootfs).
+				for _, gw := range writers {
+					if err := emitErofsEntry(gw.w, absPath, fpath, info, fsys, buf); err != nil {
+						return err
+					}
+					gw.emitted[absPath] = true
+				}
+				return nil
+			}
+			dirInfo[absPath] = info
```
mattmoor: Directories whose subtree contains no non-dir entry are never emitted anywhere. They're recorded here and only materialized by `emitAncestors`, which runs for non-dir entries; after the walk there is no sweep that flushes recorded-but-unemitted dirs into any writer, top included. So empty dirs and dir-only chains (`/tmp`, `/run`, `/home`, `/var/empty`, mount points) are absent from the mounted union of a layered image. Both siblings handle this: the tar split writes every walked dir into its owning layer (layers.go:365), and `writeErofs` Mkdirs every dir unconditionally. A post-walk pass emitting any not-yet-emitted dir into `top` (or into every writer, matching the root's treatment above) would cover it.

pkg/build/erofs_layers.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	// One writer per group, plus a top writer for entries not owned by any
+	// package.
+	packageToWriter := map[string]*erofsGroupWriter{}
+	groupToWriter := map[*group]*erofsGroupWriter{}
+	writers := make([]*erofsGroupWriter, 0, len(groups)+1)
+
+	for _, g := range groups {
+		gw, err := newWriter()
```
mattmoor: Error paths in this function leak the temp layer files and their fds: `newErofsLayerFile`'s contract says the caller closes and removes, but none of the error returns do (this loop, the WalkDir failure, a `closer` failure mid-finalize, a `buildErofsLayerFromFile` failure). Compounding it, each go-erofs Writer holds an unlinked spool fd that only `Close` releases, and `Close` both writes out the image (so it can't serve as abort) and early-returns before its spool-close defer when the writer has a sticky error. The CLI is bounded by its MkdirTemp/RemoveAll wrapper, but library callers accumulate files and leak fds until process exit. A named-return cleanup that closes and removes `gw.path` for every writer on error would cover it; an Abort API upstream would make it airtight.

mattmoor (commented):
I took a deep pass over this at head 46ac139e. The single-image writer path (`--format=erofs` producing one layer) held up well under scrutiny: close-before-hash ordering, xattr and special-bit handling, and byte-determinism all check out, and the fixes from the earlier review round are all present at head. `apko erofs ls` is also in good shape.

I found two significant bugs in the multi-layer split path (package routing never matches on the real tarfs path, and dir-only subtrees are dropped from every layer) and two in the umount path (state-file trust, and a rerun instruction that cannot work). Details inline. Given where those are concentrated, one option is to descope: land the single-image format plus `ls` now, and follow up with the layering split and mount/umount hardening.

Cross-cutting notes that don't anchor to one line:

- **melange coordination**: melange#2605 currently pins a fork of apko and passes `erofs+zstd[,level=N]` format strings that this head rejects after the #2406 split; its compressed mode is dead until #2406 merges and melange re-pins. Worth a note on that PR.
- **Exported surface**: `StatePath`/`WriteState`/`LoadState`/`RemoveState`/`StateSchemaVersion` (state.go) and `Driver`/`NewDriver`/`ResolveMode` (driver_linux.go) have no callers outside the package, and the linux-only exports give the package a build-tag-dependent API. Consider unexporting before this ships as a library surface. Similarly `erofsLayer.LayerPath()` has no callers.
- **Test seam for the mount plane**: `Mount`/`Unmount` construct their `Driver` internally, so the 305 lines of orchestration in mount_linux.go (cleanup LIFO, state lifecycle, unmount policy) can't be unit-tested with a fake driver. driver_linux_test.go covers only the argv builders; nothing anywhere executes the orchestration.

Things I checked that came back clean, so others can skip them: determinism of the single-image path (go-erofs sorts xattr keys before serialization; no UUIDs or randomness; SOURCE_DATE_EPOCH plumbed through), OCI blob path traversal in erofsmount (ggcr's Hash parsing forecloses it), regressions to tar-path users from the shared-file changes (the rwosfs char-device seeding fix only changes behavior for DirFS over trees that already contain real device nodes, which no in-tree flow constructs), and the go-erofs supply chain (official erofs org, sumdb-verified tag, containerd pins the identical v0.3.1).

pkg/build/erofs_layers.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil, err
+	}
+
+	// Finalize each writer and produce v1.Layer values.
+	layers := make([]v1.Layer, 0, len(writers))
+	for i, gw := range writers {
+		if err := gw.closer(); err != nil {
+			return nil, err
+		}
+		// All layers except the final (top) carry role=overlay-lower per
```
mattmoor: Citation nit (same text appears in types/erofs.go and docs/erofs.md): spec §3.8 rule 1 actually says the final layer's role must be "either absent or `overlay-lower`", and §2.3 makes `org.erofs.role` OPTIONAL on any layer. "Every non-final layer carries overlay-lower, final carries none" is apko's own (spec-legal) producer convention, not a spec rule. See the related comment on ReadOCILayers, which enforces this convention on the consumer side and rejects spec-valid images.

pkg/build/erofs.go
```diff
@@ -0,0 +1,287 @@
...
+//
+// If buildTime is non-zero it sets the EROFS image build time (used to seed
+// per-entry mtime defaulting and recorded in the superblock), making the image
+// reproducible.
+func writeErofs(ctx context.Context, out io.WriteSeeker, fsys apkfs.FullFS, buildTime time.Time) error {
+	ctx, span := otel.Tracer("apko").Start(ctx, "writeErofs")
+	defer span.End()
+
+	var createOpts []erofs.CreateOpt
+	if !buildTime.IsZero() {
```
mattmoor: Consider passing `WithBuildTime` unconditionally (or rejecting a zero buildTime). When the option is absent, go-erofs v0.3.1 stamps `time.Now().Unix()` into the superblock in `Writer.Close`, so a caller who passes a zero `time.Time` through `WithSourceDateEpoch` silently gets a different digest on every build. The CLI is protected because `options.Default` sets SDE to `time.Unix(0,0)` (which is not `IsZero`), but library callers hit nondeterminism exactly where they'd least expect it. Same gate in splitErofsLayers (erofs_layers.go:64).

pkg/build/erofs.go
```diff
@@ -0,0 +1,287 @@
...
+			dev, err := fsys.Readnod(fsysPath)
+			if err != nil {
+				return fmt.Errorf("readnod %s: %w", fsysPath, err)
+			}
+			rdev = uint32(dev)
+		}
+		if err := w.Mknod(absPath, typeBits|uint16(mode.Perm()), rdev); err != nil {
+			return fmt.Errorf("mknod %s: %w", absPath, err)
+		}
+	case mode.IsRegular():
```
mattmoor, line 152: Hardlinks are silently exploded into independent full copies here. tarfs surfaces a hardlink as a second dirent whose linkness is only visible via `Sys()`'s `tar.Header` (TypeLink/Linkname), so this branch sees a regular file with the target's size and copies the data again; go-erofs v0.3.1 has no Link API to alias two paths to one NID. The tar layer path preserves them via `tar.FileInfoHeader`. Result: st_nlink/st_ino identity is lost and hardlink-heavy packages inflate the image. Spec §8.1 permits materializing links, so this may be an acceptable v1 tradeoff, but it deserves a comment here and a line in docs/erofs.md, ideally with a size measurement on a hardlink-heavy image. Is a Link API on the go-erofs roadmap?

pkg/build/erofs.go
```diff
@@ -0,0 +1,287 @@
...
+	default:
+		return fmt.Errorf("unsupported file mode for %s: %v", absPath, mode)
+	}
+
+	// Mkdir, Mknod and Create all take only the permission bits, so
+	// setuid/setgid/sticky have to be applied on top — losing them would
+	// silently break su/passwd/mount and unprotect /tmp. Symlinks are
+	// exempt: EROFS pins them at 0777 and chmod on one is meaningless.
+	if mode&fs.ModeSymlink == 0 {
+		if err := w.Chmod(absPath, mode); err != nil {
```
mattmoor, line 208: While pinned to v0.3.1 this workaround deserves a sharper warning: erofs/go-erofs#41 (Mkdir dropping setuid/setgid/sticky on write, and `FileInfo.Mode()` misreporting them on read) merged upstream 2026-08-02, but v0.3.1 was tagged 2026-07-21 and does not contain it. This Chmod covers the write side, and `ls` reads modes from `*erofs.Stat` rather than `FileInfo.Mode()`, so apko is correct today, but nothing in-tree records that `FileInfo.Mode()` from the pinned reader is known-wrong. The next contributor who calls it gets bitten invisibly. Suggest noting it here and planning a bump once upstream tags a release containing #41, so the workaround stops being load-bearing.

pkg/build/build.go
```diff
@@ -185,8 +185,29 @@ func (bc *Context) ImageLayoutToLayer(ctx context.Context) (string, v1.Layer, er
...
 	bc.o.TarballPath = outfile.Name()
-	defer outfile.Close()
 
+	if bc.ic.Format.Resolved() == types.LayerFormatErofs {
+		// The EROFS path does not defer Close: it hashes the finished file, so
+		// the close has to happen first and its error has to be reported
+		// rather than swallowed by a defer. No Sync is needed either —
+		// *os.File does no userspace buffering, so once Close returns, a fresh
+		// Open sees every byte.
+		outName := outfile.Name()
```
mattmoor, line 195: Nit: with `--format=erofs` and no explicit output path, the EROFS blob lands at `TempDir()/apko-<arch>.tar.gz` (TarballFileName's spelling). Worth an `.erofs` naming path for anything that sniffs by extension.

pkg/build/oci/image.go
```diff
@@ -184,6 +194,15 @@ func BuildImageFromLayers(ctx context.Context, baseImage v1.Image, layers []v1.L
 		cfg.Config.StopSignal = ic.StopSignal
 	}
 
+	// Signal EROFS-bearing manifests via os.features per
+	// erofs/erofs-image-spec §5.4 so hosts that don't implement the spec can
+	// identify and skip them without parsing layer bytes.
+	if ic.Format.Resolved() == types.LayerFormatErofs {
```
mattmoor, line 202: §5.4 is a MUST in two places: the image config's `os.features` (done here) and the index platform descriptor's `os.features` when the image is referenced from an index (reinforced by §6 and §8.1 producer requirement 8). apko always emits an index, but `ToOCIPlatform` (pkg/build/types/types.go:335) never populates `OSFeatures`, so the index half of the MUST is unimplemented while this comment cites §5.4 as satisfied. Needs plumbing, since ToOCIPlatform has no config context today, but consumers that filter on the index platform without fetching configs won't see the signal until it exists.

pkg/erofsmount/oci.go
```diff
@@ -0,0 +1,208 @@
...
+		refs = append(refs, LayerRef{
+			BlobPath:    blob,
+			Digest:      desc.Digest.String(),
+			MediaType:   string(desc.MediaType),
+			Annotations: desc.Annotations,
+			Role:        desc.Annotations[types.ErofsRoleAnnotation],
+		})
+	}
+
+	// Validate role placement: per the EROFS image spec rule, every layer
```
mattmoor: Two spec-valid shapes get rejected below, and this comment attributes the stricter rule to the spec. Per the merged spec, `org.erofs.role` is OPTIONAL on any layer (§2.3 annotations table; §7 step 1 classifies "role `overlay-lower` **or absent**" as an overlay lower at any position), and §3.8 rule 1 says the final layer's role is "absent **or** `overlay-lower`". So (a) a non-final layer with no role and (b) a final layer carrying `overlay-lower` are both conformant, and both error out here. If the intent is "accept only apko-produced layouts," that's a defensible v1 posture, but then the error text and the spec attributions (here, types/erofs.go:28-30, erofs_layers.go:240, docs/erofs.md) should say it's apko's convention. Otherwise this should accept role-absent non-final layers and overlay-lower finals.

pkg/erofsmount/oci.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, m := range images {
+		if m.Platform == nil || m.Platform.Architecture == "" || m.Platform.Architecture == arch {
+			matches = append(matches, m)
+		}
+	}
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("no manifest for arch=%q (saw: %s)", arch, availableArchList(images))
+	}
+	if len(matches) > 1 {
+		return nil, fmt.Errorf("multiple manifests match arch=%q; pass --tag to disambiguate", arch)
```
mattmoor: There is no `--tag` flag on any `apko erofs` subcommand; the tag is given as a `SOURCE:TAG` suffix. Reword to point at the actual syntax.

pkg/erofsmount/stack.go
```diff
@@ -0,0 +1,394 @@
...
+	"sort"
+	"strings"
+	"time"
+)
+
+// Stack presents N fs.FS layers as a single fs.FS using AUFS-style overlay
+// semantics. Layers are stored bottom-up: layers[0] is the base, the last
+// element is the topmost. Topmost-wins is the rule for lookups.
+//
+// Whiteout encoding (matches OCI tar layers and go-erofs Writer.Merge):
```
mattmoor: This whiteout encoding is the opposite of what the spec mandates for EROFS layers: §3.6 requires overlayfs-native whiteouts (char device 0:0, `trusted.overlay.opaque="y"`) and §8.1 item 9 says EROFS images MUST NOT contain `.wh.`-prefixed names. Since `mount` composes with kernel overlayfs, `ls` and `mount` will disagree on any spec-conformant image containing deletions: `ls` shows the whiteout chardevs as live entries and hides nothing, while the kernel hides the lower entries; conversely `ls` honors `.wh.` names the kernel would show literally. Latent for apko-produced images (the writer emits no whiteouts), but wrong-by-construction for other producers' images, and docs/erofs.md:165 claims `ls` presents "the merged view the kernel would assemble". Suggest implementing §3.6 semantics here (keeping `.wh.` handling as a documented extra for go-erofs `Writer.Merge` output if desired), or at minimum softening the doc claim.

pkg/erofsmount/mount_linux.go
```diff
@@ -0,0 +1,305 @@
...
+// (blob) mountpoint and runs umount/fusermount.
+func Unmount(ctx context.Context, dest string) error {
+	log := clog.FromContext(ctx)
+	absDest, err := filepath.Abs(dest)
+	if err != nil {
+		return fmt.Errorf("resolve dest: %w", err)
+	}
+	dest = filepath.Clean(absDest)
+
+	st, err := LoadState(dest)
```
mattmoor: Unmount executes `umount`, typically as root, on whatever paths this file names, with no validation: `LoadState` checks JSON shape and schema version only, and `unmountImage` passes `st.Mounts` straight to the driver. Anyone who can write to dest (and the docs suggest shared locations like `/mnt/x`) can plant a state file naming `/home` or `/var` and have root unmount them. Cheap containment: verify `st.Dest == dest` and reject any `Mounts` entry that escapes dest under `filepath.Rel`; every path the mount side writes is under dest (`merged`, `layers/NN`), so nothing legitimate is lost.

pkg/erofsmount/mount_linux.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	// st.Mounts is overlay-first then per-layer mounts in LIFO order. If
+	// any umount fails, stop: layer mounts that come after a still-pinned
+	// overlay would only return EBUSY noise, and continuing past an error
+	// would also leave the state file out of sync with reality. The user
+	// can rerun `apko erofs umount` after addressing whatever is keeping
+	// the mount busy.
+	for _, mp := range st.Mounts {
+		if err := unmountOne(ctx, drv, mp); err != nil {
+			return fmt.Errorf("umount %s: %w (remaining mounts left intact; rerun once they are no longer busy)", mp, err)
```
mattmoor: This rerun instruction cannot succeed. `st.Mounts` is `[merged, layers/NN...00]`; if `merged` unmounts and a layer then fails EBUSY, the rerun's first step is `umount merged`, which now fails ("not mounted", exit 32) and returns before ever reaching the busy layer. The state file is never rewritten on partial success and nothing treats not-mounted as success, so recovery is manual umount plus hand-deleting the state file. Either tolerate a not-mounted result per entry or drop successfully-unmounted entries from the state file as you go.

pkg/erofsmount/mount_linux.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+		log.Infof("unmounted %s", mp)
+	}
+	for _, sub := range []string{"merged", "upper", "work", "layers"} {
```
mattmoor: This silently discards everything written through the mount: the default mount is writable (no `--read-only`), and umount RemoveAlls `upper` with no flag or warning. If scratch semantics are intended, a sentence in the mount help text would prevent surprises; alternatively, defaulting to read-only fits the verification use case the docs describe.

pkg/build/erofs_test.go
```diff
@@ -0,0 +1,491 @@
...
+// optionalFsckErofs returns the path to fsck.erofs, or "" when erofs-utils is
+// not installed. It is deliberately optional: every caller has already parsed
+// the image with go-erofs, so fsck is a second opinion from the C reference
+// implementation rather than the only check, and contributors without
+// erofs-utils still get a green build. The log line keeps the reduced coverage
+// visible in test output instead of silently passing.
+//
+// Tests whose entire point is the erofs-utils cross-check skip explicitly
+// instead — see TestWriteErofs_FsckErofs.
+func optionalFsckErofs(t *testing.T) string {
```
mattmoor, line 381: Reasonable design, but note that in CI the second opinion never renders: no workflow installs erofs-utils, so every fsck cross-check degrades to this `t.Log` (invisible without `-v`, which `make test` doesn't pass), TestWriteErofs_FsckErofs skips outright, and the only always-on validation of written images is go-erofs reading its own output. A shared spec misunderstanding between writer and reader would sail through green. One `apt-get install erofs-utils` in go-tests.yaml (plus the harden-runner allowed-endpoints for it) would activate all the fsck legs; a small privileged job doing `apko build --format erofs` + kernel mount + fsck would close the loop end to end, and would also be the first thing anywhere to execute the mount orchestration.

pkg/build/erofs_layers.go
smoser: You're right, and it's worse than a wrong receiver: it can never succeed. Confirmed `memFileInfo.Sys()` returns a fresh `*tar.Header` (pkg/tarfs/fs.go:1012) while `Package()` is on the FileInfo (:1031), and `memFS.ReadDir` wraps with `fs.FileInfoToDirEntry`, whose `Info()` hands back that same FileInfo — so `info` is the right receiver, exactly as layers.go:323 does it.

Also confirmed your read of the consequence: `owner` stays `top` for everything, and the partial installed db still fires because it keys on the path string, so each group layer ends up claiming packages whose files aren't in it.

Moved to #2408 along with the rest of the split, with the `tarfs`-backed test as a prerequisite rather than an afterthought — a fix without it would be unverifiable for the same reason `TestSplitErofsLayers` can't catch the bug today. The silent `packageToWriter` fallback goes with it; it'll return an error rather than either falling back to `top` or panicking like the tar path.

pkg/build/erofs_layers.go
smoser: Confirmed. Dirs are recorded into `dirInfo` and materialized only by `emitAncestors`, which runs for non-dir entries, and there's no post-walk sweep — so every empty dir and dir-only chain is absent from every layer, `top` included.

Moved to #2408. Noting there that both siblings get this right, since that's the argument for which fix to pick.

pkg/build/erofs_layers.go
smoser: Confirmed, including the part that makes it awkward to fix cleanly: `Close` both finalizes the image and early-returns before its spool-close defer on a sticky error, so it can't serve as an abort.

Moved to #2408. A named-return cleanup is the fix that's available today; I've noted that an upstream Abort API is what would make it airtight, so this doesn't get filed as done-and-dusted.

pkg/build/erofs_layers.go
smoser: Fair, and it was the same conflation in four places. `erofs_layers.go` left with the split (#2408), and the copies in `types/erofs.go` and `docs/erofs.md` now say which part is the spec and which is apko's convention: §2.3 makes the annotation OPTIONAL on any layer, §3.8 rule 1 allows the final layer's role to be absent *or* `overlay-lower`, and "non-final carries `overlay-lower`, final carries none" is our own spec-legal choice of one shape.

The consumer side went the other way rather than documenting the strictness — see the reply on `ReadOCILayers`.

pkg/build/erofs.go
smoser: Fixed. The option is now always passed.

One wrinkle worth recording: it can't be passed verbatim. A zero `time.Time` has `Unix() == -62135596800`, which wraps through `uint64` into a huge value, so `erofsBuildTime` clamps zero and pre-epoch timestamps to epoch 0 — deterministic, and it matches what an unset `SOURCE_DATE_EPOCH` already means to apko.

My first test for this was vacuous and I nearly shipped it: go-erofs's default is `time.Now().Unix()`, second granularity, so two back-to-back zero-time builds agree anyway. It now pins the zero case against an explicit epoch build, which fails on the old code and passes on the new — I checked both directions.

pkg/build/erofs.go
smoser: Documented at the branch and in `docs/erofs.md`, with your framing that §3.7 makes it conformant rather than a bug.

On the roadmap question: no. There's no open issue or PR on go-erofs proposing a `Link` API. `SetNlink` exists but only sets the reported link count without sharing the inode, so using it would make `st_nlink` lie while saving nothing — worth knowing so nobody reaches for it as a fix.

No size measurement, though, and I'd rather say so than invent one: I had no hardlink-heavy image on hand, and the `coreutils` apk I checked has zero hardlink entries. The docs state the arithmetic instead — one full copy per link, rounded up to the block size, versus tar's zero-byte entries for extra links. Happy to add a real number if you have an image in mind.

pkg/build/erofs.go
smoser: Sharpened, with the dates, since they're the whole point: #41 merged 2026-08-02 and v0.3.1 was tagged 2026-07-21, so the pin has neither half of the fix.

The comment now says explicitly that the read half is unfixed too — `FileInfo.Mode()` from the pinned reader misreports setuid/setgid/sticky, and code that needs a mode must read `*erofs.Stat.Mode` off `Sys()`, as `ls.go` does. That was the invisible trap you identified: nothing in-tree recorded it.

Bump tracked in #2408, including re-checking whether this `Chmod` is still load-bearing afterwards.

pkg/build/build.go
smoser: Fixed. `Options.LayerFileName` takes the layer format and picks the extension, delegating to `TarballFileName` for tar so the two spellings can't drift; an explicitly given `TarballPath` is still honored verbatim.

pkg/build/oci/image.go
smoser: Fixed, and thanks for catching that the comment claimed §5.4 was satisfied while half of it wasn't.

`ToOCIPlatform` has no config context, as you say, but `generateIndexWithMediaType` already holds the `v1.Image`, so it reads `ConfigFile().OSFeatures` and copies it onto the descriptor's platform. Format-agnostic on purpose — `os.features` describes the image whatever produced it.

Verified on a real build: the index platform descriptor now carries `"os.features": ["erofs"]` alongside the config's. The test also pins that a tar build gains no `OSFeatures` field at all, since an empty-but-present slice would change every index digest.

pkg/erofsmount/oci.go
smoser: Fixed by accepting the conformant shapes rather than documenting the strictness — your first option. Role-absent non-final layers and `overlay-lower` finals are both fine now, at any position, per §2.3, §3.8 rule 1 and §7 step 1 (which treats absent and `overlay-lower` identically).

`device` and `overlay-data` are refused as compositions we don't implement rather than as malformed images, and an unrecognized value is refused as unknown. With those in place §3.8 rule 1 holds by construction, so the separate positional pass is gone.

`TestReadOCILayers_BadRoleOrder` asserted precisely the behavior you flagged, so it's now `TestReadOCILayers_SpecValidRoleShapes` over the three legal shapes.

pkg/erofsmount/oci.go
smoser: Fixed — it now says to name a tag as `SOURCE:TAG`, and lists the tags actually available in the layout.

pkg/erofsmount/stack.go
smoser: Fixed, and the justification I'd written into that comment turned out to be false, which is worth stating plainly: `Writer.Merge` does not produce `.wh.` names. It *consumes* them from its copy source and, in its own words, "the whiteout entries themselves are not added to the image" (mkfs.go:118, with the elimination at mkfs.go:444 and mkfs_image.go:108). So no producer emits `.wh.` inside an EROFS image — not apko, not go-erofs — and the reader was honoring an encoding that §8.1 item 9 forbids and that cannot conformantly exist.

Now implements §3.6: a whiteout is a char device with rdev 0, an opaque directory carries `trusted.overlay.opaque="y"`. Detection goes through the accessor interfaces go-erofs documents on its FileInfo (`Rdev()`, `GetXattr()`) rather than asserting `*erofs.Stat`, so it stays fs.FS-generic, and a cheap `DirEntry.Type()` check gates the `Info()` call so only char devices cost an inode read.

Two things I'd flag as judgment calls:

- Whiteouts are interpreted only when the stack has 2+ layers. §7 step 6 lets a lone EROFS layer be mounted directly as a root filesystem, and a direct mount applies no overlay semantics, so a rdev-0 char device in a single-layer image is reported as the device it is.
- `.wh.` handling is gone rather than kept as an extra. Since nothing conformant can contain such a name, dropping it makes `ls` *more* faithful on malformed input too: the kernel would show `.wh.foo` as a literal filename, and now so do we. §3.6 also says consumers must not fail on meaningless whiteouts.

`mergeDir` collapsed as a result — a whiteout occupies the name it deletes, so one pass replaces the separate live/tombstone maps, and a layer can no longer hold both an entry and its own tombstone.

Tests needed rdev and xattrs, which `fstest.MapFS` can't express, so there's a small wrapper exposing the same accessors. Plus `TestStack_Whiteouts_RealErofsLayers`, which runs the rules against images go-erofs actually wrote — that's the one that catches the fixture and the real reader drifting apart, and it confirms both that a rdev-0 `Mknod` reads back as a whiteout and that `Setxattr` of `trusted.overlay.opaque` round-trips.

`docs/erofs.md`'s "the merged view the kernel would assemble" claim is now true rather than softened.

pkg/erofsmount/mount_linux.go
smoser: Agreed, and this is the one I'd least want to ship as-is. Moved to #2408 with your containment recipe recorded verbatim, since it's exactly right: verify `st.Dest == dest` and reject any `Mounts` entry that escapes `dest` under `filepath.Rel`. Everything the mount side writes is under `dest`, so nothing legitimate is lost.

pkg/erofsmount/mount_linux.go
smoser: Confirmed — `merged` is first in `Mounts`, so the rerun's opening move is a `umount` of something already unmounted, which exits 32 before it ever reaches the busy layer. And since state is never rewritten on partial success, recovery really was manual `umount` plus deleting the state file by hand.

Moved to #2408 with both of your options noted: tolerate a not-mounted result per entry, or drop unmounted entries from the state file as it goes.

pkg/erofsmount/mount_linux.go
smoser: Moved to #2408, with the decision made: it'll default to read-only, which as you say fits the verification use case the docs describe. Writes become an opt-in rather than the default that silently discards them.

pkg/build/erofs_test.go
smoser: Fixed — `erofs-utils` is installed in the Go Tests job, which activates every fsck leg and unskips `TestWriteErofs_FsckErofs`.

Two notes. The step asserts `fsck.erofs` reached `PATH` rather than trusting apt's exit code, because a silent install failure would look exactly like a pass — which is the same failure mode you're describing. And this job had no apt step before, so there was no precedent for which Ubuntu mirror the runner resolves; I allowlisted `archive`, `azure.archive` and `security` in harden-runner, and CI will tell us which one it actually wanted.

The privileged build + kernel mount + fsck job is in #2408, filed against the mount follow-up since that's where the code it would exercise now lives.

smoser:
Thanks for the depth here, @mattmoor — the two multi-layer findings were both
real and neither was reachable by the existing tests, which is the worst
combination. Took the descope you suggested: this PR is now the single-image
writer plus `apko erofs ls`, with layering and mount/umount tracked in #2408 and
landing in that order. The pre-descope tree is preserved on
`feat/apko-erofs-full` in my fork, so nothing gets reconstructed from memory.

Per-comment replies are inline. The three cross-cutting notes:

**Exported surface.** All of `StatePath`/`WriteState`/`LoadState`/`RemoveState`/
`StateSchemaVersion` and `Driver`/`NewDriver`/`ResolveMode` left with the mount
descope, and `erofsLayer.LayerPath()` is dropped. A knock-on benefit worth
naming: with the Linux-only files gone the package has no build-tag-dependent API
at all — everything left is pure Go and builds everywhere, which was the part of
that note I liked least about the old shape. `Ls` also lost the `Options` bag it
only ever read `Arch` out of and now takes a plain `arch string`, so `Mode` and
`Options` can come back package-private when mount does. Noted in #2408.

**Test seam for the mount plane.** Agreed, and it's a prerequisite in #2408 rather
than a nice-to-have: a `Driver` injection point, so the cleanup LIFO, state
lifecycle and unmount policy are exercised by something. Right now nothing
anywhere runs them, which is how two bugs sat in 305 lines undisturbed.

**melange coordination.** Not done yet — I'd rather not comment on melange#2605
until this lands and I know what it's re-pinning to. It's recorded in #2408 under
the compression item so it doesn't get lost: melange passes
`erofs+zstd[,level=N]` strings that current apko rejects after the #2406 split, so
its compressed mode is dead until #2406 merges and melange re-pins off my fork.

Also worth flagging two things you didn't ask about but that changed:

- Your `WithBuildTime` suggestion couldn't be taken literally. Passing a zero
  `time.Time` through `uint64` wraps (`Unix()` is −62135596800), so zero and
  pre-epoch clamp to epoch 0 instead. And my first test for it was vacuous —
  go-erofs's default has one-second granularity, so back-to-back builds match
  regardless. It now pins zero against an explicit epoch build.
- The `.wh.` comment's claim that the encoding "matches go-erofs `Writer.Merge`"
  was simply wrong, which removed the only argument for keeping it. `Merge`
  consumes `.wh.` from its source and never writes such names into an image, so I
  deleted the tar handling outright instead of keeping it as a documented extra.

Verified before pushing: `go test ./...` green, `golangci-lint run -n` clean, and
a real `wolfi-base` build with `--format=erofs` produces `os.features: ["erofs"]`
on both the index platform descriptor and the config — the §5.4 half that was
missing — with `apko erofs ls` reading it back with sticky bits and device numbers
intact.

pkg/build/erofs_layers.go
mattmoor: Tracked in #2408 with the tarfs-backed test as a prerequisite and the fallback-to-error decision recorded. Resolving here.

pkg/build/erofs_layers.go
mattmoor: Tracked in #2408. Resolving here.

pkg/build/erofs_layers.go
mattmoor: Tracked in #2408, including the note that `Close` cannot serve as an abort. Resolving here.

pkg/build/erofs_layers.go
mattmoor: The in-PR copies are fixed at 32214b1f: `types/erofs.go` and `docs/erofs.md` now separate the spec text (§2.3 optional, §3.8 rule 1 absent-or-`overlay-lower`) from apko's convention. The `erofs_layers.go` copy left with the descope and is covered by #2408. Resolving.

pkg/build/erofs.go
mattmoor: Verified at 32214b1f: the option is always passed, the clamp handles zero and pre-epoch times (including the sub-second normalization case, where Go yields `sec=-1`), and the new test discriminates against the old behavior where a back-to-back comparison would not. Resolving.

pkg/build/erofs.go
mattmoor: Verified at 32214b1f: the caveat is at the writer branch and in docs/erofs.md with the copy-per-link arithmetic, and the `SetNlink` warning is stated so nobody reaches for it as a fix. One wording note, no action needed: §3.7's materialize-or-fail rule governs cross-layer hardlinks; same-layer materialization is conformant because the spec is silent on it, not because §3.7 permits it. Resolving.

pkg/build/erofs.go
mattmoor: Verified at 32214b1f: the dates and the read-half warning are in the comment, `ls.go` reads modes via `*erofs.Stat` as the comment directs, and the bump plus the Chmod re-check are tracked in #2408. Resolving.

pkg/build/build.go
mattmoor: Verified at 32214b1f: `LayerFileName` delegates to `TarballFileName` for tar so the spellings cannot drift, the test pins it, and no other reader of `TarballPath` assumes the tar spelling. Resolving.

pkg/build/oci/image.go
mattmoor: Verified at 32214b1f: the descriptor carries `os.features`, the copy is cloned so nothing aliases the config, and vanilla tar builds are unaffected (nil stays nil).

One knock-on worth an explicit ack, not a code change: `BuildImageFromLayers` DeepCopies the base image's config, so a tar build on top of a base whose config already carries `os.features` now surfaces it on the index platform descriptor, which changes that index's digest. Arguably that is exactly what §5.4 wants, but it is a digest-visible change to a non-erofs flow, and the new test pins only the empty-base case. A test pinning the base-image case (or a release-note sentence) would close it.

Minor, for accuracy: `require.Empty` also passes for an empty non-nil slice, and `os.features` carries `omitempty` in go-containerregistry, so an empty-but-present slice could not change a digest anyway. The `len > 0` guard is good hygiene; the test just does not distinguish nil from empty.

pkg/erofsmount/oci.go
mattmoor: Verified at 32214b1f against the spec text: role-absent non-final layers and `overlay-lower` finals are accepted, `device`/`overlay-data` are refused as unimplemented rather than malformed, and the by-construction §3.8 comment holds since every layer already passed the mediaType and role checks. Nit, take or leave: a spec-legal `application/vnd.erofs+zstd` layer hits the mediaType check first and gets "only handles EROFS images", which reads oddly for a compressed EROFS image. Resolving.

pkg/erofsmount/oci.go
mattmoor: Verified at 32214b1f: `SOURCE:TAG` plus the available-tags listing. Resolving.

pkg/erofsmount/stack.go
mattmoor: Verified at 32214b1f, including the go-erofs claim: `Merge` consumes `.wh.` markers and never writes them (mkfs.go:441-455, mkfs_image.go:105-120). One nuance for the record: `Merge` flattens rather than translates. Deletions are applied at build time and no whiteout representation survives into the written image, so nothing in go-erofs produces the whiteouts Stack decodes. That does not change the conclusion here; apko never emits them either, and the real-layer test writes them explicitly via `Mknod`/`Setxattr`.

The 2+ layer gate, the rdev-0 and opaque detection through the accessor interfaces, and the literal `.wh.` behavior all match the spec text, and `TestStack_Whiteouts_RealErofsLayers` pins the fixture wrapper and the real reader together, which was the drift risk in having a fixture at all.

Two pre-existing corners where the merge algorithm still diverges from kernel overlayfs, so "the merged view the kernel would assemble" remains slightly stronger than what is delivered:

1. A middle-layer whiteout (or plain file) at a directory's own name does not cut off lower layers when a higher layer recreates the directory. Given L0 `d/x`, L1 whiteout on `d`, L2 dir `d/y`, the kernel shows `d` containing only `y`; Stack shows `x` and `y`. `lookup` and `mergeDir` treat a `readDirOn` failure as "parent absent in this layer", which is indistinguishable from "parent is a non-directory or tombstone here".
2. Opacity is not inherited by descendant directories. With top-layer `a` opaque containing `a/b`, and lower-layer `a/b/c`, the kernel hides `c` (the cut covers the whole subtree); Stack lists it.

Both predate this PR (the `.wh.` code had the same shape), both are unreachable for apko-produced images (single-layer post-descope), and neither blocks landing. Suggest either a known-deviations sentence in docs/erofs.md or an item in #2408 next to the layered-whiteout one; a per-directory layer horizon computed during descent would fix both. Related and smaller: `isOpaqueDir` swallows `statOn` errors, so a corrupt xattr region silently un-hides content instead of surfacing the error; `isWhiteout`'s equivalent is partially self-correcting since the leaked entry's own Stat re-raises.

pkg/erofsmount/mount_linux.go
mattmoor: Tracked in #2408 with the containment recipe recorded. Resolving here; this stays the first thing to land in the mount follow-up.

pkg/erofsmount/mount_linux.go
mattmoor: Tracked in #2408 with both options recorded. Resolving here.

pkg/erofsmount/mount_linux.go
mattmoor: Tracked in #2408 with the read-only-default decision made. Resolving here.

pkg/build/erofs_test.go
mattmoor: Verified at 32214b1f: the install step asserts `fsck.erofs` reached PATH under `bash -e`, so a silent apt failure fails the job, and the cross-checks paid for themselves immediately by catching the `--xattrs` portability issue. The privileged build-mount-fsck job is rightly filed with #2408. Resolving.

mattmoor (approved):
Approving the descoped scope. All in-scope fixes from the review verified at 32214b1f: always-stamped build time with the pre-epoch clamp, `os.features` on both the config and the index platform descriptor, spec-legal role shapes accepted, the §3.6 whiteout encoding in Stack pinned against real go-erofs layers, the `.erofs` default name, the documented hardlink and go-erofs v0.3.1 caveats, and erofs-utils running the fsck cross-checks in CI (which caught the `--xattrs` portability issue on their first run, a good sign the check earns its keep). The descope is clean: no dangling references to the mount plane or the splitter anywhere, and #2408 accurately tracks the deferred findings.

Two non-blocking notes left on reopened threads: the index digest change for tar builds atop base images whose config carries `os.features` deserves an explicit ack or a pinning test, and the `ls` merged view has two pre-existing kernel-overlayfs divergences worth a docs sentence or a #2408 item.

pkg/build/oci/image.go
smoser: Acked explicitly in the squash message, which now spells out that `os.features` is copied for every format, that a tar build on a base image carrying it will see its index digest change, and that it is a no-op for a base declaring no features. Since we squash-merge, that is the text that lands in main's history.

The pinning test is filed in #2408 rather than added here — you're right that it's the stronger closure, and it wants to go in with the `require.Nil` change.

On that: thanks for the `omitempty` correction. It also means the justification in that commit's message — that an empty-but-present slice "would change every index digest" — was wrong. It can't. That claim exists only in the commit body and nowhere in the tree, so writing our own squash message retires it instead of it being concatenated into main by the default squash body. Worth knowing that was the actual mechanism, since the `len > 0` guard I wrote it to defend is hygiene, not a digest fix.

pkg/erofsmount/stack.go
smoser: Filed in #2408 as its own section, next to the layered-whiteout item as you suggested, and I reproduced both against the merged code before filing so the tracking item has the concrete shapes rather than a description:

| Layers | Stack | kernel |
|---|---|---|
| L0 `d/x`; L1 whiteout on `d`; L2 dir `d/y` | `d` = `[x y]` | `d` = `[y]` |
| top `a` opaque with `a/b`; lower `a/b/c` | `a/b` = `[c]` | `a/b` = `[]` |

Your single-root-cause reading holds: both come from `readDirOn` failure being indistinguishable from "parent is a tombstone or a non-directory in this layer", and the per-directory layer horizon is filed as the fix for both. `isOpaqueDir` swallowing `statOn` errors is filed next to it for the same reason — it's the same absent-versus-broken confusion, so the two probably want fixing together.

The docs deviation note is filed as part of that item rather than done here, so I should be clear about the consequence of landing as-is: `docs/erofs.md` currently claims a merged view the kernel would assemble, and that stays slightly stronger than what ships until the horizon work lands. Flagging it plainly rather than leaving it implied.

Also noted your `Merge` refinement — flattens rather than translates, so no whiteout representation survives into the written image at all. That strengthens the conclusion rather than changing it: there is no producer path anywhere, not merely no apko path, that emits what the old `.wh.` code decoded.

pkg/build/oci/image.go
mattmoor: Addressed in #2410: the base-image case is pinned by TestGenerateIndex_PropagatesOSFeaturesFromBaseImage and the tar assertion strengthened to require.Nil. Resolving.

pkg/erofsmount/stack.go
mattmoor: Addressed in #2410: docs now describe the two divergences and link #2408 (which tracks the layer-horizon fix), and isOpaqueDir fails closed with a test placed to discriminate. isWhiteout symmetry noted on that PR, take or leave. Resolving.

PR #2269, opened 2026-06-09
expandapk: materialize uncompressed .dat.tar atomically

### LINKED ISSUE #2268: Concurrent use of package cache may end in "unexpected EOF"
Similar to #1903, we are seeing occasional "unexpected EOF" errors during image build:

```
2026/06/07 04:39:36 DEBU buildImage failed: installing apk packages: installing packages: expanding libselinux-python (ver:3.10-r0 arch:aarch64): unexpected EOF
```

However, for these, there is no retry and no indication of networking problems.

Some instrumentation of apko has instead shown that concurrent use of the package cache by multiple apko processes can result in an interleaving that has one apko process use a truncated .tar file. Consider:

1. Two concurrent apko builds (A and B) with shared cache start slightly staggered
1. A loads package 1 into the cache, saves it as a .tar.gz
1. B find package 1 in the cache, but only in its compressed form and so it starts decompressing it into the .tar
1. A sees the decompressed .tar file in the cache (but fails to see/cannot see it is not yet fully decompressed yet)
1. A reads the incomplete decompressed .tar file (and fails with "unexpected EOF") or reads the empty .tar (and no failure is flagged, but no files are copied into the image either!)
1. B finishes decompressing
1. B installs package 1 successfully

The chances of hitting this without instrumentation to widen the race window are very slim (in my local test I needed ~50 processes to get a failure).

apko should only make the decompressed .tar file available in the package cache once it's been completely decompressed.

markusboehme:
PackageData() decompresses the cached .dat.tar.gz into the uncompressed .dat.tar by creating that file in place (os.Create) and streaming the decompressed bytes into it. .dat.tar is a content-addressed path that can be shared by multiple apko processes through a common --cache-dir (e.g. parallel image builds on one runner). A concurrent process can therefore read the file while it is still being written and cause one of these problems:

- reading it mid-stream fails the build with "installing packages: expanding <pkg>: unexpected EOF"
- reading it at zero length yields a valid empty tar, so the build succeeds but the package contributes no files while the apk database still records it as installed

Add a writeFileAtomic helper that streams into a temporary file in the same directory and atomically renames it into place, and use it from PackageData. Readers then observe either no file (and materialize their own) or the complete file. The temporary file is created alongside the destination so the rename stays on one filesystem.

Add deterministic tests for writeFileAtomic covering the atomicity invariant (the destination is never observable mid-write) and the error path (a failed write leaves no destination and no leftover temp).

Fixes: #2268

### DIFF
--- a/pkg/apk/expandapk/atomic_test.go
+++ b/pkg/apk/expandapk/atomic_test.go
@@ -1,0 +1,114 @@
+package expandapk
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// midWriteChecker is an io.Reader that, after the first chunk it returns has
+// been written by the copier, asserts that the destination path does not yet
+// exist. If the write were done in place (os.Create on the destination), the
+// destination would already hold a partial file at this point.
+type midWriteChecker struct {
+	t       *testing.T
+	dst     string
+	data    []byte
+	off     int
+	checked bool
+}
+
+func (r *midWriteChecker) Read(p []byte) (int, error) {
+	if r.off > 0 && !r.checked {
+		r.checked = true
+		if _, err := os.Stat(r.dst); !errors.Is(err, fs.ErrNotExist) {
+			r.t.Errorf("destination %q is observable mid-write (stat err=%v); write is not atomic", r.dst, err)
+		}
+	}
+	if r.off >= len(r.data) {
+		return 0, io.EOF
+	}
+	// Cap each Read at 64 KiB to force several Read calls, so a chunk is
+	// flushed before the next destination check.
+	n := min(copy(p, r.data[r.off:]), 64*1024)
+	r.off += n
+	return n, nil
+}
+
+func TestWriteFileAtomic_DestinationNotVisibleUntilComplete(t *testing.T) {
+	dir := t.TempDir()
+	dst := filepath.Join(dir, "data.tar")
+
+	payload := bytes.Repeat([]byte("x"), 256*1024)
+	r := &midWriteChecker{t: t, dst: dst, data: payload}
+
+	if err := writeFileAtomic(dst, r); err != nil {
+		t.Fatalf("writeFileAtomic: %v", err)
+	}
+	if !r.checked {
+		t.Fatal("mid-write check never ran; the test would not catch a regression")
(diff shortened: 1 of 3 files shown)


pkg/apk/expandapk/atomic_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	return 0, r.err
+}
+
+func TestWriteFileAtomic_FailedWriteLeavesNoDestination(t *testing.T) {
+	dir := t.TempDir()
+	dst := filepath.Join(dir, "data.tar")
+
+	payload := bytes.Repeat([]byte("y"), 256*1024)
+	boom := errors.New("boom")
```
a-crate: This is kind of vague. `errors.New("already read")`?

pkg/apk/expandapk/expandapk.go
```diff
@@ -186,29 +186,19 @@ func (a *APKExpanded) PackageData() (*os.File, error) {
 		return nil, fmt.Errorf("parsing %q: %w", a.PackageFile, err)
 	}
 
-	uf, err = os.Create(a.TarFile)
-	if err != nil {
-		return nil, fmt.Errorf("opening tar file %q: %w", a.TarFile, err)
-	}
-
-	buf := pooledSlice()
```
a-crate, line 194: This is the only caller of `pooledSlice()` I can find. If we're not going to use it here it's basically just dead code we can remove.

markusboehme: The re-use of slices should not be needed here, as [the documentation](https://pkg.go.dev/io#CopyBuffer) for `io.CopyBuffer(dst, src, buf)` says:

> If either src implements [WriterTo](https://pkg.go.dev/io#WriterTo) or dst implements [ReaderFrom](https://pkg.go.dev/io#ReaderFrom), buf will not be used to perform the copy.

Since `os.CreateTemp` returns an `os.File` and that type implements `io.ReaderFrom`, the pooled slice goes to waste.

I'll remove the implementation, thanks. `//nolint:all` at the top of `expandapk.go` had that not pop up in linting.

pkg/apk/expandapk/atomic_test.go
markusboehme: Fair! :-)

a-crate (approved):

PR #2271, opened 2026-06-10
Match apk-tools' provider comparison ordering in the solver

iainlane:
When an index holds several builds of the same package, the solver compares their provider priorities before their versions. Priorities can vary across versions. If a priority assignment is lowered between releases, an old build can permanently outrank every newer one.

The wolfi index currently hits exactly this: `py3.10-pybind11` 2.13.6-r1 is a superseded package published with priority 312. The current version 3.0.4-r0 has priority 310. Direct dependencies on `py3.10-pybind11` are resolving to the older version.

`apk` resolves things differently. Its solver selects the latest by requested name, then the latest by principal name, and only then prefers the declared provider priority, falling back to the lowest available repository:

  https://github.com/alpinelinux/apk-tools/blob/20fe3dccc423bd401b9828958124664704dedb00/src/solver.c#L641-L688

Priorities therefore only come into play on the same package if their versions tie. An unversioned provide carries no version and ties with anything, which is how different providers of a virtual name still resolve by priority.

Reorder `comparePackages` to follow that sequence: the version each candidate carries for the requested name, then package versions between builds of the same name, then provider priority, then repository order, which `PkgResolver` now tracks from the configured index order.

One deliberate divergence remains, as the previous behaviour is explicitly tested for in `TestProviderAcrossOriginVersionBump`: when apk-tools has no preference left (same repository, all else tied), prefer the higher package version so that a stale build lingering in the index cannot win the tie.

The `apk` resolution can be reproduced in a Wolfi image:

```console
$ docker run --rm cgr.dev/chainguard/wolfi-base sh -ec '
    apk update --quiet
    tar -xzOf /var/cache/apk/APKINDEX.*.tar.gz APKINDEX |
      awk -v RS="" "/\nP:py3\.10-pybind11\nV:(2\.13\.6-r1|3\.0\.4-r0)\n/" |
      grep -E "^(P|V|k):"
    apk add --simulate py3.10-pybind11 | grep pybind11'
P:py3.10-pybind11
V:2.13.6-r1
k:312
P:py3.10-pybind11
V:3.0.4-r0
k:310
(15/15) Installing py3.10-pybind11 (3.0.4-r0)
```

...upstream `apk` installs the current build even though the stale one declares a higher priority.

Other orderings were verified the same way with apk-tools 2.14.10 against synthetic indexes: a versioned provide beats a higher-priority provider of a lower version, unversioned providers resolve by priority, and providers tying on priority resolve in favour of the repository listed first. Each case is covered by a test.

### DIFF
--- a/pkg/apk/apk/repo.go
+++ b/pkg/apk/apk/repo.go
@@ -1012,7 +1012,47 @@ func (p *PkgResolver) comparePackages(compare *RepositoryPackage, name string, e
 			return 1
 		}
 
-		// check provider priority
+		// The remaining steps follow apk-tools' compare_providers ordering
+		// (latest by requested name, then latest by principal name, then the
+		// highest declared provider priority), with one deliberate divergence
+		// in the final tiebreak below.
+		// https://github.com/alpinelinux/apk-tools/blob/20fe3dccc423bd401b9828958124664704dedb00/src/solver.c#L641-L688
+
+		// Latest by requested name: compare the versions the candidates carry
+		// for the name being resolved. An unversioned provide carries no
+		// version for the name and ties with anything.
+		if iVersionStr != "" && jVersionStr != "" && iVersionStr != jVersionStr {
+			iVersion, err := cachedParseVersion(iVersionStr)
+			if err != nil {
+				return 1
+			}
+			jVersion, err := cachedParseVersion(jVersionStr)
+			if err != nil {
+				// If j fails to parse, prefer i.
+				return -1
+			}
+			if versions := CompareVersions(iVersion, jVersion); versions != equal {
+				return -1 * versions
+			}
+		}
+
+		// Latest by principal name.
+		if a.Name == b.Name && a.Version != b.Version {
+			iVersion, err := cachedParseVersion(a.Version)
+			if err != nil {
+				return 1
+			}
+			jVersion, err := cachedParseVersion(b.Version)
+			if err != nil {
+				// If j fails to parse, prefer i.
+				return -1
+			}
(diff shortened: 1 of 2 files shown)


xnox:
Yes, we want this.

To test, build art and relock all images-private; commit; build art with this cuatomized apko; relock and commit - and observe the net diff of everything we ship.

And please report back if the change has correct impact or not.

Note some images might fail to lock, or will lock to different versions. Please inspect if that new outcome is the desired one.

iainlane:
The two `build-all-examples-amd64` failures are upstream Alpine index skew, not this change: the cross-arch lock of `examples/alpine-base.yaml` needs one `libcrypto3` version present on both architectures, but x86_64 currently publishes 3.5.7-r0 while aarch64 is still on 3.5.6-r0. An apko built from `main` fails identically with the same command against the current indexes. The jobs should pass on rerun once the aarch64 index catches up.

iainlane:
Thanks for the hint. I ran the test against the current repo both with and without this fix.

The branch as first pushed regressed: apk-tools' "prefer the lowest repository" tiebreak let variant builds in earlier-listed repositories capture sonames from later ones. 6 images stopped resolving and ~75 silently picked up variant builds or downgraded providers. I've dropped that step, restoring the existing divergence to prefer the higher build version again, which we rely on, with a test to lock it in.

With that revision, relocking is identical to stock apko across the whole catalogue. The same four images fail to lock before and after, all pre-existing. The pybind11 fix doesn't change any locks; it shows up in direct resolution against the wolfi index as described in the commit message.

xnox (approved):
I always wanted to drop the silent/phantom version of virtual provides too. very good.

PR #2276, opened 2026-06-12
fix(auth): let chainctl write to terminal for interactive login

dannf:
If I'm not authenticated already, `melange` will hang when I start a build that requires auth. I can see that it has forked a `chainctl` process, which I run in `--headless` mode, but no device-login URL gets emitted to my terminal. This is because `apko` is discarding stderr.

Stream chainctl's stderr to the terminal when stderr is a TTY, so the login URL (and any errors) are visible. Keep discarding it in non-interactive contexts (CI/logs) to preserve the original intent.

Co-Authored-By: Claude Opus 4.8 (1M context) \<noreply@anthropic.com\>
Signed-off-by: dann frazier \<dann.frazier@chainguard.dev\>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	golang.org/x/oauth2 v0.36.0
 	golang.org/x/sync v0.20.0
 	golang.org/x/sys v0.46.0
+	golang.org/x/term v0.43.0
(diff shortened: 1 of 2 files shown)


EyeCantCU (approved):
Thank you, thank you, thank you

PR #2282, opened 2026-06-18
fix: fall back to GET when APKINDEX HEAD is unsupported

### LINKED ISSUE #2251: Package index fetches fail if APKINDEX.tar.gz doesn't support HEAD requests (e.g. Gitea/Forgejo package repositories)
Refiling https://github.com/chainguard-dev/go-apk/issues/203 here because I've now experienced this issue. 

Copy of original:
> The package index fetch code [errors](https://github.com/chainguard-dev/go-apk/blob/4a3a18e598d6c75907bc7aa199d9df7c8a86d974/pkg/apk/cache.go#L62) if a HEAD request to APKINDEX.tar.gz returns 405 Method Not Allowed, which means it isn't possible to use go-apk based tools (such as apko) with a Forgejo package repository.

> I think Forgejo/Gitea is probably in the wrong here, but it should be possible to gracefully handle this case.

apko error log:

> Error: locking config: resolving apk packages: for arch "amd64": error getting repository indexes: reading index https://.../alpine/v3.23/main/x86_64/APKINDEX.tar.gz: unexpected status code 405

Is there a standard that requires Alpine registries handle HEAD requests? Should this be instead solved in Forgejo?

arpitjain099:
Some Alpine registries reject HEAD requests for APKINDEX.tar.gz with 405, which currently makes index resolution fail before we ever fetch the index.

This change treats 405 and 501 from the HEAD probe as unsupported-method responses and falls back to fetching and parsing the index via GET without etag-based HEAD caching. Other non-200 statuses still fail as before.

I also added a regression test that simulates HEAD returning 405 while GET succeeds.

Fixes #2251

Verification:
- go test ./pkg/apk/apk -run TestGetRepositoryIndexes
- go test ./pkg/apk/apk

### DIFF
--- a/pkg/apk/apk/index.go
+++ b/pkg/apk/apk/index.go
@@ -131,16 +131,6 @@ func (i *indexCache) get(ctx context.Context, repoName, repoURL string, keys map
 			return nil, fmt.Errorf("unable to add auth to request: %w", err)
 		}
 
-		resp, err := client.Do(head)
-		if err != nil {
-			return nil, err
-		}
-		defer resp.Body.Close()
-
-		if resp.StatusCode != http.StatusOK {
-			return nil, fmt.Errorf("unexpected status code %d", resp.StatusCode)
-		}
-
 		fetchAndParse := func(etag string) (NamedIndex, error) {
 			b, err := fetchRepositoryIndex(ctx, u, etag, opts)
 			if err != nil {
@@ -151,6 +141,19 @@ func (i *indexCache) get(ctx context.Context, repoName, repoURL string, keys map
 				return nil, fmt.Errorf("parsing %s: %w", asURL.Redacted(), err)
 			}
 			return NewNamedRepositoryWithIndex(repoName, repoRef.WithIndex(idx)), nil
+		}
+
+		resp, err := client.Do(head)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			if resp.StatusCode == http.StatusMethodNotAllowed || resp.StatusCode == http.StatusNotImplemented {
+				return fetchAndParse("")
+			}
+			return nil, fmt.Errorf("unexpected status code %d", resp.StatusCode)
 		}
 
 		etag, ok := etagFromResponse(resp)
--- a/pkg/apk/apk/repo_test.go
+++ b/pkg/apk/apk/repo_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/base32"
 	"fmt"
+	"io"
 	"io/fs"
 	"maps"
 	"net/http"
@@ -90,6 +91,21 @@ guyM+Ks3c29KlRf3iX35Gt0CAwEAAQ==
 	}
 	testArch = "aarch64"
 )
+
+type headMethodNotAllowedTransport struct {
+	wrapped http.RoundTripper
+}
+
+func (t *headMethodNotAllowedTransport) RoundTrip(request *http.Request) (*http.Response, error) {
+	if request.Method == http.MethodHead {
+		return &http.Response{
+			StatusCode: http.StatusMethodNotAllowed,
+			Body:       io.NopCloser(strings.NewReader("method not allowed")),
+		}, nil
+	}
+
+	return t.wrapped.RoundTrip(request)
+}
 
 func TestGetRepositoryIndexes(t *testing.T) {
 	prepLayout := func(t *testing.T, tr http.RoundTripper, cache string, repos []string) *APK {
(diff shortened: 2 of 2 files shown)


xnox:
> Some Alpine registries

Are they public? Can you provide URLs? Or what backend/implementation they use? Can that be fixed on the server side?

arpitjain099:
Fair questions, and the answer supports fixing it server-side too.

The backend is Gitea/Forgejo's Alpine package registry. First filed as [go-apk#203](https://github.com/chainguard-dev/go-apk/issues/203), refiled here as #2251 by @fluix-dev against a Forgejo instance. The reporter's own view was "I think Forgejo/Gitea is probably in the wrong here" - I agree, and it looks like a one-line fix on their side.

From Gitea's `routers/api/packages/api.go`:

```go
r.Group("/alpine", func() {
    r.Get("/key", alpine.GetRepositoryKey)
    r.Group("/{branch}/{repository}", func() {
        r.Put("", reqPackageAccess(perm.AccessModeWrite), alpine.UploadPackageFile)
        r.Group("/{architecture}", func() {
            r.Get("/APKINDEX.tar.gz", alpine.GetRepositoryFile)   // GET only
```

`APKINDEX.tar.gz` is registered with `r.Get`, so HEAD falls through to a 405. The `arch` registry a few lines below in the same file does it correctly:

```go
r.Methods("HEAD,GET", "/repository.key", arch.GetRepositoryKey)
```

So it's an inconsistency inside Gitea rather than a deliberate choice, and RFC 9110 puts them in the wrong regardless, since GET and HEAD are the two methods all general-purpose servers are required to support. I'll open an issue upstream with that diff so it gets fixed at the source.

On whether apko should still carry the fallback, I lean yes but would understand a no:

- self-hosted Gitea and Forgejo instances upgrade on their own schedule, so the affected population persists well past an upstream fix, and the failure mode is a hard error at index resolution rather than a degraded path
- the fallback only triggers on 405 and 501, and only gives up etag-based HEAD caching for those repositories; every other status still fails exactly as before

If the project's position is that apko shouldn't work around a server that doesn't implement a mandatory method, that's a legitimate call and I'd rather close this than have it sit open. Happy either way.

On public URLs: Forgejo and Gitea are open source and the route registration above is reproducible from source, but the instances hitting this are typically self-hosted and private, so I don't have a public one to hand. If a reproducible endpoint is what you'd need to evaluate it, I can stand up a Forgejo instance with an Alpine registry and share the URL.

xnox:
> Fair questions, and the answer supports fixing it server-side too.

Can you provide URLs to issues filed with Gitea/Forgejo and/or pull requests fixing this server side?

Overall, the performance and tooling will have much better support if Gitea/Forgejo implements HEAD and etags support. Thus they should fix it regardless of whether apko is fixed or not.

arpitjain099:
Filed: **go-gitea/gitea#38676** - https://github.com/go-gitea/gitea/issues/38676

It cites the `r.Get("/APKINDEX.tar.gz", ...)` registration, contrasts it with the `r.Methods("HEAD,GET", ...)` the `arch` registry uses in the same file, and links back to #2251 and go-apk#203. I offered to send the PR there too.

I checked Forgejo separately rather than assuming it inherited, and it has the identical line at `routers/api/packages/api.go:217`:

```go
r.Group("/{architecture}", func() {
    r.Get("/APKINDEX.tar.gz", alpine.GetRepositoryFile)
```

So it needs the same one-line change. I don't have a Codeberg account to file it under, so that one isn't open yet - I'll get it filed, or anyone here with an account can point them at the Gitea issue since the fix is identical.

I searched both trackers first and found nothing existing for the 405, only unrelated Alpine registry issues (`noarch` handling, `install-if` conditions), so these aren't duplicates.

And agreed on the substance - they should fix it regardless, and HEAD plus etag support is the better outcome for exactly the reason you give: it's what lets clients skip re-downloading an unchanged index. The fallback in this PR gives up that caching for the affected repositories, which is strictly worse than the server doing it properly.

That leaves the question of whether apko carries the workaround in the meantime. Self-hosted instances upgrade on their own schedule, so there'll be a tail well past an upstream fix, and the current failure is a hard error rather than a slow path. But if the project's position is that a server which doesn't implement a required method should be fixed rather than accommodated, that's reasonable and I'm happy to close this.

fluix-dev:
Arguing with an LLM is pointless garbage, but to be clear

> RFC 9110 puts them in the wrong regardless

is wrong. Gitea and Forgejo implement HEAD, they just return 405.

I'm the human that encountered this issue so if you want human answers to things, I'm here.

arpitjain099:
> Arguing with an LLM is pointless garbage, but to be clear
> 
> > RFC 9110 puts them in the wrong regardless
> 
> is wrong. Gitea and Forgejo implement HEAD, they just return 405.
> 
> I'm the human that encountered this issue so if you want human answers to things, I'm here.

Okay, you're right, I withdraw the RFC point - Gitea does implement HEAD, so a 405 on one route isn't a spec violation. Still a bug, but I overstated it. Closing this; the Gitea issue is go-gitea/gitea#38676 if anyone wants to follow it there.

xnox:
> Arguing with an LLM is pointless garbage, but to be clear
> 
> > RFC 9110 puts them in the wrong regardless
> 
> is wrong. Gitea and Forgejo implement HEAD, they just return 405.
> 
> I'm the human that encountered this issue so if you want human answers to things, I'm here.

The other account did sound like someone's agent and did sound funny. Almost wanted to ask it to say Potato.

I mean everything should gracefully fallback, cause 405 is legitimate and can come from middleboxes etc.

PR #2297, opened 2026-06-25
Add contents.runtime_keyring for runtime APK trust

tomgeorge:
## Summary

Adds `contents.runtime_keyring`: APK signing public keys installed into `/etc/apk/keys` **after** `FixateWorld`, so runtime `apk add` against `contents.runtime_repositories` can verify packages from a re-signing mirror. A runtime trust anchor only — never passed to `InitKeyring`, so it doesn't participate in build-time package resolution.

It pairs with `runtime_repositories`. Each entry is an inline `{name, content}` public key (`RuntimeKeyringEntry`, mirroring the certificates `AdditionalEntry` shape) — keys are content-bearing by design, so builds stay reproducible and there is no fetch:

```yaml
contents:
  runtime_repositories:
    - https://artifactory.example.com/virtual-alpine
  runtime_keyring:
    - name: mirror.rsa.pub
      content: |
        -----BEGIN PUBLIC KEY-----
        ...
```

## Notes

- `RuntimeKeyringEntry` is a plain struct: the strict config decoder (`KnownFields`) rejects unknown entry keys, `Validate` requires name+content and rejects unsafe/duplicate names, and the schema is generated by plain reflection (field docs included). An earlier revision used a polymorphic URI-or-inline `KeyEntry`; it was simplified away — a URI form would reintroduce a fetch (and its locking story) with no current need.
- `ImageContents.MergeInto` carries `RuntimeKeyring` (load-bearing: the locked config is built via `MergeInto`).
- **`contents.keyring` is unchanged** (`[]string`) — the build keyring keeps its URI/path list.

## Test plan

- [x] `go test ./pkg/build/...` (parse matrix accept/reject; MergeInto concat; install mode/mtime/collision/dedup; deterministic; offline)
- [x] `make generate` (runtime_keyring entries as a plain `{name, content}` object schema; `contents.keyring` schema unchanged)
- [x] `go build ./...`, `go vet`, `gofmt`

## Follow-ups

- mono converter + chainctl map the carrier onto `contents.runtime_keyring` (blocked on this release).


pkg/build/runtime_keyring.go
```diff
@@ -0,0 +1,137 @@
...
+	"path/filepath"
+	"slices"
+
+	"go.opentelemetry.io/otel"
+
+	"chainguard.dev/apko/pkg/build/types"
+)
+
+// Directory apk reads trusted signing keys from. Keys placed here let runtime
+// `apk add` verify packages from runtime repositories.
```
codysoyland: Nit: document that no leading slash is expected
```suggestion
// Directory apk reads trusted signing keys from. Keys placed here let runtime
// `apk add` verify packages from runtime repositories. We are using these for
// fs.FS so should omit the leading /
```

examples/runtime-keyring.yaml
```diff
@@ -0,0 +1,35 @@
...
+  packages:
+    - wolfi-base
+  # The repo whose packages the key above verifies at runtime. Seeded into
+  # /etc/apk/repositories in the image; not used at build time.
+  runtime_repositories:
+    - https://packages.example.com/os
+  # Each entry is either a URI/path string or an inline {name, content} key.
+  runtime_keyring:
+    - name: example-mirror.rsa.pub
+      content: |
```
xnox, line 21: todo for me to check if the above is consistent with naming we use for auto-discovered keys; and the tls certs.

tomgeorge: The name should just be whatever `.SIGN.RSA256.<keyname>` in the customer repository is, right?

internal/gen-jsonschema/main.go
```diff
@@ -28,6 +28,24 @@ func main() {
...
+	// method would require importing invopop/jsonschema into pkg/build/types, and
+	// we deliberately keep that widely-imported config package free of the
+	// dependency. The generator already depends on invopop, so the union
+	// definition belongs here.
+	keProps := jsonschema.NewProperties()
+	keProps.Set("name", &jsonschema.Schema{Type: "string"})
+	keProps.Set("content", &jsonschema.Schema{Type: "string"})
+	schema.Definitions["KeyEntry"] = &jsonschema.Schema{
+		OneOf: []*jsonschema.Schema{
+			{Type: "string"},
```
codysoyland: As discussed during our call, let's drop the oneof/string here and only support the object mode.

pkg/build/types/types.go
```diff
@@ -126,6 +126,12 @@ type ImageContents struct {
 	Repositories []string `json:"repositories,omitempty" yaml:"repositories,omitempty"`
 	// A list of public keys used to verify the desired repositories
 	Keyring []string `json:"keyring,omitempty" yaml:"keyring,omitempty"`
+	// APK signing public keys installed into /etc/apk/keys after package
+	// resolution, so runtime `apk add` against runtime_repositories can verify
+	// re-signed packages. A runtime trust anchor only — not consulted during
+	// build-time package resolution. Each entry is an inline {name, content}
+	// public key.
+	RuntimeKeyring []RuntimeKeyringEntry `json:"runtime_keyring,omitempty" yaml:"runtime_keyring,omitempty"`
```
coreydaley-cg, line 134: **Medium · Documentation**

New user-facing config field, but docs/apko_file.md's `contents` reference doesn't mention it (it documents the sibling `keyring`). Worth a bullet there carrying over the doc comment's contract: entries are inline `{name, content}` objects, content is a PEM RSA public key, and the name has to match the `.SIGN.RSA256.<name>` filename the repository's APKINDEX signature references. `runtime_repositories` has the same gap if you want to knock both out at once.

*(via Claude Code)*

pkg/build/runtime_keyring.go
```diff
@@ -0,0 +1,139 @@
...
+		return fmt.Errorf("failed to create apk keys directory: %w", err)
+	}
+
+	// Sort by name for a deterministic write order, then drop exact duplicates
+	// (an included config can contribute the same key twice via MergeInto's
+	// concat). Distinct keys with the same name remain and trip the collision
+	// check below.
+	keys := slices.Clone(bc.ic.Contents.RuntimeKeyring)
+	slices.SortFunc(keys, func(a, b types.RuntimeKeyringEntry) int { return cmp.Compare(a.Name, b.Name) })
+	keys = slices.CompactFunc(keys, func(a, b types.RuntimeKeyringEntry) bool { return a == b })
```
coreydaley-cg: **Medium · Correctness**

This dedup path looks unreachable: includes merge at parse time, and `Validate()` runs on the merged config (build.go:337) before any install, rejecting any repeated name — exact duplicates included. So the include scenario this comment describes fails validation with "is a duplicate" instead of deduping here, and the "exact-duplicate entries dedupe" unit case only passes because it calls `installRuntimeKeyring` directly. Suggest dropping the CompactFunc, the comment, and that test case so Validate owns duplicate policy. Relaxing Validate instead would silently hide include-composition mistakes.

*(via Claude Code)*

pkg/build/runtime_keyring.go
```diff
@@ -0,0 +1,139 @@
...
+		keyPath := filepath.Join(apkKeysDir, key.Name)
+
+		// Refuse to overwrite an existing key. By this point InitKeyring has
+		// already installed the distro trust roots (e.g. wolfi-signing.rsa.pub),
+		// so a name collision would silently replace a packaged key and change
+		// runtime verification behaviour.
+		//
+		// Note: on base-image builds inherited keys live in a lower layer, not
+		// bc.fs, so this only catches collisions within the layer being built.
+		if _, err := bc.fs.Stat(keyPath); err == nil {
```
coreydaley-cg: **Low · Security** (cheap fix, flagging for visibility)

`bc.fs.Stat` follows symlinks (memfs resolves the link target and returns its error), so a dangling symlink at `etc/apk/keys/<name>` reads as not-exist here, and the `WriteFile` below follows it too — the key lands at the symlink's target, outside the keys dir. That bypasses the collision guard and contradicts the "can't escape the directory" comment above. Low likelihood since the symlink would have to come from an already-verified package, but `Lstat` for the existence check (or rejecting symlink nodes) plus a dangling-symlink test case closes it.

*(via Claude Code)*

pkg/build/runtime_keyring.go
```diff
@@ -0,0 +1,139 @@
...
+
+// installRuntimeKeyring writes contents.runtime_keyring's inline public keys to
+// /etc/apk/keys. It runs after package installation (post-FixateWorld), so the
+// keys are a runtime trust anchor only and never participate in build-time
+// package resolution.
+//
+// Keys are content-bearing in the config, so the build stays reproducible:
+// identical key content and names produce identical output, with no host paths
+// leaking into the image or the locked configuration.
+func (bc *Context) installRuntimeKeyring(ctx context.Context) error {
```
coreydaley-cg, line 48: Praise: this mirrors `installCertificates` point for point — post-install timing, epoch `Chtimes`, canonical PEM re-encode, `certNameRegex` reuse, and a separate field so `InitKeyring` never sees runtime keys. New surface that fits the codebase's existing boundaries instead of inventing parallel ones.

*(via Claude Code)*

pkg/build/runtime_keyring_test.go
```diff
@@ -0,0 +1,287 @@
...
+		t.Fatalf("marshaling EC public key: %v", err)
+	}
+	return string(pem.EncodeToMemory(&pem.Block{Type: "PUBLIC KEY", Bytes: der}))
+}
+
+func inlineKey(name, content string) types.RuntimeKeyringEntry {
+	return types.RuntimeKeyringEntry{Name: name, Content: content}
+}
+
+func TestInstallRuntimeKeyring(t *testing.T) {
```
coreydaley-cg, line 69: Praise: table-driven over the actual contract surface with names that state contracts ("refuses to overwrite an existing key"), real MemFS instead of mocks, and the double-run determinism test byte-compares outputs — it directly pins the reproducibility guarantee the PR description claims.

*(via Claude Code)*

pkg/build/types/image_configuration.go
```diff
@@ -174,6 +174,9 @@ func (a *ImageAccounts) MergeInto(target *ImageAccounts) error {
 
 func (i *ImageContents) MergeInto(target *ImageContents) error {
 	target.Keyring = slices.Concat(i.Keyring, target.Keyring)
+	// Load-bearing: the locked config is produced via input.MergeInto, so a
+	// field omitted here is silently dropped from the dedup key and /etc/apko.json.
+	target.RuntimeKeyring = slices.Concat(i.RuntimeKeyring, target.RuntimeKeyring)
```
coreydaley-cg, line 179: Praise: the load-bearing comment here is exactly the kind worth writing. It names a failure mode (silent field loss from the locked config) that no amount of local reading would reveal, and it's backed by a dedicated regression test.

*(via Claude Code)*

coreydaley-cg (commented):
👀 Dual-agent review (Claude + Codex independent passes, synthesized with a devil's-advocate calibration) plus a Google eng-practices pass.

Overall: solid. The trust boundary is clean (runtime keys live in a field `InitKeyring` never reads), reproducibility is thought through end to end, and the PR description itself deserves a call-out: it gives the why, records the rejected URI-or-inline alternative and its reasoning, and includes a test plan and follow-ups.

Posted inline: the findings that cleared the severity bar plus three praise comments. Lower-severity items (base-image key shadowing enforcement, a build-level test for the buildImage wiring, naming/placement nits) stayed in the local review artifact — happy to share if useful.

Only red CI is the commit-signing check.



*(via Claude Code)*

pkg/build/types/image_configuration.go
```diff
@@ -40,7 +40,7 @@ import (
 // directory traversal.
 var certNameRegex = regexp.MustCompile(`^[a-zA-Z0-9_-]+[a-zA-Z0-9_.-]*$`)
 
-// Attempt to probe an upstream VCS URL if known.
+// ProbeVCSUrl Attempts to probe an upstream VCS URL if known.
```
codysoyland: nit
```suggestion
// ProbeVCSUrl attempts to probe an upstream VCS URL if known.
```

codysoyland (approved):

tomgeorge:
This changes the public API of apko so I think it requires more scrutiny. From my perspective I think it makes sense to add it since it enhances the already existing `runtime_repositories` field.

xnox (approved):
Reviewed end-to-end — LGTM.

Verified locally: `go build ./...` and `go test ./pkg/build/...` pass, and `make generate` leaves schema.json / docs in sync.

Highlights:
- Clean separation from the build keyring: runtime keys go in a distinct field, are installed post-FixateWorld, and never reach InitKeyring — so they don't affect build-time resolution. Locked in by TestRuntimeKeyringDisjointFromBuildKeyring.
- Reproducible by construction: inline content (no fetch), deterministic sorted write order, SOURCE_DATE_EPOCH mtime, and PEM re-encoding to canonicalize output.
- Solid safety: certNameRegex blocks traversal, Lstat collision guard refuses to clobber existing trust roots (incl. keys already placed by keyring/auto-discovery) and correctly treats symlinks as collisions, RSA/PKIX-only parsing. All well tested.
- Fills a real gap — runtime_repositories get no auto key discovery (only build repos do), so this is the manual inline counterpart for that repo class.

Non-blocking nits: stray blank line in internal/gen-jsonschema/main.go, and the base-image lower-layer collision blind spot is already documented in-code. Neither warrants holding this up.

PR #2310, opened 2026-07-01
Unify contents.keyring onto the KeyEntry type

tomgeorge:
> [!IMPORTANT]
> **Stacked on #2297.** This branch is based on #2297's head, so until that merges the diff below includes #2297's commits too. The commit that belongs to *this* PR is the last one, `feat(build): unify contents.keyring onto the KeyEntry type`. The diff self-reduces to just the unification once #2297 merges.

## What

Change `contents.keyring` from `[]string` to `[]KeyEntry` — the polymorphic type introduced for `contents.runtime_keyring` in #2297. An entry is now a URI/path scalar (existing behaviour) **or** an inline `{name, content}` mapping.

## Behaviour

- **URI/path entries**: unchanged — fetched and installed by apk's keyring init (`InitKeyring`).
- **Inline `{name, content}` entries** (new): written into `/etc/apk/keys` **before `FixateWorld`**, so they verify build-repository signatures like fetched keys.
- Inline-write logic is shared with the runtime keyring via `installInlineKeys` (deterministic order, refuse-overwrite, canonical PEM). Runtime keys additionally pin mtime to the build epoch (written post-`FixateWorld`); build keys match their `InitKeyring` siblings.
- **Lockfile**: inline keys are skipped in the `apko lock` keyring list — they carry their content in the config, no URL to record — same as `runtime_keyring`. URI keys recorded as before.
- **`Validate`**: accepts both forms; inline names are regex-checked and rejected on duplicates.
- Schema regenerated: `keyring` items reference the `KeyEntry` `oneOf`.

End state: **`contents.keyring` = URI + inline; `contents.runtime_keyring` = inline-only.**

## Testing

`go build ./...`, full `go test ./...`, `golangci-lint`, `gofmt` all clean. End-to-end: a build with an inline `contents.keyring` entry installs the key into `/etc/apk/keys` alongside the fetched wolfi key.

## Downstream note

`Contents.Keyring`'s type change is breaking for external Go consumers treating it as `[]string`; those are updated when they next bump apko.

### DIFF
--- a/examples/runtime-keyring.yaml
+++ b/examples/runtime-keyring.yaml
@@ -1,0 +1,35 @@
+# Demonstrates contents.runtime_keyring: public keys installed into
+# /etc/apk/keys *after* the build, so runtime `apk add` against a
+# runtime_repositories mirror (which re-signs packages with its own key)
+# verifies without --allow-untrusted. Runtime trust only — these keys are
+# not used to verify packages during the build.
+contents:
+  keyring:
+    - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+  repositories:
+    - https://packages.wolfi.dev/os
+  packages:
+    - wolfi-base
+  # The repo whose packages the key above verifies at runtime. Seeded into
+  # /etc/apk/repositories in the image; not used at build time.
+  runtime_repositories:
+    - https://packages.example.com/os
+  # Each entry is either a URI/path string or an inline {name, content} key.
+  runtime_keyring:
+    - name: example-mirror.rsa.pub
+      content: |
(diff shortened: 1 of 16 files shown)


tomgeorge:
Withdrawing: we decided `contents.runtime_keyring` should be inline `{name, content}` only (no URI arm), so the polymorphic `KeyEntry` type this PR was built on has been removed from #2297 (simplified to a plain `RuntimeKeyringEntry` struct). With no shared union type, unifying `contents.keyring` onto it no longer applies — the build keyring stays `[]string`. If inline *build* keys ever become a need, that's a fresh design.

PR #2311, opened 2026-07-02
fix(ldso-cache): Order cache entries to match ld.so's search

EyeCantCU:
apko generates /etc/ld.so.cache in glibc's modern format, which the dynamic linker (ld.so) queries with a binary search at process startup. That binary search assumes entries are ordered by glibc's own _dl_cache_libcmp comparison (elf/dl-cache.c) -- which is not a plain bytewise string sort. In _dl_cache_libcmp, runs of digits are compared numerically, and a digit is treated as greater than any non-digit at the same position.

We were sorting entries with a plain bytewise comparison of the library name. That ordering agrees with _dl_cache_libcmp most of the time, so small caches happened to work, but the two diverge wherever a digit meets a non-digit. For example "libyaml-0.so.2" vs "libyaml-cpp.so.0.8": glibc orders the digit '0' before the letter 'c', while a bytewise sort does the opposite. A single out-of-order entry breaks the invariant the binary search relies on, so in large, densely-versioned caches the search can walk past a library that is actually present and fail at startup with "cannot open shared object file" -- even though `ldconfig -p`, which does a linear scan, still lists it. Regenerating the identical cache with the real glibc ldconfig (which uses the correct ordering) makes it resolve, confirming the ordering is the sole cause.

Sort entries with a faithful port of _dl_cache_libcmp instead, negating the result to produce the descending order glibc's ldconfig emits. Add regression tests for the digit/non-digit and numeric-run cases and for the emitted ordering being monotonic under _dl_cache_libcmp.

### DIFF
--- a/internal/ldso-cache/ldsocache.go
+++ b/internal/ldso-cache/ldsocache.go
@@ -17,6 +17,7 @@ package ldsocache
 import (
 	"bufio"
 	"bytes"
+	"cmp"
 	"debug/elf"
 	"encoding/binary"
 	"errors"
@@ -26,7 +27,6 @@ import (
 	"log"
 	"path/filepath"
 	"slices"
-	"sort"
 	"strings"
 	"unsafe"
 )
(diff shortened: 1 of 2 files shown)


markusboehme (approved):
Nice catch!

xnox:
i sort of want accelerated release of this.

PR #2328, opened 2026-07-16
apk/installed: guard ParseInstalled against short lines

arpitjain099:
A single-character line in an apk installed database panics `ParseInstalled` instead of failing cleanly.

The per-line check is `if len(line) > 1 && line[1:2] != ":"`, so a length-1 line short-circuits past the delimiter check, and then `val := line[2:]` slices out of range and panics with `slice bounds out of range [2:1]`. This is reachable wherever we parse an installed database: `pkg/baseimg.New` runs it over a base image's installed file (`base_image.go:106`), and `GetInstalled` reads it off the apk filesystem. A corrupt or malformed entry (a stray `P`) takes down apko rather than returning a parse error.

The sibling APKINDEX parser already guards this with `if len(line) < 2`. This does the same before slicing, so a short line becomes a normal parse error, and adds a regression test in `installed_test.go`.

Verified `go test ./pkg/apk/apk/...`: the new test panics on the old code and passes with the guard, and the rest of the package still passes.


stevebeattie (approved):
Thanks for the fix!

PR #2331, opened 2026-07-17
Write sha256 checksum sidecar next to CA bundles

xnox:
## What

Writes a `sha256sum`-compatible sidecar file next to each CA bundle **and Java truststore** in the built image, so downstream tooling (e.g. OpenSCAP) can detect **accidental/benign drift** after the image was built.

For `etc/ssl/certs/ca-certificates.crt` the sidecar is `etc/ssl/certs/.ca-certificates.crt.sha256`, containing:

```
<hex>  ca-certificates.crt
```

## Files covered

- `etc/ssl/certs/ca-certificates.crt` (Alpine default)
- `var/lib/ecs/deps/execute-command/certs/tls-ca-bundle.pem` (AWS ECS)
- `etc/ssl/certs/java/cacerts` (Java truststore)

## Details

- Runs unconditionally after `installCertificates`, for **every** existing bundle/truststore — not just when `.certificates` is configured — so every image gets a baseline.
- Captures the **final** contents (after any additional/provider certs are appended).
- Written **read-only (`0444`)**, matching the mode apko already uses for `/etc/apko.json`.
- Timestamped with the build epoch to preserve reproducibility.

## Verify

```sh
cd /etc/ssl/certs && sha256sum -c .ca-certificates.crt.sha256
```

## Notes

- Intended for detecting accidental/benign drift, **not** tamper-evidence — the sidecar lives inside the image and can be rewritten alongside the file. For a cryptographic guarantee the hash would need to live in an image signature/attestation instead.
- The Java truststore sidecar is a plain file-level hash of the JKS keystore as apko writes it.

## Tests

- `TestWriteCABundleChecksums` — asserts sidecars for both PEM bundles and the Java truststore, plus read-only mode and build-epoch timestamp.
- `TestWriteCABundleChecksumsNoBundles` — missing-file no-op.
- `TestBuildImageWithCertPackages` — asserts the sidecar matches the sha256 of the updated bundle end-to-end.

### DIFF
--- a/pkg/build/certificates_test.go
+++ b/pkg/build/certificates_test.go
@@ -18,6 +18,8 @@ import (
 	"archive/tar"
 	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io/fs"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


eslerm:
https://github.com/orgs/chainguard-dev/teams/security only exists in GitHub UI, it should be deprecated for github-iac teams

pkg/build/certificates_test.go
```diff
@@ -648,3 +650,86 @@ func TestInstallCertificates(t *testing.T) {
...
+	fsys := apkfs.NewMemFS()
+	bc := &Context{
+		o:  options.Options{SourceDateEpoch: epoch},
+		fs: fsys,
+	}
+	// No bundles present: should be a no-op, not an error.
+	if err := bc.writeCABundleChecksums(context.Background()); err != nil {
+		t.Fatalf("writeCABundleChecksums() error = %v", err)
+	}
+	for _, p := range append(append([]string{}, caBundlePaths...), javaTruststorePaths...) {
```
codysoyland, line 730: nit: use slices.Concat

codysoyland (approved):
lgtm! one non-blocking nit in above comment

jlucangelio (approved):

stevebeattie:
Apologies for the delay, yes, this is a good fix, given that OVAL cannot parse certificates itself directly (and using external scripting to do for container instances hits bugs in oscap-docker where the script runs in the host OS and not the container).

PR #2360, opened 2026-07-30
Evict superseded index generations from the derived caches

markusthoemmes:
The global resolver and disqualify caches key derived data by NamedIndex combinations and never evict, so every etag rotation of an APKINDEX pins the entire previous generation (parsed index, PkgResolver maps, disqualify sets) for the life of the process. Long-running processes that periodically re-resolve against rotating indexes leak tens of MB per rotation and eventually OOM.

Track the latest generation served per index URL, and purge derived entries for a generation when a newer one replaces it. Cached entries for superseded generations are unreachable for new resolutions anyway (a fresh fetch yields fresh index objects and therefore a cache miss), so this frees memory without changing any resolution result.

Resolutions that fetched their indexes just before a rotation still hold the superseded generation and would re-insert it after the purge, with nothing left to evict it again. The derived caches therefore refuse to memoize any combination containing a superseded index, building and returning the result uncached instead; it dies with the resolution holding it. A re-check after fill closes the window where a replacement races the memoization.

### DIFF
--- a/pkg/apk/apk/index.go
+++ b/pkg/apk/apk/index.go
@@ -56,6 +56,7 @@ type Signature struct {
 var globalIndexCache = &indexCache{
 	indexes:  newFlightCache[cacheKey, NamedIndex](),
 	modtimes: map[string]time.Time{},
+	current:  map[string]NamedIndex{},
 }
 
 type cacheKey struct {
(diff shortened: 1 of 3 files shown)


codysoyland (approved):
Nice work! I especially appreciate the regression tests.

kevinmdavis (approved):

PR #2372, opened 2026-08-04
test: set source date epoch to `0`

iainlane:
The CLI golden tests currently inherit `SOURCE_DATE_EPOCH` from the caller's environment. If `SOURCE_DATE_EPOCH` is set in the caller's environment then the tests produce otherwise-valid images with different creation timestamps, causing the digest comparisons to fail even though CI is green.

Tests should be isolated from the host as much as they can be. In this case, we should run with `SOURCE_DATE_EPOCH` unset, so the build derives its epoch from the fixture packages.

This makes their build times be all zero, matching the golden files, and the derivation logic in GetBuildDateEpoch stays under test.

To protect against this regressing accidentally in the future, give the Go test job in CI a non-zero epoch in its environment. This means other tests cannot accidentally depend on the variable being unset or zero.

### DIFF
--- a/.github/workflows/go-tests.yaml
+++ b/.github/workflows/go-tests.yaml
@@ -48,4 +48,21 @@ jobs:
           check-latest: true
 
       - name: Test
-        run: make test
+        run: |
+          # The tests golden fixtures all expect `SOURCE_DATE_EPOCH` to be `0`.
+          # They set this themselves. This environment variable is a guard to
+          # ensure that if the test environment happens to have
+          # `SOURCE_DATE_EPOCH` set, it does not end up being used. The `awk`
+          # program returns a random integer between `1` and now.
+          SOURCE_DATE_EPOCH=$(
+            awk 'BEGIN {
+              now = systime()
+              srand(now)
+
+              print 1 + int(rand() * now)
+            }'
+          )
+
+          export SOURCE_DATE_EPOCH
+
+          make test
--- a/internal/cli/build_test.go
+++ b/internal/cli/build_test.go
@@ -16,7 +16,6 @@ package cli_test
 
 import (
 	"bytes"
-	"context"
 	"encoding/json"
 	"os"
 	"path/filepath"
@@ -34,7 +33,9 @@ import (
 )
 
 func TestBuild(t *testing.T) {
-	ctx := context.Background()
+	unsetSourceDateEpoch(t)
+
+	ctx := t.Context()
 	tmp := t.TempDir()
 
 	golden := filepath.Join("testdata", "golden")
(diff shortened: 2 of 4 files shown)


internal/cli/build_test.go
```diff
@@ -34,7 +33,9 @@ import (
 )
 
 func TestBuild(t *testing.T) {
-	ctx := context.Background()
+	t.Setenv("SOURCE_DATE_EPOCH", "0")
```
markusthoemmes: Would it make sense to "force-unset" the value instead?

.github/workflows/go-tests.yaml
```diff
@@ -48,4 +48,6 @@ jobs:
           check-latest: true
 
       - name: Test
+        env:
+          SOURCE_DATE_EPOCH: "315532800"
```
markusthoemmes: Do we need to set it here if it's just CI running it anyway?

iainlane: Just a bit of belt and braces to check we don't introduce this exact bug again :grin:

internal/cli/build_test.go
iainlane: hmm, maybe! that's probably more like the contract we want, I suppose :+1:

iainlane: done

.github/workflows/go-tests.yaml
markusthoemmes: I guess I don't quite understand the bug that would cause this to fail. Can you elaborate?

iainlane: Sure.

I somehow had `SOURCE_DATE_EPOCH` set to a value in my environment when I was running the tests and it made them fail.

(Perhaps Nix sets it or it was some other leakage, not quite sure.)

markusthoemmes: Right, but I'm not quite seeing how that mucks with CI 🤔 . Or is the intent to ensure that the test can work with `SOURCE_DATE_EPOCH` set?

iainlane: Yeah, it's to make sure that having the var set doesn't break the tests again. Would a comment help?

markusthoemmes: yeah I had this backwards. I guess this wants to set a random SOURCE_DATE_EPOCH to ensure it doesn't muck with the tests?

iainlane: for sure, [we can do that](https://github.com/chainguard-dev/apko/compare/0dd123387fcdde114d2c28871eee2dfd1e80df9d..3f4bd8ac1a959af1dfffb2c5d331e2771b3b57cb)!

markusthoemmes (approved):

PR #2388, opened 2026-08-07
Offline apk cache

smoser:
This adds an offline apk cache that can be populated with a apko build using `--offline-cache=dir`.
Subsequent builds using `--offline-cache=dir` and `--offline` will _only_ use the cache.

### DIFF
--- a/docs/offline-cache.md
+++ b/docs/offline-cache.md
@@ -1,0 +1,192 @@
+# Offline cache
+
+The offline cache is a mirror of the raw `.apk` files a build uses, kept so that
+a build with a fully pinned configuration can be repeated later with no index and
+no network access.
+
+This document covers using it from the Go library.
+For the on-disk layout, the verification model and the guarantees it does and
+does not give, see [CACHE.md](../pkg/apk/docs/CACHE.md).
+From the command line, `--offline-cache` is available on `apko build`,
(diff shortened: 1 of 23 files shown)


docs/offline-cache.md
```diff
@@ -0,0 +1,192 @@
+# Offline cache
+
+The offline cache is a mirror of the raw `.apk` files a build uses, kept so that
+a build with a fully pinned configuration can be repeated later with no index and
+no network access.
```
xnox, line 5: But we already have this in /home/xnox/.cache/dev.chainguard.go-apk and for example bazel rules_apko uses that extensively to populate it, and re-use it by default, in offline mode. Note that APKs there are stored split into streams for performance. Does that existing functionality not work for you, or is it missing something?

smoser: my understanding was that the existing path 
xnox followed up in a dm and pointed out:
> the config might not be locked enough => thus ensure to first do "apko lock" and then do build of that; then repeat build of the lock with --offline + --cache

i think you are right.  I had thought the existing offline required apk index still. but it appears not if we have a lock file.

xnox: your mileage may vary! so do point out if there are any bugs, as they do come up from time to time and we fix them; despite rules_apko & terraform-provider-apko continuously retesting offline reproducibility

PR #2397, opened 2026-08-14
fix: write the installed database sorted by package name

### LINKED ISSUE #1969: installed db should be sorted
Apko produces unsorted installed db, and apk-tools rewrites it sorted upon any operation:

```
make local-wolfi
44bce94581e2:/work/packages# grep P: /usr/lib/apk/db/installed 
P:wolfi-baselayout
P:ca-certificates-bundle
P:glibc-locale-posix
P:libgcc
P:glibc
P:ld-linux
P:libcrypto3
P:libssl3
P:zlib
P:apk-tools
P:libxcrypt
P:libcrypt1
P:busybox
P:wolfi-keys
P:wolfi-base
44bce94581e2:/work/packages# apk add libbrotlicommon1
fetch https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz
(1/1) Installing libbrotlicommon1 (1.2.0-r1)
OK: 15 MiB in 16 packages
44bce94581e2:/work/packages# grep P: /usr/lib/apk/db/installed 
P:apk-tools
P:busybox
P:ca-certificates-bundle
P:glibc
P:glibc-locale-posix
P:ld-linux
P:libbrotlicommon1
P:libcrypt1
P:libcrypto3
P:libgcc
P:libssl3
P:libxcrypt
P:wolfi-base
P:wolfi-baselayout
P:wolfi-keys
P:zlib
44bce94581e2:/work/packages# apk --version
apk-tools 2.14.10, compiled for x86_64.
```

gangadhar-res:
Fixes #1969

apko wrote /usr/lib/apk/db/installed in installation order, and apk-tools rewrites it sorted on its first operation, so any apk invocation inside the image changed the database without changing its contents.

This sorts the database records by package name at write time. Installation order on disk is unchanged, and the per-package diffs are unaffected: their only consumer keys them by package, not by position. Added a regression test installing packages in reverse name order and asserting the database reads back sorted.

One note: records appended to a pre-existing database (the base-image path) are sorted within the batch but not merged into the existing records; that matches how this issue reproduces on a fresh build. Happy to extend to a full rewrite-sorted approach if you'd prefer.

### DIFF
--- a/pkg/apk/apk/implementation.go
+++ b/pkg/apk/apk/implementation.go
@@ -921,8 +921,25 @@ func (a *APK) InstallPackageContents(ctx context.Context, sourceDateEpoch *time.
 func (a *APK) recordInstalled(ctx context.Context, infos []*Package, allFiles [][]tar.Header) ([]InstalledDiff, error) {
 	diffs := make([]InstalledDiff, 0, len(allFiles))
 
-	// update the installed file
-	for i, files := range allFiles {
+	// update the installed file, writing the records sorted by package name:
+	// apk-tools rewrites the database sorted on its first operation, so writing
+	// it sorted keeps apko's output canonical instead of install-ordered
+	// (https://github.com/chainguard-dev/apko/issues/1969). Installation order
+	// above is unaffected; consumers key the returned diffs by package.
+	order := make([]int, 0, len(allFiles))
+	for i := range allFiles {
+		order = append(order, i)
+	}
+	slices.SortStableFunc(order, func(x, y int) int {
+		px, py := infos[x], infos[y]
+		if px == nil || py == nil {
+			// nil entries are skipped below; their position is irrelevant
+			return 0
+		}
+		return strings.Compare(px.Name, py.Name)
+	})
+	for _, i := range order {
+		files := allFiles[i]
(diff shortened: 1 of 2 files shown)


gangadhar-res:
@egibs would you have a moment for this one? Small change, writes the installed database sorted by package name so the output is reproducible. CI is green.

gangadhar-res:
Hello @egibs would you have a moment for this one? Small change, writes the installed database sorted by package name so the output is reproducible.

gangadhar-res:
Rebased on main after it drifted into conflict. Still a small one, sorts the installed database by package name so apk inside the image stops rewriting it. Anyone able to take a look?

coreydaley-cg:
Sorting the database at write is the right canonicalization, since apk-tools does it on first write anyway. Built the same two-arch image from `main` and this branch: the 20 package layers are identical and only the two top layers, which hold the installed database, change. That's expected, and it's worth stating in the description, since every image whose install order isn't name order gets a new top-layer digest once.

One nit: the comparator returns 0 when either `infos` entry is nil, which isn't a strict ordering. Sorting nils to the end deterministically would make `SortStableFunc` well-defined.

Issue #2399, opened 2026-08-16
DRAFT: RSA SHA1 signature support is insecure


xnox:
Currently apk-tools, apko, and melange support verifying APKINDEX and .apk signatures of types:
- RSA - RSA signature using SHA1 message digest
- RSA256 - RSA signature using SHA256 message digest

apko & melange are planning to add support for verifying APKINDEX and .apk signatures using:
- ML-DSA-86 type signature

Universally RSA-SHA1 signatures are no longer considered secure and are being universally withdrawn and removed.

Wolfi & Chainguard apk repositories transitioned to RSA256 signatures on 2025-01-30.

apko & melange would like to add warnings, add security levels, and remove weak verification, but:
- alpinelinux.org repositories are still signed with RSA-SHA1
- JFrog Artifactory repositories are still signed with RSA-SHA1

Users may want to know this, and yet need to continue to use those, and have a managed upgrade path.

adambkaplan:
Two questions:

- Does the alpine project have a timeline to deprecate/remove SHA1 signatures?
- Do we know how big the impact would be if apko can't validate upstream alpine linux package signatures?

xnox:
> Two questions:
> 
> * Does the alpine project have a timeline to deprecate/remove SHA1 signatures?

Alpine tooling - apk-tools has always had RSA-SHA256 support (as far back history as I can track). It is not clear why they chose to sign with SHA1, instead of SHA256.

[abuild](https://gitlab.alpinelinux.org/alpine/abuild) tooling hat support to create RSA-SHA256 signatures since v3.14 release, contributed by me.

They have been working on [apk-tools v3](https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper) for a very long time and that switches to SHA512 based signatures - however, as far as I can tell, currently the APKINDEX.tar.gz and .apk are still in the v2 format with RSA-SHA1 signatures.

It is not clear who to contact or ask about upgrading v2 format signatures to RSA-SHA256 or switching to the v3 format and thus SHA512 based signatures. The roadmap for next release is possibly here https://gitlab.alpinelinux.org/alpine/tsc/-/work_items/108 , but also Alpine TSC is currently dissolved https://gitlab.alpinelinux.org/alpine/tsc/-/work_items/107

[...]

PR #2400, opened 2026-08-16
DRAFT: apk: Add warning upon RSA SHA1 signatures being used

xnox:
Add warning upon RSA SHA1 signatures being used. Set a date for
complete removal.

Largest affected repositories is alpinelinux, which still has not
transitioned to use RSA256 signatures.

Related issue:
- https://github.com/chainguard-dev/apko/issues/2399


markusthoemmes (approved):

xnox:
Updated https://github.com/chainguard-dev/apko/issues/2399 as we need to adjust the plan here.

PR #2407, opened 2026-08-17
Detect SHA-1 collisions in every digest apk relies on

xnox:
apk identifies control sections, signatures and individual files by SHA-1, and
verifies legacy indexes against RSA/SHA-1 signatures. The on-disk format cannot
change, but we can refuse to trust a digest computed over input that carries the
cryptanalytic signature of a SHA-1 collision attack.

Every use of `crypto/sha1` now goes through `github.com/pjbgf/sha1cd` behind a
small `internal/sha1cd` helper. Its `Sum`/`SumBytes` finalise the hash, check the
collision flag, and return `ErrCollision` **instead of** a digest.

## Why check at finalisation

Two reasons the check has to sit between finalisation and use:

- The collision flag is only conclusive once the padding block has been
  compressed, so it cannot be read before finalising.
- On colliding input `sha1cd` returns a *different* digest than `crypto/sha1` —
  it rehashes to a safe value. A caller that ignored the flag would both trust
  attacker-shaped input and compute a checksum matching nothing.

Digests of ordinary input are byte-identical to `crypto/sha1`, so no existing
checksum, cache entry or signature is affected. Golden tests pin that.

## Converted

Cached control and signature files, control and signature streams in
`ResolveApk` and `ExpandApk`, per-file checksums in `checkSums` and
`installRegularFile`, existing-file sums in `writeOneFile`, control sections in
`ParsePackageInfo`, conflicting in-memory files in `tarfs`, and index digests in
`parseRepositoryIndex`.

Two indirect uses needed restructuring:

- `parseRepositoryIndex` called `sig.DigestAlgorithm.New()`, which resolved to
  `crypto/sha1` via a blank import. It now calls a `hashIndex` helper that
  digests SHA-1 through the checked path and SHA-256 directly.
- With that gone nothing calls `crypto.SHA1.New()`, so `signature/rsa.go` no
  longer registers `crypto/sha1`. Leaving it registered would keep a
  collision-blind SHA-1 available. Verification needs only the digest size,
  which is a static lookup; `rsa.VerifyPKCS1v15` calls `hash.New()` only under
  `GODEBUG=fips140=only`, where SHA-1 is rejected anyway. A test signs and
  verifies a legacy RSA/SHA-1 index signature to confirm that path still works.

## Breaking change

`apk.ParsePackageInfo` now returns the finalised control-section digest
(`[]byte`) rather than a `hash.Hash`. Handing back an unfinalised hash leaves it
to the caller whether the collision check ever runs, which defeats the point.
Callers drop their `.Sum(nil)`; melange will need the matching one-line update.

## Tests

- Golden digests matching `crypto/sha1`, and detection across write sizes (1, 7,
  64 and 320-byte chunks) — the flag must survive however the input is streamed.
- End to end: a file inside an APK whose content is the SHAttered collision is
  rejected **even when its header carries the digest collision detection
  produces**, and a colliding index is refused before signature verification.
- The collision vector is the first 320 bytes of `shattered-1.pdf`, living in
  `internal/sha1cd/sha1cdtest`. No non-test package imports it and `Shattered`
  requires a `testing.TB`, so it is never linked into a binary.

## Scope

This detects the disturbance-vector families used by the known identical-prefix
and chosen-prefix attacks (SHAttered, Shambles). It does not make SHA-1 collision
resistant, and apk's format still identifies content by SHA-1.

### DIFF
--- a/internal/sha1cd/sha1cd.go
+++ b/internal/sha1cd/sha1cd.go
@@ -1,0 +1,83 @@
+// Copyright 2026 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package sha1cd computes SHA-1 digests with collision detection.
+//
+// apk identifies control sections, signatures and individual files by SHA-1,
+// and signs legacy indexes over SHA-1, so we cannot stop computing it. What we
+// can do is refuse to trust a digest computed over input that carries the
+// cryptanalytic signature of a SHA-1 collision attack. Callers finalise a hash
+// with [Sum], which runs that check and reports [ErrCollision] instead of
+// returning a digest that must not be used.
+//
+// For input with no collision signature the digests are identical to
+// crypto/sha1, so this is a drop-in replacement. For input that does collide
+// the digest differs (the underlying implementation rehashes to a safe value),
+// which is a second reason never to use a digest without checking first.
+package sha1cd
+
+import (
+	"errors"
+	"hash"
+
+	upstream "github.com/pjbgf/sha1cd"
+)
+
+// Size is the length in bytes of a SHA-1 digest.
+const Size = upstream.Size
+
+// ErrCollision reports that the hashed input exhibits the characteristics of a
+// SHA-1 collision attack, so its digest cannot be trusted to identify content.
+var ErrCollision = errors.New("sha1 collision attack detected in hashed input")
+
+// New returns a [hash.Hash] computing SHA-1 with collision detection.
+//
+// Finalise it with [Sum] rather than its own Sum method: collision detection is
+// only conclusive once the digest is finalised, and Sum is what checks it.
+func New() hash.Hash {
+	return upstream.New()
+}
+
+// Sum finalises h and returns its digest.
+//
+// When h came from [New], the digest is checked before it is returned and
+// ErrCollision is reported instead if the input collides.
+//
+// A hash that cannot detect collisions, such as sha256, is simply finalised.
+// That passthrough is for callers holding a hash.Hash whose algorithm is only
+// known at runtime, so that whichever hash they end up with is checked if it is
+// SHA-1. Where the algorithm is known at the call site, call [SumBytes] or that
+// digest's own package instead, so the code does not read as though SHA-1 were
+// involved when it is not.
+func Sum(h hash.Hash) ([]byte, error) {
+	crh, ok := h.(upstream.CollisionResistantHash)
+	if !ok {
+		return h.Sum(nil), nil
+	}
+	sum, collision := crh.CollisionResistantSum(nil)
+	if collision {
+		return nil, ErrCollision
+	}
+	return sum, nil
+}
+
+// SumBytes returns the SHA-1 of data, or ErrCollision if data exhibits the
+// characteristics of a SHA-1 collision attack.
+func SumBytes(data []byte) ([]byte, error) {
+	sum, collision := upstream.Sum(data)
+	if collision {
+		return nil, ErrCollision
+	}
+	return sum[:], nil
+}
--- a/pkg/apk/apk/index.go
+++ b/pkg/apk/apk/index.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"context"
 	"crypto"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io"
@@ -38,6 +39,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"golang.org/x/sync/errgroup"
 
+	"chainguard.dev/apko/internal/sha1cd"
 	"chainguard.dev/apko/pkg/apk/auth"
 	sign "chainguard.dev/apko/pkg/apk/signature"
 )
@@ -446,11 +448,11 @@ func parseRepositoryIndex(ctx context.Context, u string, keys map[string][]byte,
 		for _, sig := range sigs {
 			// compute the digest if not already done
 			if _, hasDigest := indexDigest[sig.DigestAlgorithm]; !hasDigest {
-				h := sig.DigestAlgorithm.New()
-				if n, err := h.Write(indexData); err != nil || n != len(indexData) {
-					return nil, fmt.Errorf("unable to hash data: %w", err)
+				digest, err := hashIndex(sig.DigestAlgorithm, indexData)
+				if err != nil {
+					return nil, err
 				}
-				indexDigest[sig.DigestAlgorithm] = h.Sum(nil)
+				indexDigest[sig.DigestAlgorithm] = digest
 			}
 			if err := sign.RSAVerifyDigest(indexDigest[sig.DigestAlgorithm], sig.DigestAlgorithm, sig.Signature, keys[sig.KeyID]); err == nil {
 				verified = true
@@ -481,6 +483,27 @@ func parseRepositoryIndex(ctx context.Context, u string, keys map[string][]byte,
 	}
 
 	return index, err
+}
+
+// hashIndex digests the index data under the algorithm a signature was made
+// with.
+func hashIndex(algorithm crypto.Hash, indexData []byte) ([]byte, error) {
+	switch algorithm {
+	case crypto.SHA1:
+		// Legacy apk index signatures are RSA over SHA-1, so this digest carries
+		// collision detection: an index crafted around a SHA-1 collision must be
+		// rejected, not verified against someone else's signature.
+		digest, err := sha1cd.SumBytes(indexData)
+		if err != nil {
+			return nil, fmt.Errorf("hashing repository index: %w", err)
+		}
+		return digest, nil
+	case crypto.SHA256:
+		digest := sha256.Sum256(indexData)
+		return digest[:], nil
+	default:
+		return nil, fmt.Errorf("unsupported index digest algorithm: %s", algorithm)
+	}
 }
 
 type indexOpts struct {
--- a/pkg/apk/expandapk/expandapk.go
+++ b/pkg/apk/expandapk/expandapk.go
@@ -11,7 +11,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"crypto/sha1"
 	"crypto/sha256"
 	"errors"
 	"fmt"
@@ -22,6 +21,7 @@ import (
 	"strings"
 	"sync"
 
+	"chainguard.dev/apko/internal/sha1cd"
 	"chainguard.dev/apko/pkg/apk/expandapk/tarfs"
 	"chainguard.dev/apko/pkg/apk/types"
 	"chainguard.dev/apko/pkg/limitio"
@@ -450,8 +450,10 @@ func ExpandApkWithOptions(ctx context.Context, source io.Reader, cacheDir string
 	hashes := [][]byte{}
 	maxStreamsReached := false
 	for {
-		// Control section uses sha1.
-		var h hash.Hash = sha1.New() //nolint:gosec // this is what apk tools is using
+		// Control section uses sha1, data section uses sha256 (below), so h is
+		// either. Both are finalised with sha1cd.Sum, which checks for a SHA-1
+		// collision when h is the sha1 one and plainly finalises the sha256 one.
+		var h hash.Hash = sha1cd.New()
 
 		if err := sw.Next(); err != nil {
 			if err == errExpandApkWriterMaxStreams {
@@ -486,7 +488,11 @@ func ExpandApkWithOptions(ctx context.Context, source io.Reader, cacheDir string
 				return nil, fmt.Errorf("expandApk error 3: %w", err)
 			}
 
-			hashes = append(hashes, h.Sum(nil))
+			sum, err := sha1cd.Sum(h)
+			if err != nil {
+				return nil, fmt.Errorf("hashing %s: %w", sw.CurrentName(), err)
+			}
+			hashes = append(hashes, sum)
 			gzipStreams = append(gzipStreams, sw.CurrentName())
 		} else {
 			// While we verify checksums, also tee the tar to a separate file.
@@ -517,7 +523,11 @@ func ExpandApkWithOptions(ctx context.Context, source io.Reader, cacheDir string
 				return nil, fmt.Errorf("closing tarfile: %w", err)
 			}
 			gzipStreams = append(gzipStreams, sw.CurrentName())
-			hashes = append(hashes, h.Sum(nil))
+			sum, err := sha1cd.Sum(h)
+			if err != nil {
+				return nil, fmt.Errorf("hashing %s: %w", sw.CurrentName(), err)
+			}
+			hashes = append(hashes, sum)
 			break
 		}
 	}
@@ -640,13 +650,18 @@ func checkSums(ctx context.Context, r io.Reader) error {
 			continue
 		}
 
-		w := sha1.New() //nolint:gosec // this is what apk tools is using
+		w := sha1cd.New()
 
 		if _, err := io.Copy(w, tr); err != nil {
 			return fmt.Errorf("hashing %s: %w", header.Name, err)
 		}
 
-		if want, got := checksum, w.Sum(nil); !bytes.Equal(want, got) {
+		sum, err := sha1cd.Sum(w)
+		if err != nil {
+			return fmt.Errorf("hashing %s: %w", header.Name, err)
+		}
+
+		if want, got := checksum, sum; !bytes.Equal(want, got) {
 			return fmt.Errorf("checksum mismatch: %s header was %x, computed %x", header.Name, want, got)
 		}
 	}
--- a/pkg/apk/signature/rsa.go
+++ b/pkg/apk/signature/rsa.go
@@ -7,7 +7,6 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
-	_ "crypto/sha1" //nolint:gosec
 	_ "crypto/sha256"
 	"crypto/x509"
 	"encoding/pem"
@@ -76,6 +75,14 @@ func RSASignDigest(digest []byte, digestType crypto.Hash, keyFile, passphrase st
 // RSAVerifyDigest is exported for use in tests and verifies a
 // signature over the provided hash of a message. The key file must be
 // in the PEM format.
+//
+// digestType may be crypto.SHA1, for the legacy apk index signatures that are
+// still in the wild. Note that crypto/sha1 is deliberately not registered as
+// crypto.SHA1 anywhere in apko: every SHA-1 digest apko computes goes through
+// chainguard.dev/apko/internal/sha1cd, which refuses digests of input bearing
+// the signature of a collision attack, and a registered crypto.SHA1 would offer
+// a collision-blind way to compute one. Verification does not need the hash
+// registered, only its digest size.
 func RSAVerifyDigest(digest []byte, digestType crypto.Hash, signature []byte, publicKey []byte) error {
 	if len(digest) != digestType.Size() {
 		return errDigestLength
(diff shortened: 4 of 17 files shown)


pkg/apk/signature/rsa.go
```diff
@@ -76,6 +75,14 @@ func RSASignDigest(digest []byte, digestType crypto.Hash, keyFile, passphrase st
...
 // signature over the provided hash of a message. The key file must be
 // in the PEM format.
+//
+// digestType may be crypto.SHA1, for the legacy apk index signatures that are
+// still in the wild. Note that crypto/sha1 is deliberately not registered as
+// crypto.SHA1 anywhere in apko: every SHA-1 digest apko computes goes through
+// chainguard.dev/apko/internal/sha1cd, which refuses digests of input bearing
+// the signature of a collision attack, and a registered crypto.SHA1 would offer
+// a collision-blind way to compute one. Verification does not need the hash
+// registered, only its digest size.
```
stevebeattie, line 85: This comment's central claim isn't true, and the blank-import removal it justifies doesn't buy anything.

`crypto/sha1` is still linked into every apko binary and its `init()` still calls `crypto.RegisterHash(crypto.SHA1, New)`. It arrives via `crypto/x509` — imported by this very file — plus `crypto/tls`, `github.com/google/uuid`, `go-jose/v3`, `golang.org/x/crypto/ssh`, `keystore-go/v4`, `ProtonMail/go-crypto/openpgp/packet` and `sigs.k8s.io/release-utils/hash`. None of those are removable from a registry client.

Probing from inside package `signature` on this branch:

```
crypto.SHA1.Available() = true
crypto.SHA1.New() works, collision-blind digest = a9993e364706816aba3e25717850c26c9cd0d89d
```

So a collision-blind `crypto.SHA1` is still one call away, and dropping `_ "crypto/sha1"` from this file changed nothing about that.

It does add a small cost. `RSAVerifyDigest(…, crypto.SHA1, …)` now depends on `rsa.VerifyPKCS1v15` never reaching `hash.New()` *and* on x509 continuing to pull sha1 in. The first half holds today — in go1.26 `New()` is only called under `fips140only.Enforced()`, and `crypto.Hash.Size()` is a static table lookup — so the PR body is right about that. But `TestVerifyLegacySHA1IndexSignature` can't actually validate it: the test passes because sha1 is registered transitively, so it would pass even if the reasoning were wrong.

**Suggestion:** restore `_ "crypto/sha1"`, or keep the removal and drop the security rationale from this comment. The real guard against reintroducing plain SHA-1 is **gosec G505**, already enabled in `.golangci.yml` (only G115 is excluded) — worth citing that instead. Two caveats worth a word: G505 is exactly what the `//nolint:gosec` comments this PR removes were suppressing, so it's bypassable the same way; and `gosec` is excluded for `_test.go`, so tests can reintroduce `crypto/sha1` silently.

Note also that the PR body's FIPS reasoning has one ordering slip. In `crypto/rsa/fips.go:344,387` the `hash.New()` inside `fips140only.ApprovedHash(fips140hash.Unwrap(hash.New()))` is evaluated *before* SHA-1 can be rejected, and `crypto.Hash.New()` panics on an unregistered hash. So under `GODEBUG=fips140=only` an actually-unregistered SHA-1 would panic here, not return the intended error. Harmless today for the same reason the rest of this comment is inaccurate — x509 keeps it registered — but it means the removal has downside without upside. (`_ "crypto/sha256"` is genuinely needed for that same path, so it should stay.)

pkg/apk/expandapk/expandapk.go
```diff
@@ -640,13 +650,18 @@ func checkSums(ctx context.Context, r io.Reader) error {
...
 
 		if _, err := io.Copy(w, tr); err != nil {
 			return fmt.Errorf("hashing %s: %w", header.Name, err)
 		}
 
-		if want, got := checksum, w.Sum(nil); !bytes.Equal(want, got) {
+		sum, err := sha1cd.Sum(w)
+		if err != nil {
+			return fmt.Errorf("hashing %s: %w", header.Name, err)
+		}
```
stevebeattie, line 662: Worth calling out a behaviour change here that isn't in the PR description — though on inspection it's smaller than it first looks, and this code is the right response to it.

Adopting sha1cd means any package that legitimately ships a known collision blob — `sha1collisiondetection` or `hashclash` test data, the SHAttered PDFs in a PDF or security-training corpus, AV/YARA sample sets — becomes uninstallable. That's true *independently* of the `ErrCollision` return, because sha1cd rehashes colliding input to a different digest. For the fixture in this PR:

```
crypto/sha1 : f92d74e3874587aaf443d1db961d4e26dde13e9c
sha1cd      : 7117b3cb9225aaf0d8ef1a40e493957b0bf8693d
```

A package built by abuild/apk-tools carries the first value in `APK-TOOLS.checksum.SHA1`, so without this check the failure would surface as `checksum mismatch: … header was f92d74e3…, computed 7117b3cb…` — accurate but baffling. Returning `ErrCollision` turns that into a real diagnosis, so this is an improvement rather than the cause. Two asks:

1. **A line in the PR description.** "A package containing a known SHA-1 collision blob can no longer be installed" is user-visible, and the cause isn't obvious from a build failure. Worth stating next to the Scope section.
2. **Is an escape hatch wanted?** There's no opt-out and the failure is permanent for that package — a re-fetch hits the same detector, and `isRetryableError` is allowlist-based so it correctly declines to retry rather than looping. Probability is low: sha1cd tests 32 specific disturbance vectors rather than heuristics, so false positives on non-adversarial input are negligible, and Wolfi packages are generally stripped of test data. Worth noting this site is defence-in-depth either way — the data section is already authenticated by the signed index → control-section digest → `datahash` (SHA-256) chain, so the per-file SHA-1 tag isn't the trust anchor here. Your call whether that adds up to a flag.

internal/sha1cd/sha1cd.go
```diff
@@ -0,0 +1,83 @@
...
+// That passthrough is for callers holding a hash.Hash whose algorithm is only
+// known at runtime, so that whichever hash they end up with is checked if it is
+// SHA-1. Where the algorithm is known at the call site, call [SumBytes] or that
+// digest's own package instead, so the code does not read as though SHA-1 were
+// involved when it is not.
+func Sum(h hash.Hash) ([]byte, error) {
+	crh, ok := h.(upstream.CollisionResistantHash)
+	if !ok {
+		return h.Sum(nil), nil
+	}
```
stevebeattie, line 67: This passthrough silently permits the exact mistake the package exists to prevent:

```go
sha1cd.Sum(sha1.New())  // returns a digest and a nil error; no collision check, no signal
```

`*sha1.digest` isn't a `CollisionResistantHash`, so it takes the `!ok` branch and comes back looking like a checked digest. The doc comment above asks callers to use `SumBytes` where the algorithm is known statically, but that's a convention, and nothing enforces it. Given that `crypto.SHA1` is still registered in every apko build (see my comment on `rsa.go`) and gosec G505 is both `//nolint`-suppressible and excluded for `_test.go`, this is the whole guarantee resting on documentation.

The `var h hash.Hash = sha1cd.New()` pattern at `expandapk.go:456` is also the only site that actually needs the polymorphism, and even there the two possibilities are known at compile time.

Cheap fix that keeps the sha256 path working — reject any non-detecting hash of SHA-1's size:

```go
crh, ok := h.(upstream.CollisionResistantHash)
if !ok {
	if h.Size() == Size {
		return nil, fmt.Errorf("refusing to finalise a %d-byte hash that cannot detect collisions", Size)
	}
	return h.Sum(nil), nil
}
```

That turns a silent bypass into a loud one at the only place it can happen, and costs nothing for sha256.

Relatedly: `TestSumPassesThroughOtherHashes` pins the positive case, but there's no negative case asserting `Sum` refuses a collision-blind SHA-1 — which is the behaviour that matters here.

pkg/apk/signature/rsa.go
xnox: init() of modules are performed in import order. thus we can replace `_ "crypto/sha1"` import with a direct import of "github.com/pjbgf/sha1cd" maybe? because ideally, i want to ensure that sha1cd is the last registered handler of sha1, even if prior imports indirectly pulled in crypto/sha1 already.

alternatively, I could write custom init() that calls into `internal/sha1cd` to trigger hash registration? that could work too.

I am also very worried about using sha1cd universally - because it creates different hash for the collision detected inputs; but unfortunately that hash is still stable - meaning signature created with sha1cd would verify with sha1cd!

Lack of hash bindings to keys is also very annoying due to pure RSA PKCSv1.5 usage here; as RSA-PSS would be much better - cause then there is hash binding to the key.

pkg/apk/expandapk/expandapk.go
xnox: Note that melange which creates .apks for us has had sha1cd implementation since February this year - intentionally to ensure as many APK-TOOLS.checksum.SHA1 are already known to be sha1cd safe, see:
- https://github.com/chainguard-dev/melange/pull/2357

Anybody who legitimately wants to ship collision blobs; should encode them in some other format - like for example the hex-encoded representation used in this test suite.

Note git also uses sha1cd - and rejects storing collision blobs outright.

internal/sha1cd/sha1cd.go
xnox: ..... and the package was created in the first place because "casts" were assumed to be ugly.

I actually originally wanted to directly import sha1cd library, and indeed cast each hash explicitely. Cause imho that's a better way than an arbitrary internal indirection - which is not obvious directly. Like it was done in melange before at https://github.com/chainguard-dev/melange/pull/2357/changes

Maybe i should rewrite this code with direct sha1cd usage.

pkg/apk/apk/index.go
```diff
@@ -446,11 +448,11 @@ func parseRepositoryIndex(ctx context.Context, u string, keys map[string][]byte,
 		for _, sig := range sigs {
 			// compute the digest if not already done
 			if _, hasDigest := indexDigest[sig.DigestAlgorithm]; !hasDigest {
-				h := sig.DigestAlgorithm.New()
-				if n, err := h.Write(indexData); err != nil || n != len(indexData) {
-					return nil, fmt.Errorf("unable to hash data: %w", err)
+				digest, err := hashIndex(sig.DigestAlgorithm, indexData)
+				if err != nil {
+					return nil, err
```
stevebeattie, line 453: Behaviour change worth noting: this aborts the whole index fetch, where the old code fell through to the next signature on a per-signature failure. Since `indexDigest` memoizes per algorithm, the only way to reach it is a SHA-1 collision on a SHA-1-signed index — and no repo currently dual-signs across algorithms (Wolfi and `apk.cgr.dev` are `RSA256`-only, three keys on the latter; Alpine v3.21/v3.22/edge are `RSA`-only), so with nothing to fall through to, `continue` and `return` reach the same outcome today. Still marginally prefer `continue` alongside the `Warnf` already in the `else` branch below, for consistency with how every other per-signature failure is handled.

Separately, and probably its own PR rather than this one: an index where the SHA-1 signature verifies but the SHA-256 signature *fails* should be rejected outright, and `break`ing on first success currently accepts it. That asymmetry has no benign explanation — a SHA-1 collision preserves the SHA-1 signature and necessarily breaks the SHA-256 one, while corruption would break both — which makes it a strictly better collision detector than counter-cryptanalysis, catching any collision rather than the 32 disturbance vectors in sha1cd's table. It's inert against the current state of the Chainguard and Alpine indexes per above, so nothing regresses and nothing improves today; the value is having it in place before Alpine adds RSA256 alongside RSA, which is exactly when the attack becomes attractive.

pkg/apk/signature/rsa.go
stevebeattie: You're remembering real behaviour, but it was removed upstream. sha1cd used to self-register — `init() { crypto.RegisterHash(crypto.SHA1, New) }` in both `sha1cd.go` and `cgo/sha1.go` — and pjbgf/sha1cd#206 deleted it, merged 2026-04-27 and released in **v0.6.0**. Rationale given: *"This library is primarily used by go-git, which does not rely on the crypto package's registration mechanism… Therefore, it will no longer automatically self-register itself."* So v0.5.0 and earlier would have done what you describe; the v0.6.0 this PR pins does not.

Which makes the import-order approach inert now: `go list -deps` over all of sha1cd's packages contains no `crypto` entry at all, only `klauspost/cpuid`. Importing it registers nothing, so it can't become the last registered handler. And init order is dependency order, not source import order regardless — `crypto/sha1` is a transitive dep of `crypto/x509`, which this file imports, so its `init()` runs first however the import block is arranged.

The history is also the strongest argument against building on this. melange#2357 pinned sha1cd **v0.4.0**, which *did* self-register, so melange was overriding `crypto.SHA1` process-wide from February onward. That override disappeared when melange moved to v0.6.0 — silently, with no compile error and no failing test. A security property that can evaporate in a routine dependency bump isn't a property you want to depend on.

The custom-`init()` variant *would* work — `crypto.RegisterHash` is `hashes[h] = f` with no double-registration panic, and an `init()` in `internal/sha1cd` runs after its transitive deps. I checked that empirically: the override wins, silently. But that's the argument against it, not for it.

**Overriding `crypto.SHA1` would break legitimate SHA-1 verification elsewhere in the process.** x509 and tls both route SHA-1 through the registry on their signature paths — `crypto/x509/x509.go:1009` (`h := hashType.New()` in `checkSignature`, `crypto.SHA1` for `SHA1WithRSA`/`ECDSAWithSHA1`/`DSAWithSHA1`, so CSRs, CRLs and OCSP), `crypto/crypto.go:250` (`crypto.SignMessage`, reached from `x509.go:1599` for signing), and `crypto/tls/auth.go:23` (`verifyHandshakeSignature`, including TLS 1.2's default peer algorithm set). Overriding it and re-verifying an honestly-signed SHA1WithRSA CSR gives `crypto/rsa: verification error`.

Because sha1cd rehashes colliding input, the damage is shaped badly: those paths behave normally on ordinary input and diverge only on adversarial input — in code that has nothing to do with apk. And it still wouldn't cover the direct callers (`x509.go:1755` `sha1.Sum` for SubjectKeyId, `tls/prf.go`, `tls/key_agreement.go`), so you'd get partial coverage plus global breakage.

So I'd keep gosec G505 as the mechanism, and reduce this comment to what's true — something like:

```go
// digestType may be crypto.SHA1, for the legacy apk index signatures that are
// still in the wild. Verification needs only the digest size and algorithm OID,
// not a registered hash: rsa.VerifyPKCS1v15 reaches hash.New() only under
// GODEBUG=fips140=only. Every SHA-1 digest apko *computes* goes through
// chainguard.dev/apko/internal/sha1cd, which refuses digests of input bearing
// the signature of a collision attack; gosec G505 is what keeps crypto/sha1
// from being reintroduced.
```

On the stable-mitigated-digest worry — real critique of sha1cd, and I read it as an argument *for* this PR's shape. Because `Sum` returns `ErrCollision` instead of the digest, apko never uses the mitigated value, so it can't sign or verify over one; melange does the same in #2357, failing the build rather than emitting it. It's also the sharpest reason not to override the registry, since that would hand mitigated digests to libraries that never asked.

On PKCS#1 v1.5 vs PSS — v1.5 already binds the hash algorithm, via the DigestInfo prefix carrying the algorithm OID (`hashPrefixes["SHA-1"]` decodes to 1.3.14.3.2.26). One wrinkle: Go rejects a SHA-1-vs-SHA-256 mixup on the digest-length pre-check before it ever compares prefixes, so demonstrating the OID binding needs a same-length pair — SHA-256 vs SHA-512/256 reaches the prefix comparison and fails there. Either way the binding exists, so PSS wouldn't add it.

What's actually missing is a binding from *key* to *permitted algorithm*: apk reuses one key for `.SIGN.RSA.<key>` and `.SIGN.RSA256.<key>`, so a key can't be declared SHA-256-only. That's an apk format and keyring gap, and it's why a publisher who has moved to SHA-256 stays exposed while still emitting an RSA/SHA-1 signature clients accept — the cross-algorithm rejection point you +1'd above: an index whose SHA-1 signature verifies while its SHA-256 signature fails has no benign explanation, and rejecting it catches *any* collision rather than the 32 disturbance vectors in sha1cd's table.

internal/sha1cd/sha1cd.go
stevebeattie: Agreed the current shape is the problem — but I'd locate it in `Sum`'s *parameter type*, not in the wrapper existing. And I think there's a third option that's better than either of the two you're weighing, including better than what I suggested above.

**Direct usage with explicit casts is worse for exactly this property.** melange#2357 is the evidence: the cast and the flag check are repeated per call site, the error string is duplicated, and `sha1cd.New().(sha1cd.CollisionResistantHash)` is an unchecked assertion that panics if upstream ever changes `New()`'s return type. apko has ~10 sites. There's no chokepoint, so every new site is one forgotten `if collision` away from a silently collision-blind digest — and that's a code-review problem forever rather than a compiler problem once.

The v0.6.0 bump sharpens this. Upstream just changed observable behaviour between v0.5.0 and v0.6.0 (pjbgf/sha1cd#206 removing self-registration). Direct usage spreads that coupling across every call site; a wrapper is the one place to absorb it and the one place you'd notice.

**But don't take my earlier suggestion either.** I proposed a concrete type embedding `upstream.CollisionResistantHash`. That has a trap I missed: embedding promotes a collision-blind `Sum([]byte) []byte`, so `h.Sum(nil)` still compiles and returns `7117b3cb…` with no error. You'd end up with a package `Sum(*Hash) ([]byte, error)` and a method `Sum([]byte) []byte` — same name, opposite safety. Worse than what it fixes. (Embedding also drops `MarshalBinary`, and an exported embedded field lets `&sha1cd.Hash{}` panic or inject an arbitrary implementation.)

**The third option: don't satisfy `hash.Hash` at all.** Checking the call sites, every one consumes the hash purely as an `io.Writer` — `io.Copy` at `install.go:47`, `package_getter.go:314`, `resolveapk.go:50`, `expandapk.go:655`; `io.TeeReader` at `install.go:93`; `io.MultiWriter` at `install_test.go:379`. The only place that names `hash.Hash` is `expandapk.go:456`, and only because one variable is reused for the sha256 branch. Nothing needs `Sum`.

```go
// Hash computes SHA-1 with collision detection.
//
// It is deliberately an io.Writer and not a hash.Hash: finalising goes through
// Finalize, which can report a collision, and a Sum method would offer a
// collision-blind way to finalise it.
type Hash struct{ crh upstream.CollisionResistantHash }

func New() *Hash {
	return &Hash{crh: upstream.New().(upstream.CollisionResistantHash)}
}

func (h *Hash) Write(p []byte) (int, error) { return h.crh.Write(p) }
func (h *Hash) Reset()                      { h.crh.Reset() }
func (h *Hash) Size() int                   { return Size }
func (h *Hash) BlockSize() int              { return h.crh.BlockSize() }

func (h *Hash) Finalize() ([]byte, error) {
	sum, collision := h.crh.CollisionResistantSum(nil)
	if collision {
		return nil, ErrCollision
	}
	return sum, nil
}
```

I built and tested this. `io.Copy`, `io.TeeReader` and chunked writes all work, collision detection fires, `Reset` works, `SumBytes` is unchanged. And both bypasses become compile errors:

```
h.Sum undefined (type *sha1cd.Hash has no field or method Sum)
cannot use h (variable of type *sha1cd.Hash) as hash.Hash value: missing method Sum
```

No casts at any call site, and the single type assertion lives in `New()` where one test covers it.

`expandapk.go:456` needs a small restructure, since it's the one polymorphic site. Something like keeping the writer and the finaliser separate:

```go
sha1h := sha1cd.New()
var h io.Writer = sha1h
finalize := sha1h.Finalize
// …in the maxStreamsReached branch:
sha256h := sha256.New()
h, finalize = sha256h, func() ([]byte, error) { return sha256h.Sum(nil), nil }
```

That also retires the comment at `:454` that currently has to explain why `sha1cd.Sum` is being handed a sha256 hash.

One caller does legitimately use the collision-blind path: `checksums_test.go:67` takes `h.Sum(nil)` to get the mitigated digest for the header, which is what makes `TestCheckSumsRejectsCollision` prove the bypass is closed — the sharpest test in the PR, so worth preserving. Simplest is to pin it as a constant in `sha1cdtest` next to the vector (`7117b3cb9225aaf0d8ef1a40e493957b0bf8693d`), which also means an upstream change to the safe-rehash derivation fails a test instead of passing quietly. `TestSumPassesThroughOtherHashes` retires with the passthrough.

Worth noting the blast radius is contained either way: `internal/sha1cd` is only importable from `chainguard.dev/apko/...`, so no consumer outside the module can be affected. Of the 8 `sha1cd.Sum` call sites, 6 already pass a concrete `sha1cd.New()`; only `expandapk.go:491`/`:526` — the same variable in one loop — use the `hash.Hash` polymorphism.

Net effect on my original point: the "no negative test for the passthrough" gap disappears, because there's nothing left to test — the bypass stops compiling. That's the version of "casts are ugly" I'd argue for: one cast, in one constructor, instead of a convention every future call site has to remember.

Issue #2408, opened 2026-08-18
EROFS follow-ups: mount/umount hardening, layer splitting, go-erofs bump


smoser:
Follow-up work split out of #2249, which was descoped in response to review so
that the single-image writer and `apko erofs ls` could land on their own. Two of
the items below are bugs found in that review, not new features.

Intended order: **#2249 → mount/umount → layering.**

The full pre-descope tree is preserved on the `feat/apko-erofs-full` branch of
https://github.com/smoser/apko, so the follow-ups can branch from working code
rather than reconstruct it.

## 1. Mount / umount (lands next)
this is in #2415 .
Descoped in smoser/apko@1fe041ea: `pkg/erofsmount/mount_linux.go`,
`driver_linux.go`, `state.go`, `stub_other.go` and their tests, plus the
`apko erofs mount` and `apko erofs umount` subcommands.

- [ ] **Validate the state file before unmounting.** `Unmount` executed `umount`,
      typically as root, on whatever paths `<dest>/.apko-erofs-mount.json` named;
      `LoadState` checked only JSON shape and schema version. Anyone able to write
      to `dest` — and the docs suggested shared locations like `/mnt/x` — could
      plant a state file naming `/home` and have root unmount it. Verify
      `st.Dest == dest` and reject any `Mounts` entry that escapes `dest` under
      `filepath.Rel`. Every path the mount side writes is under `dest`
      (`merged`, `layers/NN`), so nothing legitimate is lost.
- [ ] **Make partial-unmount recovery possible.** On an EBUSY partway through,
      the error told the user to rerun `apko erofs umount`. That cannot succeed:
      `Mounts` is `[merged, layers/NN…00]`, so the rerun's first step unmounts
      `merged`, which already came down, and exits 32 before reaching the busy
      layer. The state file is never rewritten on partial success, so recovery
      was manual `umount` plus hand-deleting the state file. Either tolerate a
      not-mounted result per entry, or drop successfully-unmounted entries from
      the state file as it goes.
- [ ] **Default to read-only.** The default mount was writable and `umount`
      `RemoveAll`ed `upper` with no flag or warning, silently discarding
      everything written through the mount. Read-only fits the verification use
      case the docs describe; writes become an opt-in.
- [ ] **Add a test seam.** `Mount`/`Unmount` construct their `Driver` internally,
      so ~305 lines of orchestration (cleanup LIFO, state lifecycle, unmount
[...]

PR #2410, opened 2026-08-18
erofs: first batch of #2408 follow-ups

smoser:
First batch of the EROFS follow-ups from #2408 — the items that touch merged
`main` only, so none of them wait on the descoped mount/umount or layering code
coming back. One real behavior change, two tests, two rewords.

### `isOpaqueDir` fails closed on an unreadable layer

`isOpaqueDir` returned `false` whenever `statOn` failed, so a layer whose xattr
region could not be read was treated as one that hides nothing. That fails
*open*: the entries an opaque directory was meant to hide from lower layers get
listed anyway, and nothing tells the caller the answer was a guess. Layers come
from an OCI image that may be untrusted, so a damaged or hostile one should not
be able to widen the merged view by being unreadable.

It now returns `(bool, error)` and propagates anything that is not
`fs.ErrNotExist` — absent is the one benign case, since a layer that lacks the
directory genuinely is not opaque there. Both callers (`lookup`, `mergeDir`)
already returned errors.

The test puts the unreadable layer in the *middle* of three on purpose: that is
the only position where `isOpaqueDir` is the first thing to fail. With the
broken layer on top, `lookup`'s ancestor stat raises first and the fixture would
pass either way. Against the old code it reports
`ReadDir(etc) = [foo secret], <nil>` — the lower layer's `secret` leaking into
the merged view.

### `application/vnd.erofs+zstd` gets an accurate error

A compressed layer is a spec-legal EROFS image apko cannot read yet, but it hit
the mediaType check and was told the command "only handles EROFS images". Now
any `application/vnd.erofs+<codec>` names its codec and points at #2406.

Matching on the suffix rather than adding a media-type constant is deliberate:
the draft spec's set of codecs isn't something apko should pin down in
`pkg/build/types` to produce one error message.

### `os.features` propagation from a base image config, pinned

`generateIndexWithMediaType` copies the finished config's `os.features` onto the
index platform descriptor for every format, and `BuildImageFromLayers`
DeepCopies the base image's config. So a plain **tar** build on a base image
that already declares `os.features` surfaces them on the index descriptor,
changing that index's digest. Spec §5.4 asks for this and #2249's squash message
calls it out, but the test added there covered only the empty-base case.

Also `require.Empty` → `require.Nil` for the "tar builds declare nothing"
assertion, which is what it means to check — that the propagation doesn't invent
an empty slice. Harmless either way: `os.features` carries `omitempty` in
go-containerregistry, so an empty-but-present slice can't change a digest.

### Two claims softened to match reality

- **§3.7 was cited too strongly.** Its materialize-or-fail rule governs
  *cross-layer* hardlinks; for links within one layer the spec is silent, so
  apko materializing them is conformant but not blessed by that section. #2249's
  squash message has this right; `pkg/build/erofs.go` and `docs/erofs.md` did
  not.
- **`docs/erofs.md` promised kernel parity for `apko erofs ls`** ("the merged
  view the kernel would assemble"). It approximates it, and diverges in two
  corners: a middle-layer whiteout at a directory's own name doesn't cut off
  lower layers when a higher layer recreates the directory, and opacity isn't
  inherited by descendant directories. Both need 2+ layers, so neither is
  reachable for an image apko produces today. The doc now names them and links
  #2408, where the algorithm fix (a per-directory layer horizon) is tracked.

### Verification

`go test ./...` passes, `golangci-lint run -n` reports 0 issues, `gofmt -l` is
clean.

Unrelated note found along the way: `make lint` is currently broken on `main` —
its target installs
`github.com/golangci/golangci-lint/cmd/golangci-lint/v2@v2.2.1`, which no longer
resolves (`invalid version: unknown revision cmd/golangci-lint/v2.2.1`). CI
isn't affected, since `verify.yaml` uses `golangci-lint-action`. Left alone
here; happy to fix it separately.

### Not in this PR

The rest of #2408: restoring mount/umount (with state-file containment,
partial-unmount recovery, a read-only default and a `Driver` test seam),
restoring multi-layer splitting, the layer-horizon fix, and the go-erofs bump.
The pre-descope tree those two build on is preserved at
https://github.com/smoser/apko/tree/feat/apko-erofs-full.

Refs #2408

### DIFF
--- a/pkg/erofsmount/stack.go
+++ b/pkg/erofsmount/stack.go
@@ -15,6 +15,8 @@
 package erofsmount
 
 import (
+	"errors"
+	"fmt"
 	"io"
 	"io/fs"
 	"path"
@@ -222,7 +224,11 @@ func (s *Stack) lookup(name string) (int, error) {
 		}
 		// Not in this layer. If this layer's copy of the parent directory is
 		// opaque, it hides everything below.
-		if s.isOpaqueDir(layer, parent) {
+		opaque, err := s.isOpaqueDir(layer, parent)
+		if err != nil {
+			return -1, err
+		}
+		if opaque {
 			return -1, fs.ErrNotExist
 		}
 	}
@@ -256,7 +262,11 @@ func (s *Stack) mergeDir(name string) ([]fs.DirEntry, error) {
 		}
 		// This layer's own entries are already in; an opaque directory hides
 		// only what lies below it.
-		if s.isOpaqueDir(layer, name) {
+		opaque, err := s.isOpaqueDir(layer, name)
+		if err != nil {
+			return nil, err
+		}
+		if opaque {
 			break
 		}
 	}
@@ -288,22 +298,33 @@ func (s *Stack) isWhiteout(e fs.DirEntry) bool {
 // isOpaqueDir reports whether fsys's copy of dir is an opaque directory --
 // `trusted.overlay.opaque` set to "y", per spec §3.6 -- which hides every
 // lower layer's children of that directory.
-func (s *Stack) isOpaqueDir(fsys fs.FS, dir string) bool {
+//
+// A stat failure is reported, not swallowed: opacity hides content, so
+// answering "not opaque" for a layer we could not read would silently un-hide
+// whatever the image meant to hide. dir being absent from this layer is the one
+// benign case, and means exactly "not opaque here".
+func (s *Stack) isOpaqueDir(fsys fs.FS, dir string) (bool, error) {
 	if !s.whiteouts {
-		return false
+		return false, nil
 	}
 	info, err := statOn(fsys, dir)
-	if err != nil || !info.IsDir() {
-		return false
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return false, nil
+		}
+		return false, fmt.Errorf("stat %s to check for %s: %w", dir, overlayOpaqueXattr, err)
+	}
+	if !info.IsDir() {
+		return false, nil
 	}
 	gx, ok := info.(interface {
 		GetXattr(string) (string, bool)
 	})
 	if !ok {
-		return false
+		return false, nil
 	}
 	v, _ := gx.GetXattr(overlayOpaqueXattr)
-	return v == "y"
+	return v == "y", nil
 }
 
 // rootInfo returns FileInfo for ".". The topmost layer that has a root wins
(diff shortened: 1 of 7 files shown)


pkg/erofsmount/stack.go
```diff
@@ -288,22 +298,33 @@ func (s *Stack) isWhiteout(e fs.DirEntry) bool {
 // isOpaqueDir reports whether fsys's copy of dir is an opaque directory --
 // `trusted.overlay.opaque` set to "y", per spec §3.6 -- which hides every
 // lower layer's children of that directory.
-func (s *Stack) isOpaqueDir(fsys fs.FS, dir string) bool {
+//
+// A stat failure is reported, not swallowed: opacity hides content, so
```
mattmoor, line 313: Take or leave, for symmetry: `isWhiteout` (just above) still fails open on an `e.Info()` error, and the untrusted-layer argument in the PR body applies to it too. The failure mode is milder than the opacity one, which is presumably why it stayed: an unreadable whiteout is treated as a live chardev, which occupies the name (shadowing lower entries rather than leaking them), and a later Stat on the entry re-raises the error. But after this PR the two tombstone checks sit a screen apart with opposite error philosophies. Worth either the same `(bool, error)` treatment or a line in #2408, so the asymmetry is recorded as chosen rather than accidental.

mattmoor (dismissed):
All four commits verified against the code, and the PR body's claims are accurate.

The `isOpaqueDir` change is correct: `fs.ErrNotExist` is the right benign case, both callers propagate, and the middle-layer placement of the broken fixture (so the test discriminates against the old code instead of passing via `lookup`'s ancestor stat) is exactly the right way to avoid a vacuous regression test. The `+zstd` error still refuses the layer, and the suffix-match rationale (don't pin the draft spec's codec set into `pkg/build/types` to phrase one error) is sound. The base-image `os.features` test pins the digest-affecting case I asked about on #2249, and `require.Nil` is the assertion the comment always meant. The §3.7 and `ls` docs rewordings match the spec text and the actual divergences, now named and linked to #2408.

This answers both threads left open on #2249; I'll resolve them pointing here. One symmetry note left inline on `stack.go`, take or leave. The `make lint` breakage on main is worth its own one-liner whenever.

pkg/erofsmount/stack.go
smoser: good point. fixing

mattmoor (approved):
Re-approving at 30f66a5f. The new commit is the suggested treatment done right: isWhiteout matches isOpaqueDir's (bool, error) shape with both callers propagating, the cheap Type() path still avoids the Info() call for non-chardevs and single-layer stacks, and the mirrored test pins both ReadDir and Stat surfacing the error. The absent fs.ErrNotExist special case is correct here: DirEntry.Info() can only return not-exist for a file removed between the directory read and the stat, which cannot happen in an immutable EROFS layer.

PR #2413, opened 2026-08-19
build: make WithAnnotations and WithFormat order-independent

smoser:
* build: make WithAnnotations order-independent
  
  Same fix as the previous commit, for the other field-level Option.
  WithAnnotations copied straight into bc.ic.Annotations, so a later
  WithImageConfiguration or WithConfig replaced the map and dropped the
  annotations entirely. It now records them on the Context and
  resolveImageConfiguration merges them once, after every Option has run,
  which is the ordinary Go shape for functional options: populate an
  intermediate struct, then resolve it into the final configuration.
  
  This makes the code do what pkg/build/options.go has always said it
  does. The doc comment already read "adds annotations from commandline
  to those in the config. Commandline annotations take precedence" --
  true only if the caller happened to order the Options correctly. That
  comment now states the ordering guarantee explicitly, and
  WithImageConfiguration names both field-level Options as exempt from
  being replaced.
  
  Resolving also clones instead of copying in place. Assigning bc.ic
  copies the struct but shares its Annotations map with the caller, so
  the old maps.Copy scribbled into a map the caller still held -- and
  into a map shared by every per-arch Context that NewMultiArch builds
  from one option slice.
  
  Unlike WithFormat, this one has released consumers. Anyone who passed
  WithAnnotations before a whole-config Option was silently getting no
  annotations and will now get them, so the OCI manifest gains entries it
  did not have and the image digest changes. That is the documented
  behavior finally happening, but it is worth a release note for anyone
  asserting on digests.
  
  apko's own CLI is unaffected either way: build.go and publish.go both
  apply WithAnnotations after WithConfig, and the WithImageConfiguration
  that buildImageComponents appends carries an ic that NewOptions already
  derived from the same option slice.
  
* build: make WithFormat order-independent
  
  WithImageConfiguration and WithConfig assign bc.ic outright, so they
  silently discarded the format an earlier WithFormat had asked for. There
  was no error and no empty output: ImageLayoutToLayer branches on
  ic.Format.Resolved(), which defaults to tar, while WithTarball still
  names the output whatever the caller asked for. The result is a
  gzip-framed tar sitting at a path ending in .erofs, which only fails
  later, in whatever reads it.
  
  WithFormat now records the requested format on the Context, and
  resolveImageConfiguration applies it once, after every Option has run.
  This is the ordinary Go shape for functional options -- populate an
  intermediate struct, then resolve it into the final configuration -- as
  in go-containerregistry's remote.makeOptions, which likewise derives
  values and reports option conflicts after the apply loop rather than
  inside individual options. The deviation was the whole-config setters:
  an Option that replaces the entire configuration breaks the invariant
  that each Option owns a disjoint field, and that invariant is what
  makes apply order irrelevant.
  
  pkg/build/options.go now describes this behavior. WithFormat states
  that it wins wherever it appears in the slice, and
  WithImageConfiguration and WithConfig state that they replace the
  configuration rather than merging into it, with the field-level
  overrides exempt.
  
  apko's own CLI was unaffected, but only incidentally:
  buildImageComponents re-injects the ic that NewOptions derived from the
  same option slice, so the discarded value happened to match. Nothing
  tested that. pkg/build/lock.go, meanwhile, appends
  WithImageConfiguration after the caller's options with a
  caller-supplied config, which is unconditionally the losing order; it
  escapes notice only because the format does not affect package
  resolution. The new options_test.go pins both orderings.
  
  WithFormat shipped in 35dcb7e9 and has not been in a release, so no
  consumer can depend on the previous behavior.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -60,6 +60,11 @@ type Context struct {
 	ic types.ImageConfiguration
 	o  options.Options
 
+	// formatOverride is the layer format an explicit WithFormat asked for. It
+	// is held here rather than written straight to ic, and resolved onto ic by
+	// resolveImageConfiguration once every Option has been applied.
+	formatOverride types.LayerFormat
+
 	s6      *s6.Context
 	fs      apkfs.FullFS
 	apk     *apk.APK
@@ -239,6 +244,16 @@ func (bc *Context) checkPaths(ctx context.Context) error {
 	return nil
 }
 
+// resolveImageConfiguration applies the overrides recorded by the field-level
+// Options onto ic. It runs after every Option has been applied, so that an
+// Option which replaces ic wholesale -- WithImageConfiguration, WithConfig --
+// cannot discard them by appearing later in the slice.
+func (bc *Context) resolveImageConfiguration() {
+	if bc.formatOverride != "" {
+		bc.ic.Format = bc.formatOverride
+	}
+}
+
 // NewOptions evaluates the build.Options in the same way as New().
 func NewOptions(opts ...Option) (*options.Options, *types.ImageConfiguration, error) {
 	bc := Context{
@@ -250,6 +265,7 @@ func NewOptions(opts ...Option) (*options.Options, *types.ImageConfiguration, er
 			return nil, nil, err
 		}
 	}
+	bc.resolveImageConfiguration()
 
 	return &bc.o, &bc.ic, nil
 }
@@ -273,6 +289,7 @@ func New(ctx context.Context, fs apkfs.FullFS, opts ...Option) (*Context, error)
 			return nil, err
 		}
 	}
+	bc.resolveImageConfiguration()
 
 	// SOURCE_DATE_EPOCH will always overwrite the build flag
 	if v, ok := os.LookupEnv("SOURCE_DATE_EPOCH"); ok && len(strings.TrimSpace(v)) != 0 {
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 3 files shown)


pkg/build/build.go
```diff
@@ -239,6 +247,25 @@ func (bc *Context) checkPaths(ctx context.Context) error {
...
+	if len(bc.annotationOverrides) > 0 {
+		// Clone: ic.Annotations may still be the caller's map.
+		m := maps.Clone(bc.ic.Annotations)
+		if m == nil {
+			m = make(map[string]string, len(bc.annotationOverrides))
+		}
+		maps.Copy(m, bc.annotationOverrides)
+		bc.ic.Annotations = m
+	}
+}
```
mattmoor, line 267: I think the old semantics appended annotations, where this replaces.  I wonder if this will break things?

smoser: It still appends. The new map is seeded from the configured annotations and the overrides are copied over the top, so a key that only exists in the config survives. `TestWithAnnotationsOrderIndependent/override_before_config` pins exactly that: config `{a: configured, c: kept}` + override `{a: override, b: added}` → `{a: override, b: added, c: kept}`.

It *read* as a replace though, which is fair — `maps.Clone` followed by an assignment looks like one. Pushed 9a91268d rewriting it as allocate / copy the configured annotations / copy the overrides over them. No behavior change, but the merge and the precedence are now visible in the code rather than implied.

What does change is *when* the merge happens, and that's the part with a blast radius. Previously `WithAnnotations` merged into `ic.Annotations` as it stood at that point in the option slice, so a later `WithImageConfiguration` or `WithConfig` discarded the merge wholesale. Anyone ordering it that way was silently getting no annotations at all and will now get them — the manifest gains entries and the digest changes. That's the intended fix rather than a side effect, but it deserves a release note. apko's own CLI is unaffected either way: it applies `WithAnnotations` after `WithConfig`, and `buildImageComponents` re-injects the `ic` that `NewOptions` already derived from the same slice.

mattmoor: Yeah, ok.  I missed the `m := maps.Clone(bc.ic.Annotations)`

mattmoor (approved):

PR #2415, opened 2026-08-19
erofs: add 'apko erofs mount/umount' sub commands.

smoser:
Section 1 of #2408: brings back `apko erofs mount` / `apko erofs umount`,
descoped in 1fe041ea so #2249 could land as a writer plus `ls`, with the four
bugs that review found fixed on top — and, after review here, four more.

#2414 has landed, and `hack/test-erofs.sh` is extended to drive `apko erofs
mount` and `apko erofs umount` — the first thing anywhere that executes this
orchestration for real.

## 1. Restore, with the package API private

Everything restored had no callers outside `pkg/erofsmount`, so it comes back
unexported: the driver interface, `newDriver`, `resolveMode`, the `mode` type
and constants, `mountState`, `stateSchemaVersion`, `statePath`, `writeState`,
`loadState`, `removeState`. `Options` is replaced by an exported
`MountOptions` of plain strings, which is all the CLI needs:

```go
Mount(ctx, src, dest string, opts MountOptions) error
```

`Ls` keeps main's signature — a plain arch string, not an options bag — and
does **not** get its `--mode` flag back; it was documented as "accepted for
symmetry with mount and ignored".

The first commit deliberately restores the known-broken behavior; the ones
after it fix it. Splitting it the other way would put the root-`umount` bug on
`main` in between.

## 2. A test seam

`Mount` and `Unmount` built their driver internally, so ~300 lines of
orchestration could not be driven with a fake — `driver_linux_test.go` covered
only the argv builders, and nothing anywhere executed the orchestration.

A `driverFactory` and unexported `mountWith`/`unmountWith` take it; the
exported entry points pass `newDriver`. The kernel-vs-fuse umount choice moves
into the driver as an `Unmount(ctx, mp)` method. The fake records the mode it
is asked for, so the tests observe rather than assume the two places the mode
comes from: the `--mode` flag on the way up, the state file on the way down.

## 3. Validate the state file before unmounting — the security one

`Unmount` read `<dest>/.apko-erofs-mount.json` and handed every path it named
to `umount`, as root in kernel mode. `loadState` checked only JSON shape and
schema version. The state file lives *inside* `dest`, and the docs suggest
shared destinations — so anyone who could write there could plant a file
naming `/home` and have root unmount `/home`.

Three layers, because the first two are not enough on their own:

- **A whitelist, not containment.** Every `Mounts` entry must be absolute and
  a path a mount actually creates: `<dest>/merged` or `<dest>/layers/NN`. The
  recorded `Dest` must equal the dest being unmounted, and no entry may repeat
  (a repeat is inside dest, but unmounting it twice wedges the teardown).
  Containment alone would still accept `<dest>/upper` or `<dest>/anything`,
  and a whitelist subsumes traversal, since a path that climbs out with `..`
  matches neither shape.
- **`umount(2)` with `UMOUNT_NOFOLLOW`, not `umount(8)`.** The whitelist
  constrains the path *string*; `umount(8)` canonicalizes, so `ln -s /home
  <dest>/merged` beside a well-formed state file redirects a root umount with
  nothing to validate against. No check-then-exec can close that — whatever
  the check saw, the exec resolves again. The syscall flag refuses a symlinked
  final component in the same call that unmounts, so there is no window.
- **A resolution check for the rest.** `UMOUNT_NOFOLLOW` covers the final
  component only; a symlinked *parent* (`<dest>/layers` itself) is resolved
  during the kernel's own path walk. `checkResolved` rejects that, comparing
  against dest with dest's own symlinks resolved so a dest legitimately
  reached through one still works.

The mount side had the same hole in reverse: `MkdirAll` is satisfied by a
symlink to an existing directory, so `<dest>/merged` pointing at `/etc` would
have the layer mounted over `/etc`. `ensureDir` now insists on a real
directory.

**The residual is stated rather than papered over,** in the code comment and
in `docs/erofs.md`: the parent-component check and the unmount it guards are
two steps, so someone who can write inside `dest` can still race them. Closing
that wants `openat2(RESOLVE_NO_SYMLINKS)`, which is a much larger change. The
guidance stays "use a `DEST` only you can write to".

## 4. Make a partial unmount recoverable

On EBUSY part-way through, the error said to rerun `apko erofs umount`. That
could never work: `Mounts` is `[merged, layers/NN…00]` and the file was never
rewritten, so the rerun started at `merged` — already gone — and failed there
before ever reaching the busy layer.

Entries are now dropped as they come down, and the file is rewritten with
what's left before returning. A rejected path gets a different message: no
amount of waiting clears a symlink, so only a real umount failure carries the
retry hint.

## 5. Read-only by default

The default mount was writable, and `umount` then `RemoveAll`'d
`<dest>/upper` with no flag and no warning, silently discarding everything
written through it.

`MountOptions.ReadOnly` becomes `MountOptions.Writable`, so the zero value is
the safe one. `--read-only` becomes `--rw`. `upper` and `work` are only
created when writable, and a single-layer image now skips overlayfs by
default — that is every image apko can currently produce.

`umount` uses `os.Remove` on `upper`, not `os.RemoveAll`, which is the point
of the choice: it cannot delete a non-empty directory, so writes survive by
construction rather than by remembering not to delete them. An `upper` nothing
was written through goes away, so the dest stays reusable; one that holds
something stays, and a later `--rw` mount refuses to start on top of it rather
than stacking two sessions' writes — possibly from different images —
together.

## 6. Three more found in review here

- **Escape overlayfs option paths.** Every path in the option string derives
  from `DEST` and none were escaped. overlayfs splits on `,` and lowerdir on
  `:`, honouring `\` as an escape. A `,` fails loudly; a `:` does not — it
  turns one lowerdir into two and can compose a wrong stack.
- **Report both umount failures in fuse mode.** `fuseDriver.Unmount` tries
  kernel umount then fusermount, and discarded the first error. Since it is
  also the blob fall-back, a busy blob mount on a host without fusermount
  reported "neither fusermount3 nor fusermount found in PATH" instead of the
  real reason.
- **Claim `dest` with `O_CREATE|O_EXCL`** rather than stat-then-write, so
  "refuse to clobber an existing mount" is true rather than a race between two
  mounts aimed at one dest.

## 7. Drive it from `hack/test-erofs.sh`

Unit tests go through a fake driver, which cannot catch a wrong `mount(8)`
invocation or a layout overlayfs rejects. The privileged job from #2414 ran
`mount -t erofs` directly, so nothing anywhere ran this orchestration against
a real kernel. Added after the existing `ls`-vs-kernel comparison:

- **Read-only image mount** — the single-layer short-circuit, the state file's
  `mode`/`dest`/`writable`/`mounts`, and a diff of the mounted tree against
  the `apko erofs ls` listing the script already computed.
- **`--rw`** — the overlay layout and LIFO order; a file written through the
  mount must still be in `upper/` after `umount`, a second `--rw` mount at
  that dest must be refused, and a round trip that writes nothing must leave
  no `upper` and be repeatable.
- **A raw blob**, which carries no state file, so `umount` takes the
  single-mountpoint fall-back.
- **Three tampered-state cases**: a decoy named outside `DEST`, a symlinked
  `DEST/merged`, and — the only thing that actually exercises
  `UMOUNT_NOFOLLOW` — a symlink pointed straight at a live tmpfs and handed in
  as `DEST`, so nothing validates it first. Without the flag that tmpfs comes
  down. Each asserts the decoy is still mounted and the planted file
  untouched; asserting on a nonzero exit alone would also pass for a crash.

Cleanup reads `/proc/self/mounts` and unmounts everything under the workdir,
deepest first, so a failure mid-test does not wedge the job.

## Verification

`gofmt -l` clean, `golangci-lint run -n` reports 0 issues, `go build ./...`
and `GOOS=darwin go build ./...` both succeed (the non-Linux stubs still line
up), `SOURCE_DATE_EPOCH=0 go test ./...` passes, and `shellcheck` is clean on
the script.

Each fix was checked to fail without its change, by reverting just the fixed
lines and re-running. The one thing a unit test cannot pin is
`UMOUNT_NOFOLLOW` — unmounting a symlink to a non-mountpoint fails with or
without it — which is why the script covers it against a live mount instead.

The EROFS job is green with all of the above, and its sudo audit log shows
only two `umount(8)` executions in the whole run, both the script's own: apko
itself no longer execs it.

Two items from review are deferred to #2408 rather than fixed here: the state
file goes stale during a *succeeding* teardown (distinguishing "not mounted"
from "busy" needs `/proc/self/mountinfo`, since neither exit 32 nor `EINVAL`
does), and the fuse path has never executed anywhere, because CI runs as root.

Refs #2408


pkg/erofsmount/state.go
```diff
@@ -0,0 +1,185 @@
...
+
+// validate checks a state file before any path it names is handed to umount,
+// which in kernel mode runs as root. The file lives inside dest, so anyone who
+// can write there controls its contents; unvalidated, a planted file naming
+// /home would have root unmount /home.
+//
+// Rather than only checking containment, this requires each entry to be one of
+// the paths Mount actually creates. Nothing legitimate is rejected: mounts are
+// made at <dest>/merged and <dest>/layers/NN and nowhere else.
+func (s *mountState) validate(dest string) error {
```
mattmoor, line 159: This kills the tampered-file attack, and the traversal corpus in the tests is thorough; I threw `layers/../..`, `layers/0/../x`, sibling-prefix dests, backslashes, and relative paths at it and everything is rejected. But the check is purely lexical, and the attacker in this function's own threat model ("anyone who can write to dest") doesn't need to touch the file: `ln -s /home <dest>/merged` next to a well-formed state file passes validation, and `umount(8)` canonicalizes symlinks, so root unmounts `/home`. Same preconditions as the original bug, same impact, one step removed. (`RemoveAll` is not similarly exposed; it's lstat-based.)

The clean fix is to stop exec'ing `umount(8)` for these entries and call `unix.Unmount(mp, unix.UMOUNT_NOFOLLOW)`, which atomically refuses a symlink at the final component, no TOCTOU. That still leaves a symlinked parent (`layers` itself as a symlink), which a `Lstat` walk or openat2-style resolution would close. If the call is instead to accept the residual for a dev tool, the comment above ("requires each entry to be one of the paths Mount actually creates") and docs/erofs.md's claim should both say that holds for the file but not for the directory, so the boundary is recorded as chosen.

pkg/erofsmount/mount_linux.go
```diff
@@ -0,0 +1,303 @@
...
+	// any umount fails, stop: layer mounts that come after a still-pinned
+	// overlay would only return EBUSY noise, and continuing past an error
+	// would also leave the state file out of sync with reality. The user
+	// can rerun `apko erofs umount` after addressing whatever is keeping
+	// the mount busy.
+	// Drop each entry as it comes down and rewrite the state file if one
+	// fails, so the file always describes what is still mounted. Without that
+	// a rerun starts again at merged -- already gone -- and fails there
+	// without ever reaching the entry that was busy.
+	for len(st.Mounts) > 0 {
```
mattmoor, line 270: The busy case is fixed, and the test pins the mechanism. One residual, follow-up caliber: the file is rewritten only on failure, so during a succeeding teardown it goes stale at the first successful umount and stays stale until `removeState`. A SIGKILL or power loss anywhere in that window (which spans every subsequent umount exec) leaves entries for already-unmounted paths, and the rerun then dies at the first one with exit 32, rewrites the same list, and never progresses; "not mounted" is not tolerated anywhere, and kernel umount exits 32 for both busy and not-mounted, so tolerating it needs a `/proc/self/mountinfo` check rather than exit-code sniffing (treat "not present in mountinfo" as already done). Worth a line in #2408 as the remaining half of this fix.

pkg/erofsmount/mount_linux_test.go
```diff
@@ -0,0 +1,423 @@
...
+		// unmount the same path twice.
+		return fmt.Errorf("umount %s: not mounted", mp)
+	}
+	f.unmounted = append(f.unmounted, mp)
+	return nil
+}
+
+// factory hands the same fake back for any mode.
+func (f *fakeDriver) factory() driverFactory {
+	return func(mode) (driver, error) { return f, nil }
```
mattmoor: The factory discards its `mode` argument, so nothing in the unit layer can catch `unmountImage` ignoring `st.Mode` or `unmountBlob` regressing from its deliberate `modeFuse` choice; a regression there would pass every test here and only break for real fuse users. That is the testing-double failure mode: the fake certifies the mode plumbing as correct without observing it. Cheap fix: have the factory record the requested mode (or take a `wantMode` and error on mismatch) and assert it in the blob-fallback and unmount-image tests.

docs/erofs.md
```diff
@@ -174,25 +174,43 @@ The merge approximates what the kernel would assemble, and diverges in two corne
...
 mkdir -p /mnt/apko-erofs
+apko erofs mount out/blobs/sha256/$LAYER /mnt/apko-erofs
+ls /mnt/apko-erofs/
+file /mnt/apko-erofs/bin/sh
+apko erofs umount /mnt/apko-erofs
+```
+
+If the kernel mount mode complains "unknown filesystem type 'erofs'", the kernel module is missing on your system; install it (e.g. `linux-modules-extra-$(uname -r)` on Ubuntu) or pass `--mode=fuse` to use `erofsfuse`, which does not require root and works inside CI containers that lack the kernel module.
+
+`umount` works from `DEST/.apko-erofs-mount.json`, which `mount` wrote. It only accepts mountpoints that a mount creates under `DEST` — `DEST/merged` and `DEST/layers/NN` — so a tampered file cannot redirect a root `umount` elsewhere. Still, prefer a `DEST` only you can write to: anything else lets another user decide what your `umount` takes down within it.
```
mattmoor: "a tampered file cannot redirect a root `umount` elsewhere" is true of the file but not of the directory: a symlink at `<dest>/merged` redirects it with a fully well-formed state file (see the note on `validate`). The following sentence ("prefer a DEST only you can write to") is the honest statement of the real trust boundary; this clause currently overpromises relative to it. Reword once the symlink disposition is decided.

mattmoor (commented):
I verified all five sections of the PR body against the code, and they hold: the restored surface is private with `Mount`/`Unmount`/`MountOptions` as the only new exports, the seam is behavior-preserving (argv builders byte-identical to the pre-descope code), the resumability test is honest in exactly the way that matters (the fake's exit-32 mirroring means the rerun passes only because of the state rewrite; delete the rewrite and it goes red), read-only-by-default and upper preservation check out end to end, and the erofs CI job provably executed all four new script sections on this head, from a fork, including the behavioral decoy assertion rather than message-matching. The four bugs from the #2249 review are fixed as described.

I'm holding approval on one substantive finding, inline on `state.go`: the whitelist validates strings, and the attacker in its own threat model can reroute a root umount with a symlink instead of a tampered file. Disposition there (hardening or an explicitly documented residual) and I'm happy to approve; two independent verification passes converged on it, so I don't think it's a misread.

Smaller notes, none blocking, take or file with #2408:

- `validate` accepts duplicate `Mounts` entries (e.g. `merged` twice). No escape, since both are inside dest, but the second umount exits 32 and wedges the teardown into the rewrite loop.
- `unmountBlob` routes through the fuse driver, so on a box without fusermount the returned error is "neither fusermount3 nor fusermount found in PATH", swallowing the kernel umount's actual stderr (e.g. "not mounted"). The pre-descope inline code reported both.
- The tampered-state script section accepts any nonzero exit from `apko erofs umount`; the decoy-still-mounted assertion narrows it, but pinning the refusal (state file untouched, or the validate error text) would distinguish "refused" from "crashed first".
- The fuse path still never executes against reality anywhere: CI runs as root, so erofsfuse, fuse-overlayfs, and the fusermount fallback chain are covered only by argv-shape tests. Fine to leave noted in #2408.

pkg/erofsmount/state.go
smoser: Adopted the syscall. `fb85605f` replaces every `umount(8)` exec with `umount(2)` + `UMOUNT_NOFOLLOW`: `kernelDriver.Unmount`, the teardown closures from `MountLayer`/`AssembleOverlay`, and the kernel attempt `fuseDriver.Unmount` makes first. `buildKernelUmountArgs` is gone. The final component is now refused in the same syscall that unmounts, so there is no window.

The symlinked-parent half you called out is still check-then-syscall, so `checkResolved` stays for it, and its comment plus `docs/erofs.md` now record which half is atomic and which is a narrowed race instead of claiming the whole thing is closed.

No unprivileged test can tell the flag from its absence — unmounting a symlink to a directory that is not a mountpoint fails either way — so `hack/test-erofs.sh` proves it instead: a symlink pointed at a live tmpfs, handed in as `DEST` so nothing validates it first. Without the flag that tmpfs comes down. From the last EROFS run:

```
Error: umount /tmp/.../decoy-link: refusing to follow a symlink
/usr/bin/fusermount3 -u /tmp/.../decoy-link: exit status 1: ... Invalid argument
```

(`EINVAL` also means "not a mountpoint", so the refusal is named explicitly rather than surfaced as a bare "invalid argument". The `lstat` behind that is for the message only.)

pkg/erofsmount/mount_linux.go
smoser: Agreed this is not for here — filed to #2408.

One wrinkle worth recording: `fb85605f` made it slightly harder to detect, not easier. `umount(2)` returns `EINVAL` for "not a mountpoint", which is exactly as ambiguous as exit 32 was, so it still wants the `/proc/self/mountinfo` check rather than errno sniffing.

The duplicate-entries note from your summary is fixed here, though: `7f28264b` rejects a mountpoint listed more than once in `validate`, for the reason you gave.

pkg/erofsmount/mount_linux_test.go
smoser: Fixed in `00c9edc1`. The factory records each mode it is handed. The blob fall-back test asserts `modeFuse`, and a new test walks a mount and unmount together, checking the flag decides the first and the state file decides the second.

docs/erofs.md
smoser: Reworded in `fb85605f`, now that the disposition is settled. It now says the whitelist stops a path *named* outside `DEST`, that `UMOUNT_NOFOLLOW` refuses a symlink at `DEST/merged` with no window, and that a symlinked parent is a separate check with a real race — which is what makes the "use a `DEST` only you can write to" sentence the operative one rather than a hedge against the clause above it.

pkg/erofsmount/driver_linux.go
```diff
@@ -0,0 +1,330 @@
...
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/chainguard-dev/clog"
+	"golang.org/x/sys/unix"
+)
+
+// driver wraps the externally-invoked mount and umount commands used by Mount.
+// Two implementations exist on Linux: kernelDriver shells out to mount(8) and
```
mattmoor, line 33: Stale half of what 28acdfd7 fixed: kernelDriver no longer shells out to umount(8); unmounts go through `umount(2)` directly. Worth updating so the doc comment and Preflight tell the same story.

pkg/erofsmount/driver_linux.go
```diff
@@ -0,0 +1,330 @@
...
+	opts := overlayOpts(lowers, upper, work, readOnly)
+	if readOnly {
+		opts += ",ro"
+	}
+	// merged is its own argv element rather than an option value, so it is
+	// passed through unescaped.
+	return []string{"mount", "-t", "overlay", "-o", opts, "overlay", merged}
+}
+
+func buildFuseOverlayArgs(lowers []string, upper, work, merged string, readOnly bool) []string {
```
mattmoor, line 298: Take or leave: this feeds kernel-style escapes (`\:`, `\,`, `\\`) to fuse-overlayfs, whose own lowerdir colon-split has historically not honored `\:` the way the kernel does. A dest containing `:` was broken on the fuse leg before this PR too, so no regression there, but a dest containing a bare backslash is a narrow new fuse-leg-only regression (it previously passed verbatim). One comment acknowledging the parity assumption, or a quick check against the fuse-overlayfs option parser, would cover it.

pkg/erofsmount/driver_linux.go
```diff
@@ -0,0 +1,330 @@
...
+// mode may itself be a kernel overlay (when overlayfs over FUSE worked) or a
+// fuse-overlayfs mount, and per-layer mounts are erofsfuse; umount handles the
+// first, fusermount -u the other two.
+//
+// Both failures are reported. Unmount is also the fall-back for a blob mount,
+// whose mode is unknown, so the kernel attempt is the one that failed for the
+// interesting reason -- returning only the fusermount error turns "target is
+// busy" into "neither fusermount3 nor fusermount found in PATH".
+func (d *fuseDriver) Unmount(ctx context.Context, mp string) error {
+	kerr := kernelUnmount(ctx, mp)
```
mattmoor, line 170: Take or leave: when `kernelUnmount` fails with the symlink refusal, fusermount is still exec'd on the same path. It is safe (fusermount has carried its own internal `UMOUNT_NOFOLLOW` since the CVE-2010-3879 fix, and the CI decoy-link log shows it refusing), and state-file paths pass `checkResolved` first, so this is purely surface-shrinking: skipping the fusermount attempt when the kernel failure was the symlink refusal would also keep that refusal from being buried in a joined error.

mattmoor (approved):
Approving at 28acdfd7. I re-verified the whole response round: the held finding is settled with three layers rather than the minimum ask. `UMOUNT_NOFOLLOW` via `umount(2)` closes the final-component race atomically in every path that consumes state-file entries (kernel driver, both teardown closures, the fuse driver's kernel attempt; `buildKernelUmountArgs` is gone), `checkResolved` covers the symlinked-parent half with the residual race documented honestly in code and docs, and the mount side refuses a symlinked mountpoint up front. The e2e script now proves all three refusals behaviorally, asserting the decoy tmpfs remains mounted and the state file survives, and the EROFS job ran those sections on this head.

All six secondary notes from the prior round are fixed as well, several beyond what was asked: the `O_EXCL` state-file claim kills the mount-side race while the partial-unmount rewrite correctly keeps the atomic CreateTemp+rename, duplicate mount entries are rejected, fuse umount failures are joined instead of swallowed, the fake driver now observes the mode it is handed with real assertions, the rerun hint is only offered when a rerun can help, and the upperdir and option-escaping hardening are both welcome. The last commit also cleared the stale `umount(8)` preflight requirement I was about to note.

Three take-or-leave nits inline; none blocking.

PR #2426, opened 2026-08-21
erofs: small cleanups to erofsmount tool

smoser:
The three take-or-leave nits from the review of #2415, all in
`pkg/erofsmount/driver_linux.go`. One commit each.

### 1. The interface comment describes an unmount that no longer happens

`28acdfd7` dropped the `umount(8)` requirement from `Preflight` but left the
`driver` doc comment saying `kernelDriver` shells out to it. Text only.

### 2. Stop at a symlink refusal in fuse mode

`fuseDriver.Unmount` tries the kernel unmount and falls back to `fusermount`,
which is right when the mount was simply made by the other tool and wrong when
the kernel refused because the mountpoint is a symlink. `fusermount` has
carried its own `UMOUNT_NOFOLLOW` since the CVE-2010-3879 fix, so it refuses
the same path again and the joined error buries the reason:

```
Error: umount .../decoy-link: refusing to follow a symlink
/usr/bin/fusermount3 -u .../decoy-link: exit status 1: ... Invalid argument
```

That second line is noise once the first has decided the matter. `kernelUnmount`
now wraps a sentinel for the symlink case so `Unmount` can return it as-is.
State-file paths pass `checkResolved` before reaching a driver, so this shrinks
surface rather than fixing an exposure.

### 3. Refuse paths fuse-overlayfs cannot round-trip

`escapeOverlayPath` emits the kernel's escaping to both legs, and review asked
whether fuse-overlayfs honours it. **Measured against 1.17 rather than
assumed** — mounting a tree whose path contains each character, escaped and
verbatim, against a clean-path control:

| in `DEST` | escaped (today) | verbatim (before #2415) |
| --- | --- | --- |
| *(control)* | ok | ok |
| `,` | **ok** | broken — split into separate options |
| `:` | broken | broken |
| `\` | broken | broken |

Two things follow. The escaping is a straight win for the comma on this leg and
cannot help with the other two. And it is **not** the regression it looked
like: a bare backslash was already eaten before the escaping went in, so that
dest fails identically before and after.

What is left is a bad failure mode rather than a wrong mount — fuse-overlayfs
reports `cannot resolve path .../dd`, naming a path the user never wrote. So
`:` and `\` are refused up front on the fall-back leg, with the reason and a
pointer at `--mode=kernel`, which handles them correctly. The kernel overlay
attempted just before the fall-back is unaffected: only the tool that cannot
cope declines.

The probe script is not included — it needs a live `fuse-overlayfs` and the
result is a fact about that tool, not about apko, so it is recorded as a table
in the comment on `fuseOverlayUnsupported` instead.

### Verification

`gofmt -l` clean, `golangci-lint run -n` reports 0 issues, `go build ./...` and
`GOOS=darwin go build ./...` both succeed, and `SOURCE_DATE_EPOCH=0 go test
./...` passes. The fuse-mode symlink test was checked to fail without its
change.

Both new unit tests do run in CI, in the `go-tests` job on the non-root
runner — where the symlink test passes via the EPERM+lstat path rather than
EINVAL, since the kernel checks `may_mount()` before `path_mounted()`. What
never runs anywhere is the **e2e** fuse leg: `hack/test-erofs.sh` goes through
sudo, so `resolveMode` picks kernel. That is one of the two items already
filed against #2408.

### Review follow-ups

Four commits addressing the review, one concern each:

- `06801ac9` — move `errSymlinkedMountpoint` above `kernelUnmount`'s doc block
  (it had captured the whole UMOUNT_NOFOLLOW rationale as its own godoc), and
  fix the two comments commit 2 made untrue: the lstat is no longer "for the
  message only", and "Both failures are reported" is now conditional. The
  lstat comment records the EPERM-vs-EINVAL detail, so a later cleanup keying
  it on EINVAL does not silently break the fuse leg.
- `64166080` — run `checkFuseOverlayPaths` before the `fuse-overlayfs`
  `LookPath`, since the refusal is terminal and "not installed" is fixable.
  With that order an empty-PATH test can pin the wiring: raw paths, and
  read-only excluding upper/work. Mutation-checked both ways.
- `cabc8693` — `--mode=kernel (as root)`; the user on this leg is non-root by
  construction, so the bare suggestion bounces off `kernelDriver.Preflight`.

On the fifth comment: yes, the refusal is deliberately version-blind rather
than a versioned check — "refuse until re-measured". There is no probe, and
1.17 is what the comment records.

Refs #2408

### DIFF
--- a/pkg/erofsmount/driver_linux.go
+++ b/pkg/erofsmount/driver_linux.go
@@ -172,6 +172,12 @@ func (d *fuseDriver) Unmount(ctx context.Context, mp string) error {
 	if kerr == nil {
 		return nil
 	}
+	// A symlinked mountpoint is a refusal, not a sign that the wrong tool was
+	// tried. fusermount carries its own UMOUNT_NOFOLLOW and would only refuse
+	// it again, so running it adds nothing and buries the reason in a join.
+	if errors.Is(kerr, errSymlinkedMountpoint) {
+		return kerr
+	}
 	fm, err := lookupFusermount()
 	if err != nil {
 		return errors.Join(kerr, err)
@@ -234,6 +240,11 @@ func buildKernelLayerArgs(blob, mp string) []string {
 // This covers the final component only. A symlinked *parent* -- <dest>/layers
 // itself -- resolves during the syscall's own path walk, and checkResolved is
 // what rejects that, with the narrower race it documents.
+// errSymlinkedMountpoint reports a mountpoint whose final component is a
+// symlink. UMOUNT_NOFOLLOW refuses those, and callers use this to tell the
+// refusal apart from an ordinary unmount failure.
+var errSymlinkedMountpoint = errors.New("refusing to follow a symlink")
+
 func kernelUnmount(ctx context.Context, mp string) error {
 	clog.FromContext(ctx).Debugf("umount2: %s (UMOUNT_NOFOLLOW)", mp)
 	err := unix.Unmount(mp, unix.UMOUNT_NOFOLLOW)
@@ -245,7 +256,7 @@ func kernelUnmount(ctx context.Context, mp string) error {
 	// argument". This is for the message only -- the flag above is what
 	// provides the guarantee, so an lstat racing it cannot weaken anything.
 	if fi, lerr := os.Lstat(mp); lerr == nil && fi.Mode()&os.ModeSymlink != 0 {
-		return fmt.Errorf("umount %s: refusing to follow a symlink", mp)
+		return fmt.Errorf("umount %s: %w", mp, errSymlinkedMountpoint)
 	}
 	return fmt.Errorf("umount %s: %w", mp, err)
 }
--- a/pkg/erofsmount/driver_linux.go
+++ b/pkg/erofsmount/driver_linux.go
@@ -202,17 +202,19 @@ func (d *fuseDriver) AssembleOverlay(ctx context.Context, lowers []string, upper
 		}, nil
 	}
 
-	if _, err := exec.LookPath("fuse-overlayfs"); err != nil {
-		return nil, fmt.Errorf("kernel overlay failed and fuse-overlayfs is not installed: %w", err)
-	}
 	// Only the fall-back needs this. The kernel overlay attempted just above
 	// unescapes the way escapeOverlayPath escapes; fuse-overlayfs does not.
+	// Ahead of the LookPath below because a path fuse-overlayfs cannot
+	// round-trip is terminal, while "not installed" is fixable.
 	paths := slices.Clone(lowers)
 	if !readOnly {
 		paths = append(paths, upper, work)
 	}
 	if err := checkFuseOverlayPaths(paths...); err != nil {
 		return nil, err
+	}
+	if _, err := exec.LookPath("fuse-overlayfs"); err != nil {
+		return nil, fmt.Errorf("kernel overlay failed and fuse-overlayfs is not installed: %w", err)
 	}
 	fArgs := buildFuseOverlayArgs(lowers, upper, work, merged, readOnly)
 	if err := runCmd(ctx, fArgs[0], fArgs[1:]...); err != nil {
--- a/pkg/erofsmount/driver_linux.go
+++ b/pkg/erofsmount/driver_linux.go
@@ -164,7 +164,8 @@ func (d *fuseDriver) MountLayer(ctx context.Context, blob, mp string) (func() er
 // fuse-overlayfs mount, and per-layer mounts are erofsfuse; umount handles the
 // first, fusermount -u the other two.
 //
-// Both failures are reported. Unmount is also the fall-back for a blob mount,
+// Both failures are reported, unless the kernel refused a symlinked
+// mountpoint -- see below. Unmount is also the fall-back for a blob mount,
 // whose mode is unknown, so the kernel attempt is the one that failed for the
 // interesting reason -- returning only the fusermount error turns "target is
 // busy" into "neither fusermount3 nor fusermount found in PATH".
@@ -239,6 +240,11 @@ func buildKernelLayerArgs(blob, mp string) []string {
 	return []string{"mount", "-t", "erofs", "-o", "ro", blob, mp}
 }
 
+// errSymlinkedMountpoint reports a mountpoint whose final component is a
+// symlink. UMOUNT_NOFOLLOW refuses those, and callers use this to tell the
+// refusal apart from an ordinary unmount failure.
+var errSymlinkedMountpoint = errors.New("refusing to follow a symlink")
+
 // kernelUnmount unmounts mp with umount(2) rather than umount(8).
 //
 // UMOUNT_NOFOLLOW is the reason: it makes the kernel refuse a symlink at the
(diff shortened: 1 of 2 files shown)


pkg/erofsmount/driver_linux.go
```diff
@@ -233,6 +250,11 @@ func buildKernelLayerArgs(blob, mp string) []string {
 // This covers the final component only. A symlinked *parent* -- <dest>/layers
 // itself -- resolves during the syscall's own path walk, and checkResolved is
 // what rejects that, with the narrower race it documents.
+// errSymlinkedMountpoint reports a mountpoint whose final component is a
+// symlink. UMOUNT_NOFOLLOW refuses those, and callers use this to tell the
+// refusal apart from an ordinary unmount failure.
+var errSymlinkedMountpoint = errors.New("refusing to follow a symlink")
```
mattmoor: This var and its comment landed between `kernelUnmount`'s doc comment and the func — so the whole UMOUNT_NOFOLLOW rationale block above is now godoc for `errSymlinkedMountpoint`, and `kernelUnmount` is undocumented. Move the var (with its own comment) above the `kernelUnmount` comment block, blank line between.

pkg/erofsmount/driver_linux.go
```diff
@@ -244,7 +266,7 @@ func kernelUnmount(ctx context.Context, mp string) error {
 	// argument". This is for the message only -- the flag above is what
 	// provides the guarantee, so an lstat racing it cannot weaken anything.
 	if fi, lerr := os.Lstat(mp); lerr == nil && fi.Mode()&os.ModeSymlink != 0 {
-		return fmt.Errorf("umount %s: refusing to follow a symlink", mp)
+		return fmt.Errorf("umount %s: %w", mp, errSymlinkedMountpoint)
```
mattmoor: "This is for the message only" is no longer true: `fuseDriver.Unmount` now branches on the sentinel this lstat produces. Worth rewording, and one detail is worth capturing while doing so: on the non-root leg the errno here is EPERM, not EINVAL — the kernel checks `may_mount()` (CAP_SYS_ADMIN) before `path_mounted()`, so a non-root caller never sees EINVAL for a symlink. The errno-blind lstat is therefore what makes the classification work on exactly the leg commit 2 targets; a future cleanup that keys the sentinel on EINVAL would break it silently (the unit test, which runs non-root in CI, would be the only thing that notices). `fuseDriver.Unmount`'s "Both failures are reported." doc line is also now conditionally false.

pkg/erofsmount/driver_linux.go
```diff
@@ -196,6 +204,15 @@ func (d *fuseDriver) AssembleOverlay(ctx context.Context, lowers []string, upper
 	if _, err := exec.LookPath("fuse-overlayfs"); err != nil {
 		return nil, fmt.Errorf("kernel overlay failed and fuse-overlayfs is not installed: %w", err)
 	}
+	// Only the fall-back needs this. The kernel overlay attempted just above
+	// unescapes the way escapeOverlayPath escapes; fuse-overlayfs does not.
+	paths := slices.Clone(lowers)
+	if !readOnly {
+		paths = append(paths, upper, work)
+	}
+	if err := checkFuseOverlayPaths(paths...); err != nil {
```
mattmoor, line 213: Consider running this check before the LookPath above: the path refusal is terminal (installing fuse-overlayfs won't fix a `:` in the dest) while "not installed" is fixable, and with that order an empty-PATH unit test can pin this wiring — raw paths, readOnly excluding upper/work — the way `TestFuseDriverUnmountStopsAtASymlinkRefusal` pins Unmount's. Right now `TestCheckFuseOverlayPaths` covers only the pure function.

pkg/erofsmount/driver_linux.go
```diff
@@ -299,6 +321,30 @@ func buildFuseOverlayArgs(lowers []string, upper, work, merged string, readOnly
...
+//
+// So escapeOverlayPath earns its keep for the comma and cannot help with the
+// other two. Refuse those rather than hand fuse-overlayfs an option string it
+// will misread and then fail on with a path the user never named.
+const fuseOverlayUnsupported = `:\`
+
+func checkFuseOverlayPaths(paths ...string) error {
+	for _, p := range paths {
+		if i := strings.IndexAny(p, fuseOverlayUnsupported); i >= 0 {
+			return fmt.Errorf("fuse-overlayfs cannot handle %q in a path (%s); use --mode=kernel, or a dest without it",
```
mattmoor: The user on this leg is non-root by construction (resolveMode picks fuse when euid != 0, and the kernel overlay just failed above), so `--mode=kernel` verbatim bounces off `kernelDriver.Preflight`'s euid check. Maybe "use --mode=kernel (as root), or a dest without it".

pkg/erofsmount/driver_linux.go
```diff
@@ -299,6 +321,30 @@ func buildFuseOverlayArgs(lowers []string, upper, work, merged string, readOnly
...
+// control:
+//
+//	,  escaped works; verbatim is split into separate options
+//	:  broken both ways -- \: is not honoured as the kernel honours it
+//	\  broken both ways -- eaten, so the path never resolves
+//
+// So escapeOverlayPath earns its keep for the comma and cannot help with the
+// other two. Refuse those rather than hand fuse-overlayfs an option string it
+// will misread and then fail on with a path the user never named.
+const fuseOverlayUnsupported = `:\`
```
mattmoor, line 345: Deliberate that this refuses on every fuse-overlayfs version with no probe, keyed to the 1.17 measurement? Fine if so — the comment records the version — just confirming the intent is "refuse until re-measured" rather than a versioned check.

mattmoor (commented):
Nits only — the substance checked out: the sentinel can't strand a real mount (a race-swapped symlink means the mountpoint was already unmounted, and fusermount refuses symlinks itself), and the new check is a raw-string check on exactly the strings that reach fuse-overlayfs's `-o`, applied only on the leg that can't cope.

One PR-body correction: "none of this executes in CI" is overbroad. The *e2e* fuse leg never runs (the script goes through sudo → kernel mode), but both new unit tests do run in the `go-tests` job on the non-root runner — where the symlink test passes via the EPERM+lstat path, not EINVAL (see the inline comment).

pkg/erofsmount/driver_linux.go
smoser: Moved in `06801ac9` — the var and its own comment now sit above the `kernelUnmount` block, blank line between, so the UMOUNT_NOFOLLOW rationale is godoc for the func again.

pkg/erofsmount/driver_linux.go
smoser: Reworded in `06801ac9`. The comment now says the lstat classifies as well as describes, and records the errno detail explicitly: as root EINVAL, but a non-root caller gets EPERM because the kernel checks `may_mount()` before `path_mounted()` — so keying the sentinel on EINVAL would stop classifying on exactly the leg the branch exists for. `Unmount`'s "Both failures are reported" is now conditional on that branch.

pkg/erofsmount/driver_linux.go
smoser: Done in `64166080` — the path check now runs first, with a comment saying why (terminal vs. fixable). `TestAssembleOverlayChecksPathsBeforeLookPath` uses the empty PATH to pin the wiring: a `:` in a lowerdir is refused rather than deferred to LookPath, a `,` is not (which pins that the check sees raw paths — the escaped form carries a backslash and would be refused), and read-only leaves upper/work out of the set. Mutation-checked both ways: reverting the order fails two cases, checking escaped paths fails one.

pkg/erofsmount/driver_linux.go
smoser: Good catch — `cabc8693` makes it `use --mode=kernel (as root), or a dest without it`.

pkg/erofsmount/driver_linux.go
smoser: Deliberate, yes: "refuse until re-measured" rather than a versioned check. There is no probe and no version gate — 1.17 is only what the comment records as the measurement it is keyed to. If a later release fixes `:` or `\`, the table and the constant get revisited together.

pkg/erofsmount/driver_linux_test.go
```diff
@@ -164,6 +165,119 @@ func TestOverlayArgsEscapeSeparators(t *testing.T) {
...
+		{
+			name:     "read-only does not check upper or work",
+			lowers:   []string{"/mnt/x/layers/00"},
+			upper:    "/mnt/x:y/upper",
+			work:     "/mnt/x:y/work",
+			readOnly: true,
+			want:     "not installed",
+		},
+		{
+			name:   "writable checks upper",
```
mattmoor, line 233: Nit: the writable leg puts the `:` only in `upper`, so a mutation that appends `upper` but drops `work` still passes the suite. A symmetric "writable checks work" case would close it.

mattmoor (approved):
All three response commits verified: the moved var reattaches the UMOUNT_NOFOLLOW rationale to `kernelUnmount` as godoc; the reworded lstat comment matches the kernel's `can_umount()` ordering (root EINVAL, non-root EPERM) and correctly warns off an EINVAL-keyed sentinel; and the new ordering test kills both mutations — swapping the check back after LookPath yields "not installed" for the colon case, and the comma case pins raw-path checking since the escaped form would carry a refused backslash. Per-commit sweep of all six commits builds and passes tests on darwin and GOOS=linux, so the stack bisects cleanly. One test nit inline, not worth holding the PR for.

PR #2439, opened 2026-08-27
sbom: make internal SPDX relationships reproducible

kevinmdavis:
Sort the top-level elements copied from internal APK SBOMs before adding their relationships to the image SBOM. This prevents equivalent builds from producing different predicates due to Go map iteration order.


jdolitsky (approved):

xnox (approved):
Note, syft enumeration of components is often unstable. When this happens we report bugs to them, but yes overall we too should use stable sort even when syft isn't.

PR #2441, opened 2026-08-31
apk: feed installation from an abstract PackageContents

mattmoor:
Installation consumed *expandapk.APKExpanded concretely, coupling it to the expanded-APK carrier even though the lazy install path only needs the ordered install records and a filesystem to open their content. Introduce PackageContents as the per-package installation input, adapt the expanded APK to it, and generalize installPackage/lazilyInstallAPKFiles over it.

The control-shaped surface is one method: ControlSection returns the compressed control segment exactly as distributed, and installation derives everything from those bytes — their SHA1 is the package checksum apk records (by definition), and script/trigger extraction reads the decompressed stream. Carriers therefore cannot disagree with themselves about the checksum of the control data they present.

InstallPackageContents installs exactly a caller-supplied, pre-settled package list from its contents — no index, no resolution, no fetch — and build.WithPreResolvedPackages exposes that to image builds, alongside the existing lockfile path. Downstream callers can then carry package contents in forms of their own choosing without apko learning those forms.

### DIFF
--- a/pkg/apk/apk/contents.go
+++ b/pkg/apk/apk/contents.go
@@ -1,0 +1,80 @@
+// Copyright 2026 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apk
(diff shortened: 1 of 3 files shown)


pkg/build/options.go
```diff
@@ -263,6 +263,18 @@ func WithLockFile(lockFile string) Option {
 	}
 }
 
+// WithPreResolvedPackages provides the exact package set to install, in
+// order, with the contents each member installs from. The build installs
+// precisely these: no index is consulted and no dependency resolution
+// happens. The image configuration's package list still names the requested
+// world (written to /etc/apk/world); this option settles how it is satisfied.
+func WithPreResolvedPackages(contents []apk.PackageContents) Option {
```
markusthoemmes: This feels like a separate concern. Maybe split that into a separate change?

mattmoor: will push in a subsequent change!

markusthoemmes (approved):

PR #2448, opened 2026-08-31
build: install a pre-resolved package list from supplied contents

mattmoor:
InstallPackageContents installs exactly a caller-supplied, pre-settled package list from its contents — no index is consulted, no dependency resolution happens, and nothing is fetched; the package checksum is derived from each carrier's control section (its SHA1, by definition). The installed-database tail is shared with InstallPackages via recordInstalled, so both installers write the database through one code path.

build.WithPreResolvedPackages exposes this to image builds as a third way to satisfy the configured world, beside the existing lockfile path and FixateWorld: the configuration's package list still names the requested world; the option settles how it is satisfied. Downstream callers can then carry package contents in forms of their own choosing without apko learning those forms.

### DIFF
--- a/pkg/build/build_implementation.go
+++ b/pkg/build/build_implementation.go
@@ -152,7 +152,13 @@ func (bc *Context) buildImage(ctx context.Context) ([]apk.InstalledDiff, error) 
 		pkgs []apk.InstalledDiff
 		err  error
 	)
-	if bc.o.Lockfile != "" {
+	switch {
+	case len(bc.o.PreResolvedPackages) > 0:
+		pkgs, err = bc.apk.InstallPackageContents(ctx, &bc.o.SourceDateEpoch, bc.o.PreResolvedPackages)
+		if err != nil {
+			return nil, fmt.Errorf("failed installation from pre-resolved packages: %w", err)
+		}
+	case bc.o.Lockfile != "":
 		log.Debugf("Using lockfile: %s", bc.o.Lockfile)
 		lock, err := lock.FromFile(bc.o.Lockfile)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/build/build_implementation.go
```diff
@@ -152,7 +152,13 @@ func (bc *Context) buildImage(ctx context.Context) ([]apk.InstalledDiff, error)
 		pkgs []apk.InstalledDiff
 		err  error
 	)
-	if bc.o.Lockfile != "" {
+	switch {
+	case len(bc.o.PreResolvedPackages) > 0:
```
tcnghia, line 156: Could we distinguish an explicitly supplied empty set from an absent option here? `WithPreResolvedPackages([])` promises to install precisely that set without consulting indexes, but `len == 0` falls through to the lockfile or `FixateWorld` path. An empty resolved world therefore still fetches and resolves repositories (and can fail offline) instead of performing a zero-package install.

pkg/build/options.go
```diff
@@ -263,6 +263,18 @@ func WithLockFile(lockFile string) Option {
...
 }
 
+// WithPreResolvedPackages provides the exact package set to install, in
+// order, with the contents each member installs from. The build installs
+// precisely these: no index is consulted and no dependency resolution
+// happens. The image configuration's package list still names the requested
+// world (written to /etc/apk/world); this option settles how it is satisfied.
+func WithPreResolvedPackages(contents []apk.PackageContents) Option {
+	return func(bc *Context) error {
+		bc.o.PreResolvedPackages = contents
```
tcnghia, line 273: What about `NewMultiArch` callers? It reuses this same option for every architecture, so one contents slice is installed into every target context. `InstallPackageContents` does not compare `PackageInfo.Arch` with the context architecture, which lets an x86_64 slice be installed into the aarch64 image without an error. Could we key these contents by architecture, or reject incompatible multi-architecture use?

tcnghia (approved):

mattmoor:
I have claude looking into the comments to follow up, since this auto-merged.

PR #2462, opened 2026-09-04
apk: stop bypassing the resolver cache for superseded indexes

markusthoemmes:
A PkgResolver built over a set of index objects is correct for that set regardless of whether a newer generation has been fetched since. The currency gate treated freshness as a cache-validity question and made concurrent resolutions rebuild the resolver over the full index almost every time. Replace the tries and the purge-on-rotation logic with flat LRU maps keyed by index identity so stale generations age out instead.

### DIFF
--- a/pkg/apk/apk/shameful_global_caches.go
+++ b/pkg/apk/apk/shameful_global_caches.go
@@ -16,205 +16,121 @@ package apk
 
 import (
 	"context"
+	"fmt"
 	"maps"
+	"reflect"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
+
+	lru "github.com/hashicorp/golang-lru/v2"
 
 	apkometrics "chainguard.dev/apko/pkg/metrics"
 )
 
+// maxResolverCacheEntries bounds the number of distinct index combinations
+// each derived cache retains. The live set is tiny (a handful of repos times a
+// couple of arches), so the bound only matters for superseded generations
+// that in-flight resolutions keep requesting: those age out once nothing asks
+// for them anymore.
+const maxResolverCacheEntries = 16
+
+// indexIdentity returns a key that is unique to the index object itself. A
+// NamedIndex is immutable, so anything derived from a specific set of index
+// objects stays valid for as long as those objects are around. Remote indexes
+// are deduplicated by (url, etag) in globalIndexCache, so the same generation
+// always yields the same object.
+func indexIdentity(idx NamedIndex) string {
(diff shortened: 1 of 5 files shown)


pkg/apk/apk/shameful_global_caches.go
```diff
@@ -18,203 +18,108 @@ import (
...
-	// A generation replacement that raced the fill has already run its
-	// purge, so purge its index ourselves.
-	for _, idx := range indexes {
-		if !globalIndexCache.isCurrent(idx) {
-			r.forgetIndex(idx)
+// getOrFill returns the cached value for indexes, computing and inserting it
+// on a miss. The boolean reports whether the value was already cached.
+func (c *lruCache[V]) getOrFill(indexes []NamedIndex, fill func() V) (V, bool) {
+	c.Lock()
+	defer c.Unlock()
```
kevinmdavis, line 59: I wonder if we'd want to singleflight this or do a two pass locking approach where we avoid holding the lock while we're running `newPkgResolver`? `newPkgResolver` seems like it can sometimes take several hundred milliseconds.

(I find that kind of surprising on its own, but we may want to account for that)

pkg/apk/apk/shameful_global_caches.go
```diff
@@ -18,203 +18,108 @@ import (
...
-
-type resolverCache struct {
+// maxResolverCacheEntries bounds the number of distinct index combinations
+// each derived cache retains. A multi-arch resolution inserts one resolver
+// entry per arch and one disqualify entry, so the live set is that times the
+// number of distinct repo combinations resolved concurrently. Superseded
+// generations that in-flight resolutions still request stay cached and age
+// out once enough newer combinations have been inserted. Every entry pins a
+// whole index generation, so raising this trades memory for hit rate.
+const maxResolverCacheEntries = 16
```
kevinmdavis, line 33: My initial hunch is that we'd want this to be higher, especially when accounting for multiple architectures. Thoughts?

pkg/apk/apk/shameful_global_caches.go
markusthoemmes: Done via a `sync.OnceValue` deferring to outside the lock. Good shout!

pkg/apk/apk/shameful_global_caches.go
markusthoemmes: Raised to 64 for now.

kevinmdavis (approved):

PR #2463, opened 2026-09-04
apk: bound global caches with LRU eviction

kevinmdavis:
Bound completed entries in apko's shared flight caches and the parsed index cache with hashicorp/golang-lru. Capacities are set at each initialization site so they can be tuned independently, while concurrent loads remain coalesced and recently used entries stay hot. The derived resolver and disqualify caches are left as is since #2462 bounds those separately.

### DIFF
--- a/pkg/apk/apk/cache.go
+++ b/pkg/apk/apk/cache.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/hashicorp/golang-lru/v2/simplelru"
 	"go.opentelemetry.io/otel"
 	"golang.org/x/sync/singleflight"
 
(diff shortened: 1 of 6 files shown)


pkg/apk/apk/cache.go
```diff
@@ -26,72 +26,90 @@ import (
...
 
 	"chainguard.dev/apko/pkg/paths"
 )
 
 type flightCache[K comparable, V any] struct {
-	mux   sync.RWMutex
-	cache map[K]func() (V, error)
+	mux     sync.Mutex
+	flights map[K]*flightCacheEntry[V]
+	lru     *simplelru.LRU[K, V]
```
markusthoemmes: Can we store `func() (V, error)` in the LRU instead of `V`? I think that'd make a lot of extra complexity go away.

pkg/apk/apk/package_getter.go
```diff
@@ -101,11 +101,11 @@ type PackageGetter interface {
 	GetPackage(ctx context.Context, pkg InstallablePackage) (*expandapk.APKExpanded, error)
 }
 
+const packageCacheMaxEntries = 64
```
markusthoemmes: This feels a little small.

kevinmdavis: Bumped to 256.

markusthoemmes (approved):

PR #2480, opened 2026-09-08
fix(apk/fs): make memfs Lstat report symlinks

### LINKED ISSUE #1543: Replaces of symlinks does not appear to work
with the modified wolfi-base as shown in https://github.com/wolfi-dev/os/pull/43832/files , which has two subpackages, that both ship the same symlink, to different destinations and with replaces declared of one another, one can install them with apk tools, but not with apko:

```
$ make package/wolfi-base
$ make local-wolfi

# apk add package-with-symlink
(1/1) Installing package-with-symlink (1-r8)
OK: 14 MiB in 16 packages
# readlink /etc/some.conf 
../run/target1.conf

# apk add package-with-symlink-replaces
(1/1) Installing package-with-symlink-replaces (1-r8)
OK: 14 MiB in 17 packages
# readlink /etc/some.conf 
../run/target2.conf
# exit



$ make test/wolfi-base
Testing package wolfi-base with version wolfi-base-1-r8 from file wolfi-base.yaml
/home/xnox/go/bin/melange test wolfi-base.yaml --repository-append /home/xnox/wolfi-dev/os/packages --keyring-append local-melange.rsa.pub --arch x86_64 --pipeline-dirs ./pipelines/ --repository-append https://packages.wolfi.dev/os --keyring-append https://packages.wolfi.dev/os/wolfi-signing.rsa.pub --test-package-append wolfi-base --debug  --source-dir ./wolfi-base/
2025/02/26 11:31:22 INFO building test workspace in: '/tmp/melange-guest-815115064-main' with apko
2025/02/26 11:31:22 [DEBUG] GET https://packages.wolfi.dev/os/apk-configuration
2025/02/26 11:31:22 INFO setting apk repositories: [/home/xnox/wolfi-dev/os/packages https://packages.wolfi.dev/os]
2025/02/26 11:31:22 INFO image configuration:
2025/02/26 11:31:22 INFO   contents:
2025/02/26 11:31:22 INFO     build repositories: []
2025/02/26 11:31:22 INFO     runtime repositories: []
2025/02/26 11:31:22 INFO     keyring:      []
2025/02/26 11:31:22 INFO     packages:     [package-with-symlink-replaces wolfi-base]
2025/02/26 11:31:22 INFO installing package-with-symlink (1-r8)
2025/02/26 11:31:22 INFO installing package-with-symlink-replaces (1-r8)
2025/02/26 11:31:22 INFO ERROR: failed to test package. the test environment has been preserved:
2025/02/26 11:31:22 INFO   workspace dir: /tmp/melange-workspace-2334097068
2025/02/26 11:31:22 INFO   guest dir: /tmp/melange-guest-815115064
2025/02/26 11:31:22 ERRO failed to test package: unable to build guest: unable to generate image: installing apk packages: installing packages: installing package-with-symlink-replaces (ver:1-r8 arch:x86_64): unable to install files for pkg package-with-symlink-replaces: unable to install symlink from etc/some.conf -> ../run/target2.conf: symlink ../run/target2.conf /tmp/melange-guest-815115064-main/etc/some.conf: file exists
make: *** [Makefile:115: test/wolfi-base] Error 1
```

Why is replaces not being honored by apko for symlinks, when apk-tools does?
Note that replaces is honored by apko for regular files.

Eljees:
Fixes #1543.

`memfs.Lstat` called `getNode`, which resolves symlinks in *every* path component including the last one. So `Lstat` on a path that is itself a symlink returned the target's `FileInfo` and never set the `ModeSymlink` bit — exactly the caveat spelled out in the comment that used to sit above `Lstat`. Any caller that checks "is this already a symlink?" before replacing it therefore saw a regular file, which is what the issue reports.

`getNodeCountLinks` now takes a `followFinal` flag. `getNode` (and therefore `Stat`) passes `true` and behaves exactly as before; `Lstat` goes through a variant that passes `false`, so a symlink in the **final** component is returned as itself while intermediate components are still resolved — POSIX `lstat(2)` semantics. Symlink targets encountered during resolution are always followed fully, regardless of the outer call.

Three tests in `pkg/apk/fs/memfs_lstat_test.go`:

* `Lstat` on a symlink reports `ModeSymlink` — fails on `main`, passes here;
* `Stat` on the same path still resolves to the target (mode and size), so the existing behaviour is pinned;
* `Lstat` on a regular file reached through an intermediate directory symlink (`/usr/lib64 -> lib`) still resolves that intermediate component and does not report a symlink.

`go test ./pkg/apk/fs/...`, `go vet` and `gofmt` are clean.

### DIFF
--- a/pkg/apk/fs/memfs.go
+++ b/pkg/apk/fs/memfs.go
@@ -54,19 +54,36 @@ func NewMemFS() FullFS {
 	}
 }
 
-// getNode returns the node for the given path. If the path is not found, it
-// returns an error.
+// getNode returns the node for the given path, following a symlink in the final
+// path component (as Stat would). If the path is not found, it returns an error.
 func (m *memFS) getNode(path string) (*node, error) {
-	return m.getNodeCountLinks(path, 0)
+	return m.getNodeCountLinks(path, 0, true)
 }
-func (m *memFS) getNodeCountLinks(path string, linkDepth int) (*node, error) {
+
+// getNodeNoFollowFinal is like getNode, but if the final path component itself
+// names a symlink, it returns the symlink node itself rather than resolving it --
+// matching POSIX lstat(2) semantics. Intermediate path components are still
+// resolved through symlinks normally, exactly like getNode.
+func (m *memFS) getNodeNoFollowFinal(path string) (*node, error) {
+	return m.getNodeCountLinks(path, 0, false)
+}
+
+func (m *memFS) getNodeCountLinks(path string, linkDepth int, followFinal bool) (*node, error) {
 	if path == "/" || path == "." {
 		return m.tree, nil
 	}
 	parts := strings.Split(path, pathSep)
+	// find the index of the last non-empty part, so we know when we are
+	// resolving the final path component (relevant only when !followFinal).
+	lastIdx := -1
+	for i, part := range parts {
+		if part != "" {
+			lastIdx = i
+		}
+	}
 	node := m.tree
 	traversed := make([]string, 0)
-	for _, part := range parts {
+	for i, part := range parts {
 		if part == "" {
 			continue
 		}
@@ -82,8 +99,10 @@ func (m *memFS) getNodeCountLinks(path string, linkDepth int) (*node, error) {
 		if !ok {
 			return nil, os.ErrNotExist
 		}
-		// what if it is a symlink?
-		if childNode.mode&os.ModeSymlink != 0 {
+		// what if it is a symlink? Resolve it unless this is the final path component
+		// and the caller asked us not to follow it (lstat semantics).
+		isFinalComponent := i == lastIdx
+		if childNode.mode&os.ModeSymlink != 0 && (followFinal || !isFinalComponent) {
 			newDepth := linkDepth + 1
 			if newDepth > maxLinks {
 				return nil, fmt.Errorf("maximum symlink depth exceeded")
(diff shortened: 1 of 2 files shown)


coreydaley-cg:
`Lstat` semantics are right and `Stat` is untouched; the tests pin both. One thing before this can close #1543: every `Lstat` caller in the tree is an existence check, and the replace logic in `install.go:221` still uses `Stat` and then tests `ModeSymlink`, which `Stat` can never report. So this makes `Lstat` correct without connecting it to the path #1543 is about. Either switch that call to `Lstat` here, or drop the "Fixes" and leave #1543 open.

Also, the new test file needs the license header the rest of the repo carries.

Eljees:
Both points addressed, and you were right that the PR did not reach what #1543 is about.

**The `install.go` side.** `installAPKFiles`'s `tar.TypeDir` case is meant to accept a path that already exists as a symlink to a directory, but it asked `Stat` for the mode. `Stat` resolves the final component, so it can never report `ModeSymlink`, and every such path fell through to `MkdirAll` instead — the branch was unreachable. It now asks `Lstat` for the entry itself and keeps `Stat` for the target, which still has to be a directory. The separate `Stat` that decides whether the directory already existed, and so whether its metadata gets set below, is untouched: it is about the resolved path, and swapping it would change what happens to a dangling symlink.

**The test needed a second look.** My first version only asserted that the symlink survived, and it passed with `Stat` too — `MkdirAll` over a path that already resolves to a directory is a no-op in memory, so the symlink was still there either way. That is the same weakness you flagged, so I checked it the same way you would: put `Stat` back, keep every test, and see what fails. Nothing did.

What the branch actually decides is whether `MkdirAll` is reached at all, so `TestInstallAPKFilesKeepsSymlinkToDirectory` now records that. With `Stat` restored it fails:

```
--- FAIL: TestInstallAPKFilesKeepsSymlinkToDirectory
    install_test.go:741:
        Error:    []string{"lib"} should not contain "lib"
        Messages: an existing symlink to a directory has to be accepted as is, without reaching MkdirAll
```

and with `Lstat` it passes. The layout it uses is the merged-usr one apko builds for itself in `InitDB`: `usr/lib` is the directory, `lib` the symlink to it.

**The license header** is on `pkg/apk/fs/memfs_lstat_test.go` now.

Also rebased onto current `main` (`a770072`); `go test ./pkg/apk/apk/ ./pkg/apk/fs/` and `gofmt -l` are clean on Go 1.27.1.

PR #2493, opened 2026-09-10
tarfs: store a compact record per member and materialize tar.Header on request

### LINKED ISSUE #2492: tarfs: Entry retains a full tar.Header per file; store a compact record
`tarfs.New` (`pkg/apk/expandapk/tarfs/tarfs.go:207`) keeps an `Entry` per tar member holding a full `tar.Header` copy, including `PAXRecords` and `Xattrs` maps and the header's string fields, plus a `fs.FileInfo`. Callers that cache expanded packages keep these indexes alive for the process lifetime, so every file of every cached package carries the full header.

In a 7-day averaged heap profile of a long-running process, `tarfs.New` accounts for 0.9 GB in-use, of which `archive/tar.parsePAX` is 350 MB and `readHeader` another 260 MB. Those are PAX maps and header strings retained after parsing, not the index itself.

A compact entry (name, size, mode, uid/gid, mod time, typeflag, linkname, offset) with the PAX map dropped once the header is resolved would likely halve the row. `FileInfo` can be derived on demand from the compact record. Anything that needs a full `tar.Header` back can re-read it from the offset, which is what the reader path already does for content.

coreydaley-cg:
`tarfs.New` kept a full `tar.Header` per member, twice: once copied into `Entry.Header` and once more through `fi`, which held `hdr.FileInfo()` on the reader's original header. Each member also carried its `PAXRecords` map, which for apk packages is one record (the checksum) costing ~320 bytes of map to hold ~55 bytes of string. Callers that cache expanded packages keep these indexes alive for the process lifetime. Measured on real apks, that came to 700-1,200 bytes resident per member, and in a heap profile of a long-running process `tarfs.New` accounted for 0.9 GB in-use (#2492).

## Change

The index now holds a compact `entry` per member: the header's scalar fields, `Uname`/`Gname` and PAX keys interned through `unique`, PAX records as a slice, and access/change times behind a pointer that is nil for the apk case. `entry` implements `fs.FileInfo` and `fs.DirEntry`, with the mode computed once from `Header.FileInfo()` so it matches `archive/tar` exactly, so `Stat`, `ReadDir`, and `fs.WalkDir` never touch a `tar.Header`.

The record is 136 bytes. Everything a typical apk member does not have sits behind one pointer that is nil for a regular file or directory: access and change times, device numbers, PAX records other than the checksum, and any mtime that `int64` nanoseconds cannot hold (`archive/tar` builds every `ModTime` with `time.Unix`, so `time.Unix(0, n)` reproduces it exactly and the guard is that reconstruction, not a range check). Every relocated field keeps its original width. The apk checksum record is hex-decoded into a fixed field when it is the exact lowercase spelling apk-tools writes, an idea from #2433, and re-encoded on materialization; any other spelling stays a verbatim PAX record so the header still round-trips byte for byte.

The exported API is unchanged. `Entry`, `Entry.Header`, `Entries()`, and `File.Entry` keep their types; they are now materialized from the compact record on request.

Two runtime behaviours do change, neither documented before. `Entries()` used to return the resident records directly, so callers shared one mutable `PAXRecords` map and a mutation was visible to the next caller (and racy between concurrent ones); each call now returns independent copies. And `Sys()` builds a header per call, so it no longer returns a stable pointer — `io/fs` promises no identity, but `archive/tar` happened to. `Entries()` builds its slice on each call, so a caller pays for full headers only while holding the result. The one in-tree consumer, `contents.Entries()`, already copies into `[]tar.Header` and is unaffected. `Sys()` still returns a `*tar.Header`.

Materialized headers reproduce `PAXRecords` and the deprecated `Xattrs` view exactly as `archive/tar` fills them, since the install path reads the apk checksum and `SCHILY.xattr.*` records from there.

## Measured

Resident heap per indexed member, real apks, `runtime.MemStats` around `tarfs.New`:

| apk (aarch64) | members | before | after |
|---|---|---|---|
| git 2.55.0 | 241 | 1,028 B | 285 B |
| coreutils 9.11 | 113 | 1,016 B | 266 B |
| glibc 2.43 | 35 | 889 B | 268 B |
| openssl 3.6.4 | 19 | 743 B | 274 B |
| ca-certificates-bundle | 15 | 803 B | 302 B |
| busybox 1.38.0 | 12 | 788 B | 733 B |
| python-3.13 | 12 | 748 B | 250 B |
| bash 5.3 | 10 | 1,246 B | 891 B |
| curl 8.22.0 | 10 | 743 B | 261 B |
| wolfi-base | 5 | 723 B | 281 B |
| **weighted, 472 members** | | **979 B** | **303 B** |

The resident record is 136 bytes; the rest is per-member strings and index slots. `BenchmarkNew` and `BenchmarkOpen` are in the package so both figures can be re-derived.

Resident is the number that matters here. Materializing costs transient allocation on the read path: collecting every member's header, which is what `contents.Entries()` does once per package install, runs ~840 B per member against ~230 B on main, and each `Open` builds the `File.Entry` it exposes, about three more allocations per regular file. That is garbage as soon as the caller is done with it, where the per-member record is retained for as long as the package is cached. Index build time is within run-to-run noise of main on the same sample, with fewer bytes allocated.

### What that means at scale

Measured above: **979 to 303 bytes per indexed member**, weighted across 10 real Wolfi packages (472 members total; `wolfi-base`, `busybox`, `glibc`, `ca-certificates-bundle`, `python-3.13`, `git`, `curl`, `openssl`, `bash`, `coreutils`).

Linear extrapolation from that figure, since the record is per member and the saving scales with retained members:

| resident indexed members | before | after | saved |
|---|---|---|---|
| 10,000 | 10 MB | 3 MB | 7 MB |
| 100,000 | 98 MB | 30 MB | 68 MB |
| 500,000 | 490 MB | 152 MB | 338 MB |
| 1,000,000 | 979 MB | 303 MB | 676 MB |

To place yourself on it: the sample averages ~48 indexed members per package (median 14, max 229 for `git`). A one-shot build that indexes a 50-package image sits near the top of the table and will not notice. A long-running process that caches expanded packages accumulates members for its lifetime and lands further down — the 1,000,000 row is roughly the 0.9 GB `tarfs.New` figure from the heap profile in #2492.

Nothing here changes for a caller that indexes a package and drops it. The saving is entirely in what stays resident.

## Equivalence, end to end

apko built from `main` and from this branch, same locked package versions, same `SOURCE_DATE_EPOCH`:

| build | result |
|---|---|
| layered aarch64 (`origin` strategy, non-root `run-as`, explicitly owned paths) | `image.tar` byte-identical |
| every blob in that OCI layout | identical, `diff -r` clean |
| unlayered x86_64, single tar layer | byte-identical |
| `build-minirootfs` plain tar, 399 members | byte-identical |
| ownership and mode of every member in that tar | identical |
| SBOMs (~247 KB each) | identical except the line where apko stamps its own build version |

The layered config runs as a non-root user with owned paths, so the uid/gid path is live rather than all-zero. `build-minirootfs` is the one caller that reads the layer file directly, so it covers the plain-tar output contract rather than only the compressed blob.

Re-verified at the final commit with a two-arch (aarch64 and x86_64) layered build: all 28 blobs, both configs, the manifest, and `index.json` identical by content; SBOMs identical except the line carrying apko's own version. The outer `image.tar` differs only in entry order in the two-arch case, which is pre-existing nondeterminism in the multi-arch tarball writer and not touched here.

## Tests

The package had none. Added a randomized round-trip property test plus fixture tests.

`TestRoundTripProperty` generates headers across the field space — full-width uid/gid, every typeflag, device numbers, PAX records including empty values, names long enough to force a PAX path record, optional access and change times — writes them with `tar.Writer`, and compares every materialized header and `FileInfo` against what `archive/tar` produces for the same bytes. 200 seeded runs, deterministic, 0.03s. With the fixture tests disabled it catches, on its own: the `int32` narrowing, uid/gid swapped, `devminor` dropped, `format` dropped, `ChangeTime` dropped, `Uname` not restored, `Linkname` dropped, empty-valued PAX records skipped, the `Xattrs` mirror dropped, and `fileMode` derived by hand instead of from `archive/tar`.

Fixture tests: header round-trip against `archive/tar` for directories, a setuid file with checksum and xattr PAX records, a symlink, a hardlink, a PAX long name, and a member with access/change times (`reflect.DeepEqual`, and materialized maps are independent copies); `FileInfo` field-for-field against `Header.FileInfo()` including `Sys()`; and `Open`/`Stat`/`ReadDir`/`Readlink` behavior including symlink following and the synthesized root. Both the `PAXRecords` and `Xattrs` maps are mutation-checked for independence across materializations, and `Entry`'s `fs.DirEntry` methods are compared against the same reference. Fixture members cover ownership above `MaxInt32`, char/block/fifo device nodes, a GNU-format member, and an empty-valued xattr record, so the round-trip can fail on a narrowed or dropped field.

Fixes #2492.

### DIFF
--- a/pkg/apk/expandapk/tarfs/tarfs.go
+++ b/pkg/apk/expandapk/tarfs/tarfs.go
@@ -101,7 +101,6 @@ type paxRecord struct {
 type entry struct {
 	name     string
 	linkname string
-	dir      string
 	uname    unique.Handle[string]
 	gname    unique.Handle[string]
 	size     int64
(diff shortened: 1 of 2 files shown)


coreydaley-cg:
Found #2433 after opening this, same diagnosis and fix shape, two weeks earlier. This keeps `Entry.Header` as an exported field by materializing on request, where #2433 makes it a method. I've folded in your decoded-checksum field, gated on the exact lowercase spelling so uppercase and `Q1` forms still round-trip verbatim. If you'd rather carry #2433 forward instead, say so and I'll close this.

coreydaley-cg:
Closing in favor of #2433, which has the same diagnosis and fix and is the one being carried forward. The decoded-checksum field originated there anyway. #2492 stays open and #2433 closes it.

PR #2512, opened 2026-09-19
paths: tolerate rare incomplete reads in the concurrency test

stevebeattie:
`TestReplaceCachedFileConcurrent` has been flaky since it landed. **`main` is currently red on it** — the `test` job has failed on the last two commits (`bb7b498f`, `2bfb5f2e`).

A reader intermittently fails to read `dst` — `ENOENT`, or `EISDIR` on a path that is only ever a symlink — a handful of times in tens of thousands of reads. Locally that is roughly 8 runs in 60.

## It does not appear to be `ReplaceCachedFile`

The destination is a symlink on every observation and is never unlinked; the one branch that could briefly remove it (`RemoveAll`, for a planted directory) never runs; disabling the orphan cleanup changes nothing. The same failure reproduces outside this repository in a standalone C program using nothing but `open`, `fstat`, `read`, `symlink` and `rename` — no apko, no Go.

Whether that makes it kernel behaviour or something still wrong in how these entries are published is being chased separately in **PSEC-2866**, which holds the detail. That question is worth answering, but it should not keep `main` red while it is open, so **this is a short-term unblock** and the tolerance it adds should be revisited once there is an answer.

## Not a blanket "ignore errors"

That would have been worse than the flake. An implementation that unlinks before symlinking produces nothing *but* misses, so a test that shrugs at misses passes it — confirmed by writing that implementation and watching the naive fix accept it.

Two assertions replace the one:

- **A read that completes must return the published content.** A torn or stale publish still fails hard; that is the actual promise.
- **Reads that do not complete must stay rare.** The two cases are far apart — zero incomplete reads out of ~34,000 for a correct implementation, against ~51,000 out of ~78,000 for unlink-then-symlink — so the bound is one per cent of completed reads, with a floor for short runs.

A rare incomplete read is in any case indistinguishable from a cache miss to every caller here, and `getPackageImpl` already treats a miss as something the refetch repairs.

## Verification

Both directions: **200 consecutive runs clean** on the real implementation, and the non-atomic implementation still fails every time, now reporting the rate that gave it away.

Also `go build ./...`, `go vet ./...`, `gofmt`, `goimports`, `golangci-lint` (0 issues), and `make test` under a random `SOURCE_DATE_EPOCH`.


jlucangelio (approved):
This is better than keeping main red.

PR #2516, opened 2026-09-21
fix(apko): fix CON-2732

jlucangelio:
- **passwd, apk: range-check uid/gid instead of truncating to uint32**
- **apk: refuse to validate owners without a 64-bit int**
- **apk: apply the 64-bit-int guard at the archive boundary only**
- **apk: exercise the owner boundary matrix against real tar bytes**


stevebeattie (approved):

fthomasella:
**Security review: GHSA-rj6x-hwpv-vhvf (UID/GID integer truncation). No bypasses found.**

I audited this fix. Details of what was verified:

**Parse side (the advisory itself)**
- `parseID` (`ParseUint(s, 10, 32)`) is sound by construction: the only success
  path yields a value in [0, 2^32-1], so a truncating cast is impossible rather
  than merely checked afterwards. Both `/etc/passwd` (UID + GID) and
  `/etc/group` (GID) go through it, and I confirmed these are the only
  passwd/group parsers in the tree (`shadow` has no numeric id field).
- Failure is fail-closed at every level: `Parse` → `Load` aborts on the first
  bad line (the entry never enters `Entries`, so there is no zero-value
  `UID=0` struct reuse) → `ReadOrCreate*File` → the errgroup aborts the build.
  `WriteFile` only runs after a successful load, so there are no partial writes.
- No false rejections for legitimate distro content: `uid_t`/`gid_t` are u32,
  and leading zeros ("0000") still parse as an explicitly declared 0.

**Archive side**
- `checkArchiveOwner` runs before any fs mutation on both install paths
  (streaming and lazy), and the entries validated on the lazy path are the same
  objects mounted via `WriteHeader`, so the EROFS writer (which reads owners
  from `Sys()`) only ever sees validated headers.
- Refusing outright on 32-bit `int` platforms is the right call:
  `archive/tar` truncates before apko sees the value ("Integer overflow
  possible" in the stdlib), and a GNU base-256 owner carries no PAX record to
  re-read, so a post-hoc range check cannot recover it.
- The `AddInstalledPackage` write-side check plus the `parseInstalledPerms`
  read-side range check close the installed-DB round trip — this also prevents
  an unparseable DB that `hasUsrMergeBaseImage` would swallow into picking the
  wrong layout.
- The YAML config surface is covered structurally (`types.UID/GID` are
  `*uint32`, so out-of-range values fail at unmarshal), and `run-as` is a
  string passed through to the OCI config — resolution happens in the runtime,
  which fails loudly on out-of-range numerics.

**Tests**: the boundary matrix exercised against real tar bytes (PAX + GNU
base-256, both field widths) is exactly right — the fixture generator beats
hand-built headers.

**Non-blocking observations** (not regressions from this PR):
1. `uidGidFromInfo` (pkg/build/erofs.go) falls back to `(0, 0)` when
   `FileInfo.Sys()` is not a `*tar.Header` — a silent default-to-root. It is
   pre-existing and unreachable today (every mounted node carries a validated
   header), but it is adjacent in spirit to this advisory; worth a follow-up
   to fail or warn instead of defaulting to root.
2. Library consumers only: a mid-stream owner rejection leaves entries written
   before the bad one with no DB record (as documented in `abortInstalls`).
   That is a data-integrity degradation, not a privilege issue, and strictly
   better than the pre-fix silent root — fine as-is given the CLI aborts.

No residual vector for the truncation class found. Good to go from a security
standpoint.

